PR #64, opened 2022-01-06
Upgrade smithy to 1.16 and use the new input/output

alextwoods:
*Description of changes:*
Upgrade smithy dependency to 1.16 (minimum, required to use the new getInputShape/getOutputShape that don't return options).  

Note: Tests are currently failing, still need to apply the new transform on models for input/output shapes.


white_label/sig/white_label/types.rbs
```diff
@@ -10,24 +10,20 @@
 module WhiteLabel
   module Types
 
+    ClientError: untyped
```
mullermp, line 13: Did ordering change?

white_label/lib/white_label/stubs.rb
```diff
@@ -238,6 +245,7 @@ def self.default(visited=[])
       end
 
       def self.stub(http_resp, stub:)
+        data = {}
```
mullermp, line 248: Is this a bug fix? How are these being added?

mullermp (approved):

white_label/sig/white_label/types.rbs
alextwoods: Yeah, it looks like the transform results in some order changes. But I dont think those are significant.  For many generators we do a sort by name.

white_label/lib/white_label/stubs.rb
alextwoods: This was actually added in the documentation update that I did. 
This code was incorrect (data did not exist and so these would error out).  I made a few changes in the Base builders that made this a bit more consistent so that implementations that extend them don't need to initialize data themselves.

PR #67, opened 2022-01-11
Prerelease checkpoint 3

mullermp:
PR will fix a bunch of other issues


codegen/build.gradle.kts
```diff
@@ -249,14 +249,12 @@ subprojects {
         tasks["spotbugsTest"].enabled = false
 
         // Configure the bug filter for spotbugs.
-        tasks.withType<com.github.spotbugs.SpotBugsTask> {
-            ignoreFailures = false
-
-            effort = "max"
-            excludeFilterConfig = project.resources.text.fromFile("${project.rootDir}/config/spotbugs/filter.xml")
-            reports {
-                xml.setEnabled(false)
```
mullermp, line 258: How do we enable html and disable xml reports?

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/ErrorsGeneratorBase.java
```diff
@@ -35,7 +36,7 @@
 /**
  * ErrorGeneratorBase that protocol generators can extend to get
  * common functionality.
- *
+ * <p>
```
mullermp, line 39: Random <p>

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/ErrorsGeneratorBase.java
```diff
@@ -162,23 +163,29 @@ public void renderRbsErrorCode() {
...
 
     private void renderServiceModelErrors() {
-        errorShapes.forEach(error -> {
-            String errorName = symbolProvider.toSymbol(error).getName();
-            ErrorTrait errorTrait = error.getTrait(ErrorTrait.class).get();
-            String apiErrorType = getApiErrorType(errorTrait);
-
-            renderServiceModelError(errorName, apiErrorType);
-        });
+        errorShapes
```
mullermp, line 166: Is this refactorable so we don't have to do it twice?

alextwoods: Yes, but technically changes the interface slightly (set to list).  fine for pre-release here, so I've done that.

PR #70, opened 2022-01-19
Fix docs with "$"

alextwoods:
*Description of changes:*
Don't attempt to format doc strings that may have "$" in them.


codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/RubyCodeWriter.java
```diff
@@ -90,7 +90,8 @@ public RubyCodeWriter writeYardAttribute(String attribute, Runnable task) {
      */
     public RubyCodeWriter writeDocstring(String docstring) {
         doc(() -> {
-            write("$L", docstring);
+
+            writeWithNoFormatting(docstring);
```
mullermp, line 94: I think this change isn't necessary

alextwoods: It's not necessary, but I find it it more clear as `writeWithNoFormatting` expresses the actual intent here.

mullermp (approved):

PR #73, opened 2022-01-25
Large number of fixes for aws json protocol tests

alextwoods:
*Description of changes:*
See: https://github.com/alextwoods/aws-sdk-ruby-v4/pull/1

Fixes have been ported into rails json and have protocol tests added.


seahorse/lib/seahorse/number_helper.rb
```diff
@@ -1,14 +1,12 @@
...
+  # A module that provides helper methods to convert Numeric objects to
   # protocol specific serializable formats.
   # @api private
   module NumberHelper
     class << self
-      # @param [Number,String] n
+      # @param [Number,String] input
       # @return [String] The serialized number
       def serialize(input)
         if input == ::Float::INFINITY then 'Infinity'
```
mullermp, line 12: Should these instead be case statements? You might be able to check ::Float::NAN instead of calling .nan?

mullermp (approved):

seahorse/lib/seahorse/number_helper.rb
alextwoods: I did have this as a case statement, but rubocop got mad at me for this because it's not straightforward comparisions. 
 And Nan is never equal to nan, so that check doesn't work.

PR #74, opened 2022-01-26
Major generator base refactorings.

alextwoods:
*Description of changes:*

There are two primary changes, applied across builder, parser and stubs generators:
1. Create a new top level base class (eg BuilderGeneratorBase), which iterates shapes and generates a skeleton.  It is 100% protocol agnostic.  Many protocols should use this as the base implemetnation.
2. Refactor the old "Http" base genertors to "Rest" base - these apply for protocols that fully implement http binding traits (eg, rails-json, Aws restJson and Aws restXml).  

Instead of generating the skeletons for build/parse methods - push that down to implementors and require them to implement the skeletons of those methods as well - this removes assumptions about the signatures of those methods.

This does NOT change the signatures of operation builders/parsers (these are used by the middleware and must conform to that interface. protocols should NOT change those).

For the changes downstream in the AWS generators see: https://github.com/alextwoods/aws-sdk-ruby-v4/pull/2.  


mullermp (approved):

mullermp (approved):
Nice!

PR #77, opened 2022-02-16
Transform waiter path field names using new JmesPath ExpressionSerializer

alextwoods:
*Description of changes:*
Uses the ExpressionSerializer introduced in: https://github.com/awslabs/smithy/pull/1059

### DIFF
--- a/codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/WaitersGenerator.java
+++ b/codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/WaitersGenerator.java
@@ -15,6 +15,7 @@
 
 package software.amazon.smithy.ruby.codegen.generators;
 
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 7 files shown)


codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/WaitersGenerator.java
```diff
@@ -300,4 +319,168 @@ public Void visitUnknown(Matcher.UnknownMember unknown) {
...
+                    expression.getLeft().accept(this),
+                    expression.getRight().accept(this),
+                    expression.getLine(),
+                    expression.getColumn()
+            );
+        }
+
+        @Override
+        public JmespathExpression visitSlice(SliceExpression expression) {
+            return expression; // TODO
```
mullermp, line 472: Are these optional or mandatory for us?

alextwoods: This one - I'm on the fence about.  Theres nothing that gets transformed in this, but we shouldn't re-use the existing expression object (all of the visitors need to return a copy).

mullermp (approved):

PR #79, opened 2022-03-09
Add query param matcher

mullermp:
Add query param matcher for AWS Query tests. It compares the parsed CGI string (as a hash) but also checks float precision.


codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/ErrorsGeneratorBase.java
```diff
@@ -152,14 +152,17 @@ public void renderRbsErrorCode() {
         rbsWriter.write("def self.error_code: (untyped http_resp) -> untyped");
     }
 
-    protected Set<Shape> getErrorShapes() {
+    protected ArrayList<Shape> getErrorShapes() {
```
alextwoods, line 155: You should be able to just use List here

alextwoods (approved):

PR #81, opened 2022-03-10
Convert stubs to Output types - apply params/validation

alextwoods:
*Description of changes:*
When stubs are supplied (eg when initializing the client or calling `stub_responses`), convert/transform Hash params to modeled Output Types using the Params/validators (as is done for inputs).  This gives the same validation that we get on normal method calls and will raise an error when stubs are invalid - this keeps our stub code from needing to handle error checking/invalid cases.

TODO: All of the stub defaults need to return the correct types.

### DIFF
--- a/codegen/projections/high_score_service/lib/high_score_service/client.rb
+++ b/codegen/projections/high_score_service/lib/high_score_service/client.rb
@@ -386,5 +386,23 @@ module HighScoreService
 
       StringIO.new
     end
+
+    def transform_stubs(operation_name, stubs)
+      param_class = case operation_name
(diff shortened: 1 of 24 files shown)


codegen/projections/high_score_service/lib/high_score_service/client.rb
```diff
@@ -386,5 +386,23 @@ def output_stream(options = {}, &block)
 
       StringIO.new
     end
+
+    def transform_stubs(operation_name, stubs)
```
mullermp, line 390: I'm not a fan of this... should the class just be passed in to stubbing middleware somehow?

PR #82, opened 2022-03-11
Convert stubs to Output types at call time

alextwoods:
*Description of changes:*

When a stubbed operation is called convert/transform Hash params to modeled Output Types using the Params/validators (as is done for inputs). This gives the same validation that we get on normal method calls and will raise an error when stubs are invalid - this keeps our stub code from needing to handle error checking/invalid cases.

This is done at operation stub/call time to be consistent with the use of Procs as a stub type. 

### DIFF
--- a/codegen/projections/high_score_service/lib/high_score_service/client.rb
+++ b/codegen/projections/high_score_service/lib/high_score_service/client.rb
@@ -102,6 +102,7 @@ module HighScoreService
         data_parser: Parsers::CreateHighScore
       )
       stack.use(Hearth::Middleware::Send,
+        stub_params_class: Params::CreateHighScoreOutput,
         stub_responses: options.fetch(:stub_responses, @stub_responses),
         client: Hearth::HTTP::Client.new(logger: @logger, http_wire_trace: options.fetch(:http_wire_trace, @http_wire_trace)),
         stub_class: Stubs::CreateHighScore,
@@ -159,6 +160,7 @@ module HighScoreService
         data_parser: Parsers::DeleteHighScore
       )
       stack.use(Hearth::Middleware::Send,
+        stub_params_class: Params::DeleteHighScoreOutput,
         stub_responses: options.fetch(:stub_responses, @stub_responses),
         client: Hearth::HTTP::Client.new(logger: @logger, http_wire_trace: options.fetch(:http_wire_trace, @http_wire_trace)),
         stub_class: Stubs::DeleteHighScore,
@@ -222,6 +224,7 @@ module HighScoreService
         data_parser: Parsers::GetHighScore
       )
       stack.use(Hearth::Middleware::Send,
+        stub_params_class: Params::GetHighScoreOutput,
         stub_responses: options.fetch(:stub_responses, @stub_responses),
         client: Hearth::HTTP::Client.new(logger: @logger, http_wire_trace: options.fetch(:http_wire_trace, @http_wire_trace)),
         stub_class: Stubs::GetHighScore,
@@ -281,6 +284,7 @@ module HighScoreService
         data_parser: Parsers::ListHighScores
       )
       stack.use(Hearth::Middleware::Send,
+        stub_params_class: Params::ListHighScoresOutput,
         stub_responses: options.fetch(:stub_responses, @stub_responses),
         client: Hearth::HTTP::Client.new(logger: @logger, http_wire_trace: options.fetch(:http_wire_trace, @http_wire_trace)),
         stub_class: Stubs::ListHighScores,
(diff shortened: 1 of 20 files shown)


codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/middleware/MiddlewareBuilder.java
```diff
@@ -161,7 +161,9 @@ public void addDefaultMiddleware(GenerationContext context) {
                         transport.getTransportClient().render(context))
                 .operationParams((ctx, operation) -> {
                     Map<String, String> params = new HashMap<>();
+                    Shape outputShape = ctx.getModel().expectShape(operation.getOutputShape());
                     params.put("stub_class", "Stubs::" + symbolProvider.toSymbol(operation).getName());
+                    params.put("stub_params_class", "Params::" + symbolProvider.toSymbol(outputShape).getName());
```
mullermp, line 166: I think "params_class" is fine here. There's nothing special about it being a stub.

hearth/lib/hearth/middleware/send.rb
```diff
@@ -11,12 +11,16 @@ class Send
       # @param [Class] stub_class A stub object that is responsible for creating
       #   a stubbed response. It must respond to #stub and take the response
       #   and stub data as arguments.
+      # @param [Class] stub__params_class A Params class responsible for
```
mullermp, line 14: Doc name is wrong

hearth/lib/hearth/middleware/send.rb
```diff
@@ -50,9 +54,17 @@ def apply_stub(stub, context, output)
         when Class
           output.error = stub.new
         when Hash
-          @stub_class.stub(context.response, stub: stub)
+          @stub_class.stub(context.response,
+                           stub: @stub_params_class.build(
+                             stub,
+                             context: "stub[#{context.operation_name}]"
+                           ))
```
mullermp, line 61: These aren't very readable... maybe split it out on multiple lines?

hearth/spec/hearth/middleware/send_spec.rb
```diff
@@ -113,23 +125,35 @@ module Middleware
...
           context 'stub is a hash' do
             let(:stub_hash) { { param1: 'value' } }
+            let(:output_type) { StubOutput.new(**stub_hash) }
+
             before { stubs.add_stubs(operation, [stub_hash]) }
 
             it 'uses the stub class to stub the response' do
+              expect(stub_params_class).to receive(:build)
+                .with(stub_hash, context: 'stub[operation]')
+                .and_return(output_type)
```
mullermp, line 135: Why create StubOutputParams (with implementation) if you are mocking it? I'd say make a double like stub_class instead of defining StubOutputParams

mullermp (changes requested):

hearth/lib/hearth/middleware/send.rb
```diff
@@ -50,9 +54,17 @@ def apply_stub(stub, context, output)
         when Class
           output.error = stub.new
         when Hash
-          @stub_class.stub(context.response, stub: stub)
+          @stub_class.stub(context.response,
+                           stub: @stub_params_class.build(
+                             stub,
+                             context: "stub[#{context.operation_name}]"
```
mullermp, line 60: I think just "stub" for context is fine, since operation is known when invoking.

hearth/lib/hearth/middleware/send.rb
alextwoods: I had them more readable but rubocop complained and autocorrected to this.  But updated to something else.

codegen/projections/high_score_service/lib/high_score_service/client.rb
```diff
@@ -102,6 +102,7 @@ def create_high_score(params = {}, options = {}, &block)
         data_parser: Parsers::CreateHighScore
       )
       stack.use(Hearth::Middleware::Send,
+        stub_params_class: Params::CreateHighScoreOutput,
```
mullermp, line 105: We will have to regenerate these right?

hearth/.rubocop.yml
```diff
@@ -24,6 +26,10 @@ Metrics/ClassLength:
   Exclude:
     - 'lib/hearth/middleware_builder.rb'
 
+Metrics/ParameterLists:
+  Exclude:
+    - 'lib/hearth/middleware/send.rb'
```
mullermp, line 31: What's the deal with these? Should we condense the arg list?

mullermp (approved):

PR #83, opened 2022-03-12
Use Smithy CodegenIntegration

alextwoods:
*Description of changes:*
Use the new Smithy CodegenIntegration interfaces introduced in 1.18 (https://github.com/awslabs/smithy/pull/1109/files)


codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/GenerationContext.java
```diff
@@ -56,38 +60,48 @@ public GenerationContext(RubySettings rubySettings,
...
+        return integrations;
     }
 
-    public ServiceShape getService() {
+    public ServiceShape service() {
         return service;
     }
 
-    public ApplicationTransport getApplicationTransport() {
+    public ApplicationTransport applicationTransport() {
```
mullermp, line 94: These don't follow Java convention of getThing. But I assume you did this because to be consistent with CodeGen context?

mullermp (approved):

PR #84, opened 2022-03-14
Add query param abstraction

mullermp:
For query/rest xml, have a glorified hash wrapper builder thing.


hearth/lib/hearth/query/param_list.rb
```diff
@@ -0,0 +1,57 @@
...
+      end
+
+      # @return [Array<Param>] Returns an array of sorted {Param} objects.
+      def to_a
+        @params.values.sort
+      end
+
+      # @return [String]
+      def to_s
+        to_a.map(&:to_s).join('&')
```
alextwoods, line 53: How does this handle list/set of params? the @httpQuery param requires us to do eg:
`foo = ["a", "b"]` serialize to `foo=a&foo=b`

https://awslabs.github.io/smithy/1.0/spec/core/http-traits.html#httpquery-trait

mullermp: Ah.. It doesn't. I mostly lifted from v3. Is that mainly rest?

alextwoods: Yeah, I believe only rest support the @httpQueryTrait.

hearth/lib/hearth/query/param.rb
```diff
@@ -0,0 +1,53 @@
...
+
+      # @return [Boolean]
+      def <=>(other)
+        name <=> other.name
+      end
+
+      private
+
+      def serialize(name, value)
+        value ? "#{escape(name)}=#{escape(value)}" : "#{escape(name)}="
```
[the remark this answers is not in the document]
mullermp: I decided to do them in the protocol builders. I think node/xml work similarly (expect strings)?

alextwoods (approved):

PR #89, opened 2022-03-18
Add SkipTestTrait for ignoring/skipping protocol tests

alextwoods:
*Description of changes:*
Add a trait to allow skipping a protocol test.


mullermp (commented):
Why the need for two trait classes?

alextwoods:
There is only one trait class (SkipTestsTrait) - but it defines a list of tests to skip.
We can only apply the trait once per operation, but we may need to skip multiple tests, so we accept a list.  This works identically to the way protocol test traits are defined.

mullermp:
But there's a trait/SkipTest.java and trait/SkipTestsTrait.java

alextwoods:
We could move the SkipTest class somewhere else if you prefer, but I followed the design for the HttpResponseTestsTrait and HttpResponseTestCase class  (both are in the software.amazon.smithy.protocoltests.traits package).

mullermp (approved):

PR #94, opened 2022-03-23
Update reserved words

mullermp:
This change updates reserved words for all ruby keywords and for instance methods of a Ruby struct. This removes "name" and other CLASS methods.

There are different approaches we can take:

* Use classes instead of structs for types. Enumerable methods would not need to be reserved words. This approach involves a lot of change.
* (This PR) add enumerable as reserved words. V3 -> V4 migration is slightly more complex but allows for enumerable methods on Struct to be used without hinderance.
* Do nothing, always overwrite enumerable methods. V3 does this today, but we haven't had customer reachouts.


mullermp:
Needs discussion on approaches.

alextwoods (approved):

PR #95, opened 2022-03-23
Add metadata to Hearth::Output and Errors, and return Hearth::Output on successful calls

mullermp:
Breaking change.

Return Hearth::Output from operations, instead of the output structure. This is similar to V3 (without delegation however).

This allows middleware to set context values and have it be accessible to users.

### DIFF
--- a/codegen/smithy-ruby-codegen-test/integration-specs/errors_spec.rb
+++ b/codegen/smithy-ruby-codegen-test/integration-specs/errors_spec.rb
@@ -10,6 +10,17 @@ module WhiteLabel
         error = ApiError.new(http_resp: http_resp, error_code: 'error')
         expect(error).to be_a(Hearth::HTTP::ApiError)
       end
+
+      it 'stores a request_id' do
+        http_resp = Hearth::HTTP::Response.new
+        error = ApiError.new(
+          http_resp: http_resp,
+          request_id: '123',
+          error_code: 'error',
+          message: 'error message'
+        )
+        expect(error.request_id).to eq('123')
+      end
     end
 
     describe ApiClientError do
--- a/codegen/smithy-ruby-codegen-test/integration-specs/errors_spec.rb
+++ b/codegen/smithy-ruby-codegen-test/integration-specs/errors_spec.rb
@@ -7,26 +7,15 @@ module WhiteLabel
     describe ApiError do
       it 'inherits the base protocol api error' do
         http_resp = Hearth::HTTP::Response.new
-        error = ApiError.new(http_resp: http_resp, error_code: 'error')
+        error = ApiError.new(http_resp: http_resp, metadata: {}, error_code: 'error')
         expect(error).to be_a(Hearth::HTTP::ApiError)
-      end
-
-      it 'stores a request_id' do
-        http_resp = Hearth::HTTP::Response.new
-        error = ApiError.new(
-          http_resp: http_resp,
-          request_id: '123',
-          error_code: 'error',
-          message: 'error message'
-        )
-        expect(error.request_id).to eq('123')
       end
     end
 
     describe ApiClientError do
       it 'inherits the base client api error' do
         http_resp = Hearth::HTTP::Response.new
-        error = ApiClientError.new(http_resp: http_resp, error_code: 'error')
+        error = ApiClientError.new(http_resp: http_resp, metadata: {}, error_code: 'error')
         expect(error).to be_a(ApiError)
       end
     end
@@ -34,7 +23,7 @@ module WhiteLabel
     describe ApiServerError do
       it 'inherits the base client api error' do
         http_resp = Hearth::HTTP::Response.new
-        error = ApiServerError.new(http_resp: http_resp, error_code: 'error')
+        error = ApiServerError.new(http_resp: http_resp, metadata: {}, error_code: 'error')
         expect(error).to be_a(ApiError)
       end
     end
@@ -42,7 +31,12 @@ module WhiteLabel
     describe ApiRedirectError do
       it 'inherits the base client api error' do
         http_resp = Hearth::HTTP::Response.new
-        error = ApiRedirectError.new(location: nil, http_resp: http_resp, error_code: 'error')
+        error = ApiRedirectError.new(
+          location: nil,
+          http_resp: http_resp,
+          metadata: {},
+          error_code: 'error'
+        )
         expect(error).to be_a(ApiError)
       end
 
@@ -52,6 +46,7 @@ module WhiteLabel
           http_resp: http_resp,
           location: 'location',
           error_code: 'error',
+          metadata: {},
           message: 'error message'
         )
         expect(error.location).to eq('location')
@@ -64,7 +59,7 @@ module WhiteLabel
         data = Types::ClientError.new(message: 'error message')
         # don't test fakeProtocol parsers
         expect(Parsers::ClientError).to receive(:parse).with(http_resp).and_return(data)
-        error = ClientError.new(http_resp: http_resp, error_code: 'error')
+        error = ClientError.new(http_resp: http_resp, metadata: {}, error_code: 'error')
         expect(error.data).to eq(data)
         expect(error).to be_a(ApiClientError)
         expect(error.message).to eq('error message')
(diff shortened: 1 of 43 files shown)


mullermp:
WIP - needs update on tests and potentially breaking CI (it SHOULD break?)

alextwoods (commented):
Looks good overall.  We should add an rbs for Output now thats its public.

hearth/lib/hearth/http/api_error.rb
```diff
@@ -9,7 +9,6 @@ def initialize(http_resp:, **kwargs)
         @http_status = http_resp.status
         @http_headers = http_resp.headers
         @http_body = http_resp.body
-        @request_id = http_resp.headers['x-request-id']
```
mullermp, line 12: This is considered "common" and not "standard" in http spec, so I opted to remove it here.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/middleware/MiddlewareStackStep.java
```diff
@@ -42,8 +42,8 @@ public String toString() {
                 return "Retry";
             case SERIALIZE:
                 return "Serialize";
-            case BUILD:
-                return "Build";
+            case VALIDATE:
```
mullermp, line 45: Build is confusing with serialize. I think we needed one at the validate step.

codegen/smithy-ruby-rails-codegen/middleware/request_id.rb
```diff
@@ -0,0 +1,16 @@
+module Middleware
```
mullermp, line 1: This might be global scope isn't it? I think we'd need to write out Service::Middleware

codegen/smithy-ruby-rails-codegen/middleware/request_id.rb
```diff
@@ -0,0 +1,16 @@
...
+  # on the context.
+  # @api private
+  class RequestId
+    def initialize(app)
+      @app = app
+    end
+
+    def call(input, context)
+      output = @app.call(input, context)
+      output.metadata[:request_id] = context.response.headers['x-request-id']
```
mullermp, line 12: I don't like that I had to parse here (for data) and in error parser (for errors). I'd like one place to put this information.

codegen/smithy-ruby-rails-codegen/middleware/request_id.rb
alextwoods: Yeah, this either needs to be in a core gem or needs to be codegened out with a unique namespace....

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/ApplicationTransport.java
```diff
@@ -126,7 +126,7 @@ public static ApplicationTransport createDefaultHttpApplicationTransport() {
             List<Middleware> middleware = new ArrayList();
             middleware.add((new Middleware.Builder())
                     .klass("Hearth::HTTP::Middleware::ContentLength")
-                    .step(MiddlewareStackStep.BUILD)
+                    .step(MiddlewareStackStep.SERIALIZE)
```
alextwoods, line 129: Why is this moved to Serialize?

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/middleware/MiddlewareStackStep.java
alextwoods: These steps were intended to match those used by other SDKs. See for example
* https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/modules/_aws_sdk_middleware_stack.html
* https://aws.github.io/aws-sdk-go-v2/docs/middleware/#:~:text=You%20can%20customize%20AWS%20SDK,step's%20input%20and%20output%20types.

The meanings were intended to be:
* Serialize - Serialize the given parameter input into a request data structure that can be confusmed by the transport layer (eg, an HTTP:Request).
* Build - Provide additional transformation of the Request after serialization (eg ContentLength)

I agree however that because of our choice to use "Builders" for Serialization that this is a bit confusing.  I don't think VALIDATE is the right name for this step though.

codegen/smithy-ruby-rails-codegen/middleware/request_id.rb
alextwoods: I agree - I'd rather be able to do this in a single place.  But I don't believe its possible without some interface tweak, either adding output to the error parser interface (to allow access to metadata) or to add write access to the request_id.

We can add either of those later if we want (but taking them away would be hard).  So, lets leave this for now and accept some duplication.

alextwoods (commented):
Overall looks good and I think we should move forward with changing our return type.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/ApplicationTransport.java
mullermp: I removed Build because it didn't make sense, it's now "validate". Content length should be after serialize.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/middleware/MiddlewareStackStep.java
mullermp: I just undid it. Can re-evaluate later.

alextwoods (approved):

codegen/smithy-ruby-rails-codegen/middleware/request_id.rb
mullermp: Talked offline, metadata is now a property of errors.

codegen/smithy-ruby-codegen-test/integration-specs/errors_spec.rb
```diff
@@ -10,6 +10,17 @@ module Errors
         error = ApiError.new(http_resp: http_resp, error_code: 'error')
         expect(error).to be_a(Hearth::HTTP::ApiError)
       end
+
+      it 'stores a request_id' do
```
alextwoods: Does this still apply or should it be metadata?

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/ErrorsGeneratorBase.java
```diff
@@ -141,13 +144,13 @@ private void renderRbsBaseErrors() {
     }
 
     /**
-     * Called to render the error_code method.
+     * Called to render the error_code method body.
```
alextwoods, line 147: Unrelated nit: We should probably describe what error code is (ie, the name of the class to use....  Because this function also may map the actual data to a generated class, including reserved word name changes...).

hearth/lib/hearth/http/error_parser.rb
```diff
@@ -30,22 +30,17 @@ class ErrorParser
       #
       # @param [Array<Class<ApiError>>] errors Array of Error classes
       #   modeled for the operation.
-      #
-      # @param [callable] error_code_fn Protocol specific function
-      #   that will return the error code from a response, or nil if
-      #   there is none.
-      def initialize(error_module:, success_status:, errors:, error_code_fn:)
+      def initialize(error_module:, success_status:, errors:)
```
alextwoods, line 33: Nit: add to documenation of error_module, it must contain definition of error code function

hearth/lib/hearth/http/error_parser.rb
```diff
@@ -30,22 +30,17 @@ class ErrorParser
...
         @error_module = error_module
         @success_status = success_status
         @errors = errors
-        @error_code_fn = error_code_fn
       end
 
       # Parse and return the error if the response is not successful.
       #
       # @param [Response] response The HTTP response
```
alextwoods, line 41: Nit: Update @params for metadata

alextwoods (approved):

codegen/smithy-ruby-codegen-test/integration-specs/errors_spec.rb
mullermp: Fixed and deleted

PR #98, opened 2022-03-30
Add support for unboxed (default) values in Types

alextwoods:
*Description of changes:*
 Set non-nillable values to defaults in Types.

### DIFF
--- a/codegen/smithy-ruby-codegen-test/integration-specs/types_spec.rb
+++ b/codegen/smithy-ruby-codegen-test/integration-specs/types_spec.rb
@@ -59,8 +59,8 @@ module WhiteLabel
       end
     end
 
-    describe ListCitiesInput do
-      subject { ListCitiesInput.new }
+    describe DefaultsTestInput do
+      subject { DefaultsTestInput.new }
 
       it 'has a default value for default_number' do
         expect(subject.default_number).to be 0
@@ -68,6 +68,10 @@ module WhiteLabel
 
       it 'has a default value for default_bool' do
         expect(subject.default_bool).to be false
+      end
+
+      it 'unboxed_number defaults to nil' do
+        expect(subject.boxed_number).to be nil
       end
     end
   end
--- a/codegen/smithy-ruby-codegen-test/integration-specs/types_spec.rb
+++ b/codegen/smithy-ruby-codegen-test/integration-specs/types_spec.rb
@@ -58,5 +58,17 @@ module WhiteLabel
         expect(unknown_union.to_h).to eq({ unknown: { key: 'key', value: 'value' } })
       end
     end
+
+    describe ListCitiesInput do
+      subject { ListCitiesInput.new }
+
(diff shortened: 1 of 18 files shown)


codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/TypesGenerator.java
```diff
@@ -267,6 +293,19 @@ private void renderUnionToSMethod(MemberShape memberShape) {
         }
     }
 
+    private class MemberDefaultVisitor extends ShapeVisitor.Default<String> {
+
+        @Override
+        protected String getDefault(Shape shape) {
```
mullermp, line 299: What about for Strings, blobs, etc? Is default the right visitor method?

codegen/smithy-ruby-codegen-test/model/component-test/main.smithy
```diff
@@ -74,10 +75,34 @@ union Union {
...
 structure ClientError {
   Message: String
 }
 
 @error("server")
-structure ServerError {}
\ No newline at end of file
+structure ServerError {}
+
+operation DefaultsTest {
```
mullermp, line 87: This file is supposed to be for shared components. Can you move this out into another file?

codegen/smithy-ruby-codegen-test/integration-specs/types_spec.rb
```diff
@@ -58,5 +58,21 @@ module Types
         expect(unknown_union.to_h).to eq({ unknown: { key: 'key', value: 'value' } })
       end
     end
+
+    describe DefaultsTestInput do
+      subject { DefaultsTestInput.new }
```
mullermp, line 63: I believe subject is implicitly defined already

mullermp (approved):

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/TypesGenerator.java
alextwoods: Currently yes - only the number types and String are supported.  We could implement all of the numbers individually here or just use the default.  Thoughts?

codegen/smithy-ruby-codegen-test/integration-specs/types_spec.rb
alextwoods: Ah, yes it is.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/TypesGenerator.java
mullermp: What's safest? If Blob is supported, then we're writing 0, right?

PR #99, opened 2022-03-30
Retryable error support

mullermp:
Support retryable and throttling in generated code and in middleware.

Controversial change includes not having backoff unless it's a throttling error. Standard and adaptive retry modes can mitigate this but it's a reasonable first approach.


codegen/smithy-ruby-codegen-test/integration-specs/errors_spec.rb
```diff
@@ -64,6 +64,16 @@ module Errors
         expect(error).to be_a(ApiClientError)
         expect(error.message).to eq('error message')
       end
+
+      it 'is retryable without throttling' do
+        http_resp = Hearth::HTTP::Response.new
+        data = Types::ClientError.new
+        # don't test fakeProtocol parsers
+        expect(Parsers::ClientError).to receive(:parse).with(http_resp).and_return(data)
```
alextwoods, line 72: This is outside the scope of this PR - but I'm a bit concerned we can't initialize our errors without them parsing.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/ErrorsGeneratorBase.java
```diff
@@ -74,7 +89,7 @@ public void render(FileManifest fileManifest) {
                 .call(() -> renderErrorCodeBody())
                 .closeBlock("end")
                 .call(() -> renderBaseErrors())
-                .call(() -> renderServiceModelErrors())
+                .call(() -> renderServiceModelErrors(new ErrorsVisitor()))
```
alextwoods, line 92: Is there a reason we create and pass the visitor here rather than letting that method handle it?

alextwoods (approved):

PR #100, opened 2022-03-30
Add support for Streaming

alextwoods:
*Description of changes:*
Output Streaming:
- Set the response body to output_stream only on streaming operations.
- In parsers, we need to NOT read the body.  In protocols that support @httpPaylaod, @streaming can only be applied to payload members and must be on a blob.  (@streaming on a union is event streaming and will be handled differently).  

TODO:
Input streaming:
- Modify builders, add support
- Modify content length
- Modify validators

Testing:
We don't get coverage of streaming behavior from protocol tests so we'll need to add some integration tests.  

### DIFF
--- a/codegen/smithy-ruby-codegen-test/integration-specs/streaming_spec.rb
+++ b/codegen/smithy-ruby-codegen-test/integration-specs/streaming_spec.rb
@@ -3,15 +3,21 @@ require_relative 'spec_helper'
 
 module WhiteLabel
   describe Client do
-
     describe '#streaming_operation' do
-      context 'block is provided' do
-        let(:block_io) { double("BlockIO") }
-        it 'creates and uses a blockIO as the body' do
-          client = Client.new(stub_responses: true)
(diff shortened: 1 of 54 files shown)


codegen/smithy-ruby-codegen-test/integration-specs/streaming_spec.rb
```diff
@@ -79,7 +79,28 @@ module WhiteLabel
             client.streaming_operation({}, output_stream: output_stream)
           end
         end
+      end
+
+      # TODO: create an operation with @requireLength
```
mullermp: Should we be doing this?

mullermp (approved):
Nice!

codegen/smithy-ruby-codegen-test/integration-specs/streaming_spec.rb
alextwoods: Think I forgot to push the latest updates - tests for requireLength should be there.

mullermp (approved):

PR #103, opened 2022-04-04
Fix Host label formatting

alextwoods:
*Description of changes:*
Ensure host label members are correctly formatted.  Adds white label integration specs as well.

### DIFF
--- a/codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/ApplicationTransport.java
+++ b/codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/ApplicationTransport.java
@@ -84,7 +84,7 @@ public final class ApplicationTransport {
                 .type("string")
                 .documentation("Endpoint of the service")
                 .initializationCustomization("@endpoint = options.fetch(:endpoint, options[:stub_responses] "
-                        + "? 'https://example.com' : nil)")
+                        + "? 'http://localhost' : nil)")
                 .build();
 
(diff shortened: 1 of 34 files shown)


codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/ApplicationTransport.java
```diff
@@ -83,6 +83,8 @@ public static ApplicationTransport createDefaultHttpApplicationTransport() {
                 .name("endpoint")
                 .type("string")
                 .documentation("Endpoint of the service")
+                .initializationCustomization("@endpoint = options.fetch(:endpoint, options[:stub_responses] "
```
mullermp, line 86: I don't like example.com as a default. It's not owned by us and we can expect it not to be set. I'd rather raise an error in initialize if endpoint is not set (preferred) or use localhost as a reasonable default.

alextwoods: Sure - I'll switch it to localhost.  This only gets applied when stubbing is true so I don't think it matters much what it is.

mullermp: It could have a side affect, like if stub_responses isn't applied (send middleware swap?) or some other failure.

mullermp (approved):

PR #104, opened 2022-04-15
Port standard and adaptive retry into retry middleware

mullermp:
Doesn't include any code generation changes or testing yet.

Changes necessary -
* Client initialize will need to make an instance of RetryQuota and ClientRateLimiter.
* Operation will need to call stack.use with:
** max_attempts (config)
** retry_mode (config)
** adaptive_retry_wait_to_fill (config)
** instances of retry quota and client rate limiter
** new instance of error inspector (use generic Hearth and aws-sdk-core for AWS)
* Testing on retry middleware and new methods for http response/headers


mullermp:
Needs code generation support

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/middleware/MiddlewareBuilder.java
```diff
@@ -191,6 +180,63 @@ public void addDefaultMiddleware(GenerationContext context) {
...
+                .build();
+
+        Middleware retry = (new Middleware.Builder())
+                .klass("Hearth::Middleware::Retry")
+                .step(MiddlewareStackStep.RETRY)
+                .addConfig(maxAttempts)
+                .addConfig(retryMode)
+                .addConfig(adaptiveRetryWaitToFill)
+                .addParam("error_inspector_class", "Hearth::Retry::ErrorInspector")
+                .addConfig(retryQuota)
```
alextwoods, line 236: I think these should be params rather than Config, because we dont want to do an options.fetch on them

hearth/lib/hearth/middleware/retry.rb
```diff
@@ -5,49 +5,111 @@ module Middleware
...
+      # rubocop:enable Metrics/AbcSize
 
       private
 
-      # https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
-      def backoff_with_jitter(attempt)
-        # scales like 1,2,4,8
-        Kernel.rand * [@max_delay, 2**(attempt - 1)].min
+      def acquire_token
+        return unless @retry_mode != 'adaptive'
```
alextwoods, line 77: unless != is a bit confusing here.  Is this logic even correct?

mullermp: Yeah I think that's correct. It's a no op for standard.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/middleware/MiddlewareBuilder.java
mullermp: I don't think that works because I need to do an intializationCustomization

alextwoods (approved):
Nice - LGTM!

mullermp:
Finally...

PR #106, opened 2022-05-03
Config struct support

mullermp:
Introduces a Config struct that is code generated per service. Config is expected to be initialized with `Config.new(stub_responses: true)` and will resolve default values (using providers) from registered middleware. Config is expected to be passed into Client classes.


codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/CodegenOrchestrator.java
```diff
@@ -264,14 +232,76 @@ private void generateYardOpts() {
...
+        List<ClientConfig> clientConfigList = unorderedConfig.stream()
+                .sorted(Comparator.comparing(ClientConfig::getName))
+                .collect(Collectors.toList());
+
+        LOGGER.fine("Client config: "
+                + clientConfigList.stream().map((m) -> m.toString()).collect(Collectors.joining(",")));
+
+        ConfigGenerator configGenerator = new ConfigGenerator(context);
+        configGenerator.render(clientConfigList);
+        configGenerator.renderRbs();
```
alextwoods, line 296: does this not need the clientConfigList?
I'd probably rather pass clientConfigList to the constructor and then have render and renderRbs not take arguments

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/ClientGenerator.java
```diff
@@ -63,41 +57,10 @@ public ClientGenerator(GenerationContext context) {
...
-                .collect(Collectors.toList());
-
-        LOGGER.fine("Client config: "
-                + clientConfig.stream().map((m) -> m.toString()).collect(Collectors.joining(",")));
-
         this.symbolProvider = new RubySymbolProvider(model, context.settings(), "Client", false);
     }
 
-    public void render() {
+    public void render(MiddlewareBuilder middlewareBuilder) {
```
alextwoods, line 63: I'd still lean towards this being added to the constructor.  It also keeps us from having to pass it around to all of the individual methods.  Would we ever call render multiple times for the same Generator w/ different middleware? I don't think we would.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/CodegenOrchestrator.java
mullermp: RBS render will not need it AFAIK

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/ClientGenerator.java
mullermp: Maybe we should stick it on context then

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/CodegenOrchestrator.java
alextwoods: Don't we need to know fields and types the struct has to generate RBS for it?

mullermp:
Remaining work - Hearth testing, verification, and decision on build vs build! and confusion around using .new()

mullermp:
Validation should include enums and constraints of some kind.

mullermp:
Remaining work:
* do constraint validation (integer > 5) and enum validation
* review integration tests
* testing in Hearth for Resolver, Env Provider, and Configuration module

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/ApplicationTransport.java
```diff
@@ -98,23 +103,26 @@ public static ApplicationTransport createDefaultHttpApplicationTransport() {
...
+                        .dynamicProvider("proc { |cfg| Logger.new($stdout, level: cfg[:log_level]) } ")
+                        .build()
+                )
                 .documentation("Logger to use for output")
                 .build();
 
         ClientConfig logLevel = (new ClientConfig.Builder())
                 .name("log_level")
-                .type("symbol")
+                .type("Symbol")
```
alextwoods, line 125: I worry a little bit about the String usage here still.  Since we have code that requires exact values (and missing an uppercase woudl be easy).  Lets at least define constants on the Java side?  (or we could change the builder to have methods like `.SymbolType()`.  What do you think?

codegen/smithy-ruby-codegen-test/integration-specs/config_spec.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+require_relative 'spec_helper'
+
+module WhiteLabel
+  describe Config do
+    describe '#build' do
+      it 'sets member values with defaults' do
+        config_keys = {
+          adaptive_retry_wait_to_fill: false,
```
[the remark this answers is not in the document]
mullermp: This set is the base middleware, so it probably won't change often, or if it does, it's no big deal to update? Kitchen Sink is also exhaustive.

I originally had those as two tests but they were basically duplicated so I combined them.

alextwoods (approved):

PR #109, opened 2022-07-05
Upgrade Smithy to 1.22.0

mullermp:
Upgrades to Smithy 1.22 + fixes some build issues


codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/ParamsGenerator.java
```diff
@@ -59,8 +60,8 @@ public ParamsGenerator(GenerationContext context) {
         this.context = context;
         this.settings = context.settings();
         this.model = context.model();
-        this.writer = new RubyCodeWriter(context.settings().getModule());
-        this.symbolProvider = new RubySymbolProvider(model, settings, "Params", true);
+        this.writer = new RubyCodeWriter(context.settings().getModule() + "::Params");
+        this.symbolProvider = new RubySymbolProvider(model, settings, "Types", true);
```
mullermp, line 64: Is Types correct here?

alextwoods: Good question - so what we probably need is actually a few different SymbolProviders for different cases.  

The way its being used in this class we actually get symbols for both Types and Params (and in theory those are actually different symbols).

mullermp (commented):
LGTM

PR #112, opened 2022-10-20
Added directed codegen implementation

AndrewFossAWS:
*Description of changes:*

This is a basic implementation of directed codegen re-using existing ruby code generators. The goal is to ensure the same ruby code is generated and not deviate from existing codegen implementation. (There will be more PRs for refactoring and generating unions, enums, etc.)

Changes:
* Update smithy version to 1.26.0
* Added directed ruby codegen implementation
  * Added DirectedRubyCodegen class
  * Added DirectedRubyCodegenPlugin class
  * Populated existing code generation logic into DirectedRubyCodegen

Testing: 
* `./gradlew clean build`
* Confirmed generated ruby code remains the same

### DIFF
--- a/codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/RubySymbolProvider.java
+++ b/codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/RubySymbolProvider.java
@@ -15,7 +15,6 @@
 
 package software.amazon.smithy.ruby.codegen;
 
-import software.amazon.smithy.codegen.core.CodegenException;
 import software.amazon.smithy.codegen.core.ReservedWordSymbolProvider;
 import software.amazon.smithy.codegen.core.ReservedWords;
 import software.amazon.smithy.codegen.core.ReservedWordsBuilder;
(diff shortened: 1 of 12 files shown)


mullermp:
Thanks for opening a draft PR. We'll review when ready. For rubocop
```
spec/hearth/retry/client_rate_limiter_spec.rb:124:9: C: [Correctable] Style/EachForSimpleLoop: Use Integer#times for a simple loop which iterates a fixed number of times.
```

Could you also change this line to `20.times` so CI passes?

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/RubySymbolProvider.java
```diff
@@ -352,6 +351,7 @@ public Symbol operationShape(OperationShape shape) {
 
     @Override
     public Symbol resourceShape(ResourceShape shape) {
-        throw new CodegenException("Resources are not currently supported");
```
mullermp, line 355: I think we prefer to keep this error. Resource shapes would never belong in types.rb.

mullermp (commented):
Nice! Thank you. What are the overall benefits of this?

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/RubySymbolProvider.java
AndrewFossAWS: > Nice! Thank you. What are the overall benefits of this?

I think it's simply adhering to the Smithy based code generators pattern. But in terms of performance/functionality, it still the same for now as it re-uses the existing code generators.

---

Keeping the error would cause code generation to fail as directed codegen tries to call `SymbolProvider.resourceShape(...)` behind the scene whenever it encounters a ResourceShape.

I have removed `definitionFile` for now until we have a better idea on ResourceShape handling. (Ruby codgen currently doesn't handle/write ResourceShapes)

mullermp (approved):

PR #113, opened 2022-10-26
Refactor SymbolProvider usages

AndrewFossAWS:
*Description of changes:*

When using directed codegen, a single instance of SymbolProvider is created. Behind the scene, Directed codgen caches symbols returned by the SymbolProvider instance (see [here](https://github.com/awslabs/smithy/blob/main/smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/directed/CodegenDirector.java#L360-L370) and [here](https://github.com/awslabs/smithy/blob/main/smithy-codegen-core/src/main/java/software/amazon/smithy/codegen/core/CachingSymbolProvider.java)). 

We should avoid multiple instances of SymbolProvider if possible and take advantage of the caching mechanism.

This PR contains the following changes:
* Replace RubySymbolProvider initializations with `context.symbolProvider()` in code generators 
* Set default module to `Types` and definitionFile `types.rb` in SymbolProvider

Testing: Ran `./gradlew clean build` and confirm generated ruby code remains unchanged

### DIFF
--- a/codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/ApplicationTransport.java
+++ b/codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/ApplicationTransport.java
@@ -24,7 +24,6 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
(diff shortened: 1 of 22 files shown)


mullermp (approved):
Looks good thank you!!

PR #115, opened 2022-11-07
Validate Requredness + Default value

mullermp:
Validate required trait now, and remove logic in protocol builders that checks HTTP URI (should already be required).

Default + Required values should be populated by the client.

### DIFF
--- a/codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/ValidatorsGenerator.java
+++ b/codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/ValidatorsGenerator.java
@@ -106,7 +106,7 @@ public class ValidatorsGenerator extends ShapeVisitor.Default<Void> {
                 .write("")
                 .openBlock("class $L", symbolProvider.toSymbol(structureShape).getName())
                 .openBlock("def self.validate!(input, context:)")
-                .write("$T.validate!(input, $T, context: context)",
+                .write("$T.validate_types!(input, $T, context: context)",
                         Hearth.VALIDATOR,
                         context.symbolProvider().toSymbol(structureShape))
                 .call(() -> renderValidatorsForStructureMembers(members))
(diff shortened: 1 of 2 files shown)


codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/RestBuilderGeneratorBase.java
```diff
@@ -286,11 +286,6 @@ protected void renderUriBuilder(OperationShape operation, Shape inputShape) {
                                     + getter + ")"
                     );
                 }
-                writer
-                        .openBlock("if $1L.empty?", getter)
```
alextwoods, line 290: If someone has disabled validation, will we still skip validating required?  If so - would that cause this code to fail here with an unexpected error? (I think it may just result in an incorrect url being produced.  Either way, we *may* still want to keep these checks since we do allow configuration to disable validation).

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/ConfigGenerator.java
```diff
@@ -143,7 +143,7 @@ private void renderValidateMethod(List<ClientConfig> clientConfigList) {
             if (type.equals("Boolean")) {
                 type = "TrueClass, FalseClass";
             }
-            writer.write("$3T.validate!($1L, $2L, context: 'options[:$1L]')",
+            writer.write("$3T.validate_types!($1L, $2L, context: 'options[:$1L]')",
                     member, type, Hearth.VALIDATOR);
             // TODO - add constraints here
```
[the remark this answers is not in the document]
mullermp: I think so yeah.. we need constraints for string/int etc config.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/RestBuilderGeneratorBase.java
mullermp: Yes it would skip validating required. It would fail with an unexpected error (bad URL). That's the behavior today in V3 - `Aws::S3::Client.new(validate_params: false).abort_multipart_upload(key: 'foo')` will be `DELETE /foo` (missing uploadId). If we were guarding against everything in the URI, we'd want to check query params too. Personally I lean towards doing all or nothing with required and a way to configure that off.

alextwoods: I think the S3 case is a bit specific - as far as we know that is the only service where the presence of user input query parameters changes which operation is used so I dont think validating query param is the same as validating host label presence.    I'd personally lean towards keeping this check and not allowing any configuration that opts out of this.

But I'll defer to you on it if you want to remove it.

PR #117, opened 2022-11-09
Add default value population for members with `required` and `default` traits applied

AndrewFossAWS:
*Description of changes:*
* Add default value population for members with `required` and `default` traits applied
* Added mixins validation

### DIFF
--- a/codegen/projections/white_label/lib/white_label/builders.rb
+++ b/codegen/projections/white_label/lib/white_label/builders.rb
@@ -50,6 +50,12 @@ module WhiteLabel
     class MapOfStructs
     end
 
+    # Operation Builder for MixinTest
+    class MixinTest
+      def self.build(http_req, input:)
+      end
+    end
+
     # Operation Builder for PaginatorsTest
     class PaginatorsTest
       def self.build(http_req, input:)
--- a/codegen/projections/white_label/lib/white_label/builders.rb
+++ b/codegen/projections/white_label/lib/white_label/builders.rb
@@ -10,20 +10,6 @@
 module WhiteLabel
   module Builders
 
-    # Operation Builder for DefaultKitchenSink
-    class DefaultKitchenSink
-      def self.build(http_req, input:)
-      end
-    end
-
-    # Map Builder for MapOfStrings
-    class MapOfStrings
-    end
-
-    # List Builder for ListOfStrings
-    class ListOfStrings
-    end
-
     # Operation Builder for DefaultsTest
     class DefaultsTest
       def self.build(http_req, input:)
@@ -48,20 +34,20 @@ module WhiteLabel
       end
     end
 
-    # Structure Builder for Union
-    class Union
-    end
-
-    # Structure Builder for Struct
-    class Struct
-    end
-
-    # Map Builder for MapOfStructs
-    class MapOfStructs
+    # List Builder for ListOfStrings
+    class ListOfStrings
     end
 
     # List Builder for ListOfStructs
     class ListOfStructs
+    end
+
+    # Map Builder for MapOfStrings
+    class MapOfStrings
+    end
+
+    # Map Builder for MapOfStructs
+    class MapOfStructs
     end
 
     # Operation Builder for PaginatorsTest
@@ -91,6 +77,14 @@ module WhiteLabel
         http_req.body = input[:stream]
         http_req.headers['Content-Type'] = 'application/octet-stream'
       end
+    end
+
+    # Structure Builder for Struct
+    class Struct
+    end
+
+    # Structure Builder for Union
+    class Union
     end
 
     # Operation Builder for WaitersTest
--- a/codegen/projections/white_label/lib/white_label/builders.rb
+++ b/codegen/projections/white_label/lib/white_label/builders.rb
@@ -10,6 +10,20 @@
 module WhiteLabel
   module Builders
 
+    # Operation Builder for DefaultKitchenSink
+    class DefaultKitchenSink
+      def self.build(http_req, input:)
+      end
+    end
+
+    # Map Builder for MapOfStrings
+    class MapOfStrings
+    end
+
+    # List Builder for ListOfStrings
+    class ListOfStrings
+    end
+
     # Operation Builder for DefaultsTest
     class DefaultsTest
       def self.build(http_req, input:)
@@ -46,16 +60,8 @@ module WhiteLabel
     class MapOfStructs
     end
 
-    # Map Builder for MapOfStrings
-    class MapOfStrings
-    end
-
     # List Builder for ListOfStructs
     class ListOfStructs
-    end
-
-    # List Builder for ListOfStrings
-    class ListOfStrings
     end
 
     # Operation Builder for PaginatorsTest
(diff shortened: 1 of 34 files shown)


codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/ParamsGenerator.java
```diff
@@ -318,6 +359,19 @@ public Void unionShape(UnionShape shape) {
...
+            if (defaultValue.isPresent()) {
+                if (checkRequired) {
+                    writer.write("$1Lparams.key?($2L) ? $3L.build($4L, context: $5L) : $6L",
+                            memberSetter, rubySymbol, symbolProvider.toSymbol(shape).getName(), input, context,
+                            defaultValue.get());
+                } else {
+                    writer.write("$1L(params.key?($2L) ? $3L.build($4L, context: $5L) : $6L)",
+                            memberSetter, rubySymbol, symbolProvider.toSymbol(shape).getName(), input, context,
+                            defaultValue.get());
+                }
```
mullermp, line 371: For complex types, like if the default of a map is `{}`, I would think we still want to do param building for that complex type, but the VALUE inside of .build() would be `{}` instead of `params[:map]` or w/e the member is called.

codegen/smithy-ruby-codegen-test/model/weather.smithy
```diff
@@ -424,7 +424,7 @@ structure GetCityImageInput {
     @required
     imageType: ImageType,
 
-    @required,
```
mullermp, line 427: Good catch! I think you'll want to check Smithy's copy..

mullermp (commented):
Nice work! It's pretty close. I think the complex part is not correct. We'll want to probably want to modify white label's model and maybe write some integ specs...

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/ParamsGenerator.java
AndrewFossAWS: fixed

codegen/projections/white_label/lib/white_label/builders.rb
```diff
@@ -10,6 +10,20 @@
 module WhiteLabel
   module Builders
 
+    # Operation Builder for DefaultKitchenSink
```
mullermp, line 13: These should be alphabetical too. Not sure if that broke during this PR or we had a weird ordering to begin with.

codegen/projections/white_label/lib/white_label/parsers.rb
```diff
@@ -10,6 +10,20 @@
 module WhiteLabel
   module Parsers
 
+    # Operation Parser for DefaultKitchenSink
```
mullermp, line 13: Same comment wrt alphabetical

codegen/projections/white_label/lib/white_label/stubs.rb
```diff
@@ -10,6 +10,70 @@
 module WhiteLabel
   module Stubs
 
+    # Operation Stubber for DefaultKitchenSink
+    class DefaultKitchenSink
```
mullermp, line 14: Same comment wrt alphabetical

codegen/projections/white_label/spec/endpoints_spec.rb
```diff
@@ -18,14 +18,6 @@ module WhiteLabel
     describe '#endpoint_with_host_label_operation' do
       let(:label) { 'input_label' }
 
-      it 'raises when missing host label member' do
```
mullermp, line 21: This test needs to come back. I think I deleted this when I deleted the raising functionality, but Alex and I talked about keeping it in case someone uses `validate_input: false` (we still want to validate URI members). Would you be able to revive it in this PR?

codegen/smithy-ruby-codegen-test/model/component-test/main.smithy
```diff
@@ -96,3 +97,62 @@ structure ClientError {
...
 structure ServerError {}
+
+@http(method: "POST", uri: "/default_kitchen_sink")
+operation DefaultKitchenSink {
+    input: DefaultKitchenSinkInputOutput,
+    output: DefaultKitchenSinkInputOutput,
+    errors: [],
+}
+
+structure DefaultKitchenSinkInputOutput {
```
mullermp, line 108: Good idea. Do we also test defaults without required? I would think we need to do that, but for at least a few types of members like one simple type (string), one complex type (structure).

mullermp (commented):
Nice! Mostly things got out of order, and there was a test deleted by mistake (I think that was my bad). I think we'll want a defaults test without required. Otherwise looks good.

codegen/smithy-ruby-codegen-test/model/component-test/main.smithy
```diff
@@ -96,3 +97,62 @@ structure ClientError {
 @error("server")
 @retryable(throttling: true)
 structure ServerError {}
+
+@http(method: "POST", uri: "/default_kitchen_sink")
```
mullermp, line 101: I think a better home for this might be in component-test/defaults.smithy. Because we could put the default + non required test in there too.

codegen/projections/white_label/lib/white_label/builders.rb
AndrewFossAWS: fixed

codegen/projections/white_label/lib/white_label/parsers.rb
AndrewFossAWS: fixed

codegen/projections/white_label/lib/white_label/stubs.rb
AndrewFossAWS: fixed

codegen/smithy-ruby-codegen-test/model/component-test/main.smithy
AndrewFossAWS: > Do we also test defaults without required?
Yes, the `parsm_spec` verifies the member (without default + required) doesn't exist for an empty params.

codegen/smithy-ruby-codegen-test/model/component-test/main.smithy
AndrewFossAWS: fixed

codegen/smithy-ruby-codegen-test/model/component-test/main.smithy
mullermp: What I wanted to check was a member with `@default` trait that is NOT `@required`, and to verify that we DON'T use default values for it.

mullermp: I'm not sure I follow.. can you show me?

codegen/projections/white_label/spec/types_spec.rb
```diff
@@ -61,15 +61,11 @@ module Types
     describe DefaultsTestInput do
```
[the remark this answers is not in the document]
AndrewFossAWS: As discussed in slack, i will remove the default value population in types in the follow up PR. (All defaults should be populated in params)

mullermp (approved):

PR #120, opened 2023-03-14
Mark Builders, Params, Parsers, Stubs and Validators with api private

AndrewFossAWS:
*Description of changes:*
Mark Builders, Params, Parsers, Stubs and Validators with api private


codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/RubyCodeWriter.java
```diff
@@ -92,6 +92,11 @@ public void closeAllModules() {
         }
     }
 
+    public RubyCodeWriter markApiPrivate() {
```
mullermp, line 95: I would just call this "apiPrivate"

mullermp (approved):

PR #121, opened 2023-03-31
Checksum classes

alextwoods:
*Description of changes:*
Adds implementations of all of the supported Checksum algorithms as Digest like classes.  These are all basically just thin wrappers, but give us concrete classes to make future changes on if needed.


hearth/lib/hearth.rb
```diff
@@ -27,6 +27,21 @@
 require_relative 'hearth/waiters/waiter'
 require_relative 'hearth/xml'
 
+# Core SDK library
 module Hearth
   VERSION = File.read(File.expand_path('../VERSION', __dir__)).strip
+
+  # @return true if CRT is available
+  def self.use_crt?
```
mullermp, line 35: I understand why this is here but I don't know if hearth should have an AWS dependency. I am thinking crc22c should be an SDK core provided class somehow.

hearth/lib/hearth/checksums.rb
```diff
@@ -11,20 +19,38 @@ module Checksums
...
 
-          md5.update(chunk)
+          digest.update(chunk)
         end
         value.rewind
-        md5.base64digest
+        digest.base64digest
       else
-        OpenSSL::Digest.new('MD5').base64digest(value)
+        digest.tap { |d| d.update(value) }.base64digest
```
mullermp, line 33: Can we have a base64digest method on digest like before?

mullermp (commented):
Looks good overall. Nice. Haven't looked thoroughly at tests yet.

hearth/lib/hearth/checksums.rb
alextwoods: You mean allow it to take a value and both update and return it?

We technically do, because it just aliases the underlying method.  I was using the SRA defined interface here, but we could just do `d.base64digest(value)`

hearth/lib/hearth.rb
alextwoods: I dont really see a good way to move only this to AWS.  I might lean towards moving ALL of checksums to AWS (since they will only be used from there) and just leaving the existing content_md5 middleware alone here.

alextwoods:
Moved checksum classes to AWS.  Closing in favor of: https://github.com/alextwoods/aws-sdk-ruby-v4/pull/30

hearth/lib/hearth.rb
mullermp: That's better

hearth/lib/hearth/checksums.rb
mullermp: Yes I would prefer that

PR #122, opened 2023-04-05
[WIP] SRA: Request/Responses with Field/Fields

mullermp:
Implement request/responses SRA.

Tests and codegen are still WIP.


hearth/lib/hearth/http/client.rb
```diff
@@ -115,7 +114,8 @@ def configure_ssl(http)
       # @return [Net::HTTP::Request]
       def build_net_request(request)
         request_class = net_http_request_class(request)
-        req = request_class.new(request.url, request.headers.to_h)
+        # Trailers are not supported in Net::HTTP, but get all fields anyway.
```
alextwoods, line 117: I think we should probably only get headers and then fail if there are any trails

hearth/lib/hearth/http/client.rb
```diff
@@ -128,8 +128,8 @@ def build_net_request(request)
       end
 
       # @param [Net::HTTP::Response] response
-      # @return [Hash<String, String>]
-      def extract_headers(response)
+      # @return [Hash<String, Field>]
```
alextwoods, line 131: This isn't really a Field class at this point is it?  This is just going to be a string from net:http's response right?  I'd argue for keeping this as extract_headers, since thats what its doing.  `Field` is a concept on Hearth (smithy) responses, not on Net http responses

hearth/lib/hearth/http/field.rb
```diff
@@ -0,0 +1,69 @@
...
+      end
+
+      # Returns an escaped string representation of the field.
+      # @return [String]
+      def value
+        @values.compact.map { |v| escape_value(v) }.join(',')
+      end
+
+      # @return [Boolean]
+      def header?
```
alextwoods, line 53: add `trailer?` as well?

hearth/lib/hearth/http/fields.rb
```diff
@@ -0,0 +1,86 @@
...
+    class Fields
+      include Enumerable
+
+      # @param [Hash<String,Field>] fields
+      def initialize(fields = {}, encoding: 'utf-8')
+        @entries = {}
+        fields.each_pair do |key, value|
+          self[key] = value
+        end
+        @encoding = encoding
```
alextwoods, line 15: How do we actually use the encoding?  Do we need to apply the encoding at some point - either when taking a value in the  `[]=` method, or maybe more likely, in the `value` method of `Field`, eg:
```
def value
@values.compact.map { |v| escape_value(v) }.join(',').encode(@encoding)
end
```

hearth/lib/hearth/http/fields.rb
```diff
@@ -0,0 +1,86 @@
...
+
+      # @return [Enumerable<String,Field>]
+      def each_pair(&block)
+        @entries.each(&block)
+      end
+      alias each each_pair
+
+      # @return [Hash]
+      def to_hash
+        @entries.to_h { |_k, v| [v.name, v.value] }
```
alextwoods, line 71: could pass @encoding as an optional argument to value?

hearth/lib/hearth/http/request.rb
```diff
@@ -7,77 +7,75 @@ module Hearth
...
       # @param [String] http_method
-      # @param [String] url
-      # @param [Headers] headers
-      # @param [IO] body
-      def initialize(http_method: nil, url: nil, headers: Headers.new,
-                     body: StringIO.new)
+      # @param [Fields] fields
+      # @param (see Hearth::Request#initialize)
+      def initialize(uri:, http_method: nil, fields: Fields.new, **kwargs)
+        kwargs[:destination] = URI(uri)
```
alextwoods, line 16: This is a little weird with inheritance.  You cannot create an `HTTPRequest` with destination as a keyword, but uri is required.  And the type of destination changes depending on whether this is a parent or child class (and String and URI are not interchangable - so this means you can't use an HTTPRequest in place of a Request ie Liskov Substitution Principle) Feels like we're trying to have it both ways and I think its making it a bit clunkier.  

I'd lean towards either:
1.  Eliminate `uri` as an httpRequest concept and just stick with the SRA's `destination`.
2.  the SRA lists Request's `destination` as a URI - so do that rather than a String.  That simplifies this constructor a bit and then we can optionally take uri OR destination here.
3. just use uri for both cases (since again the SRA lists the type of destination in Request as URI).

hearth/lib/hearth/request.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+require 'stringio'
```
alextwoods, line 3: technically we don't require stringio here anymore since we aren't defaulting to it.

hearth/lib/hearth/http/field.rb
```diff
@@ -5,53 +5,36 @@ module HTTP
...
-
-      # Delete a value from the field.
-      # @param [String] value
-      def delete(value)
-        @values.delete(value)
-      end
-
       # Returns an escaped string representation of the field.
       # @return [String]
       def value(encoding = nil)
```
alextwoods, line 31: I think we should rename this (`value_str` or `formated` or whatever, I don't have strong feelings) and have an accessor for `value` (ie to be able to get the raw value).

mullermp: I kept value for now. We can do raw_value if we need it. I think it makes sense here.

hearth/lib/hearth/http/api_error.rb
```diff
@@ -7,16 +7,16 @@ module HTTP
...
         @http_body = http_resp.body
         super(**kwargs)
       end
 
       # @return [Integer]
       attr_reader :http_status
 
-      # @return [Hash<String, String>]
-      attr_reader :http_headers
+      # @return [Hash<String, Fields>]
```
alextwoods, line 18: should the type here just be `Fields`?

hearth/lib/hearth/http/error_parser.rb
```diff
@@ -91,7 +91,7 @@ def generic_error(error_opts)
         http_resp = error_opts[:http_resp]
         case http_resp.status
         when HTTP_3XX then @error_module::ApiRedirectError.new(
-          location: http_resp.headers['location'], **error_opts
+          location: http_resp.fields['location'], **error_opts
```
alextwoods, line 94: This should still be `headers` right? (Or the `Field` needs to be converted to a string?)

hearth/lib/hearth/http/field.rb
```diff
@@ -0,0 +1,65 @@
...
+  module HTTP
+    # Represents an HTTP field.
+    # @api private
+    class Field
+      # @param [String] name The name of the field.
+      # @param [Array|#to_s] value (nil) The values for the field. It can be any
+      #   object that responds to `#to_s` or an Array of objects that respond to
+      #  `#to_s`.
+      # @param [Symbol] kind The kind of field, either :header or :trailer.
+      #   Trailers are currently not supported by Net::HTTP.
```
alextwoods, line 13: This seems like a detail of Net::Http, not of `Field`.... but I guess since these are api private, its fine.

hearth/lib/hearth/http/middleware/content_length.rb
```diff
@@ -15,8 +15,8 @@ def initialize(app, _ = {})
         # @return [Output]
         def call(input, context)
           request = context.request
-          if request&.body.respond_to?(:size) &&
-             !request.headers.key?('Content-Length')
+          if request.body.respond_to?(:size) &&
+             !request.fields.key?('Content-Length')
```
alextwoods, line 19: switch back to headers here?

hearth/lib/hearth/http/middleware/content_md5.rb
```diff
@@ -15,7 +15,7 @@ def initialize(app, _ = {})
         # @return [Output]
         def call(input, context)
           request = context.request
-          unless request.headers.key?('Content-MD5')
+          unless request.fields.key?('Content-MD5')
```
alextwoods, line 18: Ditto here

hearth/spec/hearth/http/field_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+        end
+      end
+
+      it 'is immutable' do
+        expect { header.name = 'X-Header-2' }.to raise_error(NoMethodError)
+        expect { header.value = 'bar' }.to raise_error(NoMethodError)
+        expect { header.kind = :trailer }.to raise_error(NoMethodError)
+      end
+
+      describe '#value' do
```
alextwoods, line 29: Add a test for the encoding?

hearth/spec/hearth/http/middleware/content_md5_spec.rb
```diff
@@ -3,7 +3,7 @@
 module Hearth
   module HTTP
     module Middleware
-      describe ContentLength do
+      describe ContentMD5 do
```
alextwoods, line 6: haha, whoops.... good catch

alextwoods (approved):
Looks good - just minor comments!

hearth/lib/hearth/http/api_error.rb
mullermp: Yes

hearth/lib/hearth/http/error_parser.rb
mullermp: Good catch. Why didn't this fail in a test....

hearth/lib/hearth/http/field.rb
mullermp: I figure it was a reasonable callout. I can remove it.

hearth/spec/hearth/http/field_spec.rb
mullermp: Good catch, also added in Fields

PR #123, opened 2023-04-07
Ensure stubs write to the body instead of replacing it

alextwoods:
*Description of changes:*
Ensure that stub classes write to the body instead of replacing it.  This is important for use cases where the response body is wrapped to provide additional functionality (eg, checksum validation)


hearth/lib/hearth/middleware/send.rb
```diff
@@ -32,6 +32,9 @@ def call(input, context)
           stub = @stubs.next(context.operation_name)
           output = Output.new
           apply_stub(stub, input, context, output)
+          if context.response.body.respond_to?(:rewind)
```
mullermp, line 35: I think this should be the stubs job, to rewind itself if we are writing out a string io.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/HttpProtocolTestGenerator.java
```diff
@@ -332,7 +332,8 @@ private void renderResponseMiddleware(HttpResponseTestCase testCase) {
 
     private void renderResponseMiddlewareBody(Optional<String> body) {
         if (body.isPresent()) {
-            writer.write("response.body = StringIO.new('$L')", body.get());
+            writer.write("response.body.write('$L')", body.get());
```
mullermp, line 335: I prefer the `<<` method instead of write. It's cleaner and we can drop a set of parens. Thoughts?

hearth/lib/hearth/middleware/send.rb
alextwoods: Yeah - I'm not against pushing it to the stubs.  But it means we have to do it N places intstead of 1.  And things will be broken if we miss it in even one place.  

The second reason I opted for it here was to match the http client behavior (ie, its the responsibility of our http client to rewind the body, not the responsibility of the library that code depends on).

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/HttpProtocolTestGenerator.java
alextwoods: I prefer the `write` method.  Its the counter to `read` and in a lot of places we define our own IO like objects -  I don't want to have to remeber to define both a write and a << method.  Ruby obviously doesn't define an actual IO interface that we can count on though...

mullermp (approved):

PR #128, opened 2023-04-17
DNS host resolver and host address

mullermp:
DNS host address and resolver classes. To use them, we monkey patch IPSocket and TCPSocket similar to resolv-replace, but do so in a way that preserves existing implementation (or existing monkey patch..) by using a thread local variable containing the dns resolver class. The resolver will check ipv6 compatibility and return the ipv6 address if the machine is configured for it.

Note that this does NOT implement happy eyeballs. Happy eyeballs should ideally be implemented in Ruby's stdlib net::http. This approach should make use of that when those changes have been made.

Miscellaneous fixes are included to fields because I needed to make them for live testing on AWS.

Tests are WIP - would like opinions on approach first.


hearth/lib/hearth/dns/host_resolver.rb
```diff
@@ -0,0 +1,62 @@
...
+          options[:service],
+          options[:family],
+          options[:socktype] || :SOCK_STREAM,
+          options[:protocol],
+          options[:flags]
+        )
+      end
+
+      def ipv4_addr(addrinfo_list, options)
+        addr = addrinfo_list.find(&:ipv4?)
```
alextwoods, line 29: this is always going to find the first address right?  We want to return a different address each time this is called I thought?

hearth/lib/hearth/http/client.rb
```diff
@@ -162,3 +180,39 @@ def http_proxy_parts
...
+# globally. When opening an HTTP request, we will set a thread local variable
+# to enable custom DNS resolution, and then disable it after the request is
+# complete. When the thread local variable is not set, we will use the default
+# Ruby DNS resolution, which may be Resolv or the system resolver.
+
+# Patch IPSocket
+class << IPSocket
+  alias original_hearth_getaddress getaddress
+
+  def getaddress(host)
```
alextwoods, line 196: Looking at this code: https://github.com/ruby/ruby/blob/55c771c302f94f1d1d95bf41b42459b4d2d1c337/ext/socket/ipsocket.c#L97

It looks like socket's dns resolution will loop through all of the returned addresses and return the first "healthy" one (where we can create a successful socket connection) - I think.  With this method we're loosing out on that.  Is that a concern?

alextwoods (commented):
Overall approach looks good.

hearth/lib/hearth/dns/host_resolver.rb
```diff
@@ -0,0 +1,62 @@
...
+module Hearth
+  module DNS
+    # Resolve a host name to an address.
+    # @api private
+    class HostResolver
+      def resolve_address(options = {})
+        addrinfo_list = addrinfo(options)
+        ipv6 = ipv6_addr(addrinfo_list, options) if use_ipv6?
+        ipv4 = ipv4_addr(addrinfo_list, options)
+        [ipv6, ipv4]
```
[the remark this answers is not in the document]
mullermp: Yeah, it was originally compact, but in the monkey patch, I check presence. I think it makes sense to NOT have compact here, that way we know the first one is always ipv6 if present.

hearth/lib/hearth/dns/host_resolver.rb
mullermp: getaddrinfo will return new sets of addresses each time. Taking the first is recommended.

hearth/lib/hearth/dns/host_resolver.rb
```diff
@@ -0,0 +1,62 @@
...
+          address: addr.ip_address,
+          hostname: options[:nodename],
+          service: options[:service]
+        )
+      end
+
+      def use_ipv6?
+        begin
+          list = Socket.ip_address_list
+        rescue NotImplementedError
```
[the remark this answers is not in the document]
mullermp: Possibly? But maybe IPv6 is turned on while the application is running? This method doesn't seem too impacting. I think it looks at local addresses only.

hearth/lib/hearth/http/client.rb
mullermp: Yeah, we resolve the address and then it's passed into that method. The IP address will return itself, giving a single item to loop through and use. I guess it's a concern, but the requirement is that we need DNS control at the application layer. Unless you see a better option?

hearth/lib/hearth/dns/host_resolver.rb
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
+
+module Hearth
+  module DNS
+    # Resolve a host name to an address.
+    # @api private
+    class HostResolver
+      def resolve_address(options = {})
+        addrinfo_list = addrinfo(options)
+        ipv6 = ipv6_addr(addrinfo_list, options) if use_ipv6?
```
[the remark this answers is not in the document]
mullermp: I think that can make sense.

hearth/lib/hearth/http/client.rb
```diff
@@ -162,3 +180,39 @@ def http_proxy_parts
     end
   end
 end
+
+# These patches are based on resolv-replace
```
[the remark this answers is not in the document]
mullermp: I believe this patch still works. If resolv-replace is required before loading this, then the alias I created will be a reference to the already patched version. I can double check this.

hearth/lib/hearth/dns/host_resolver.rb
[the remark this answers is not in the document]
alextwoods: yea fair - rescuing an exception isn't as slow as trying to require a file.

hearth/lib/hearth/http/client.rb
```diff
@@ -36,15 +36,24 @@ class Client
...
         @http_wire_trace = options[:http_wire_trace]
         @logger = options[:logger]
-        @http_proxy = options[:http_proxy]
-        @http_proxy = URI.parse(@http_proxy.to_s) if @http_proxy
+        @http_proxy = URI(options[:http_proxy]) if options[:http_proxy]
         @ssl_verify_peer = options[:ssl_verify_peer]
         @ssl_ca_bundle = options[:ssl_ca_bundle]
         @ssl_ca_directory = options[:ssl_ca_directory]
         @ssl_ca_store = options[:ssl_ca_store]
+        @dns_resolver = options[:dns_resolver] || Hearth::DNS::HostResolver.new
```
alextwoods, line 56: What about defaulting to nil and not replacing the default?  We can still allow it to be overwritten and can provide the interface and an implementation, but not change it by default?

hearth/lib/hearth/http/client.rb
alextwoods: No - I don't really see any better options - I think its reasonable for what we need to do.

hearth/lib/hearth/http/client.rb
mullermp: That might be a good option..

hearth/lib/hearth/dns/host_resolver.rb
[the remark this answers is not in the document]
mullermp: Actually, I think the opposite now.. DNS resolution may not want ipv6 or even not want ipv4.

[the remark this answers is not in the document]
mullermp: I've kept it the way it is. Because HostResolver implementation (getaddrinfo) can dictate family, i.e. INET or INET6, I think it makes sense to have this class also be aware of whether it's supported or not.

hearth/lib/hearth.rb
```diff
@@ -7,6 +7,7 @@
 require_relative 'hearth/config/env_provider'
 require_relative 'hearth/config/resolver'
 require_relative 'hearth/context'
+require_relative 'hearth/dns'
```
alextwoods, line 10: RE: moving the patch out to here - I do think its somewhat net:http specific - I'd lean towards keeping it in our net:http client. (It can stay in the hearth/dns file, but maybe don't require it from here, require it from the client?).  Feel free to ignore - just a suggestion.

hearth/spec/hearth/dns/host_address_spec.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module Hearth
+  module DNS
+    describe HostAddress do
+      it 'can be initialized' do
```
alextwoods, line 6: Add tests for the readers?

hearth/lib/hearth/dns/host_resolver.rb
```diff
@@ -0,0 +1,89 @@
+# frozen_string_literal: true
+
+module Hearth
+  module DNS
+    # Resolves a host name and service to an IP address. Can be used with
+    # {Hearth::HTTP::Client} host_resolver option.
```
alextwoods, line 6: Worth linking to AddrInfo's docs for more info on the options?

alextwoods (approved):
LGTM - minor comments.

hearth/spec/hearth/dns/host_address_spec.rb
mullermp: In Hearth we've ignored/removed tests for attr_accessors. It's indirectly tested in HostResolver I think.

hearth/lib/hearth.rb
mullermp: Yeah. It's a good call out. If we add a new HTTP client, using Sockets, or even Typhoeus, curb, etc, we'd probably want this patch for this too. I think separate might make sense..

PR #130, opened 2023-04-20
Add support for specifying relative middleware ordering

alextwoods:
*Description of changes:*

Add relative middleware ordering.

Middleware may EITHER set an explicit "order" (with 0 being default) OR set a `Relative` and specify the ruby class. Middleware may only specify ONE relative ordering constraint.

When rendering the middleware stack for each operation, we first apply the service/operation predicate to get a set of all middleware for the stack, then we recursively attempt to resolve the order.  If a middleware was a Relative, we resolve the order for that middleware.  If it does not, we use the given order (which defaults to 0).  

In java code this looks like:
```java
        // Add two demo relative middlewares
        Middleware a = (new Middleware.Builder())
                .klass("Hearth::Middleware::Middleware1")
                .step(MiddlewareStackStep.SERIALIZE)
                .relative(new Middleware.Relative(Middleware.Relative.Type.BEFORE, "Hearth::Middleware::Middleware2"))
                .build();

        Middleware b = (new Middleware.Builder())
                .klass("Hearth::Middleware::Middleware2")
                .step(MiddlewareStackStep.SERIALIZE)
                .relative(new Middleware.Relative(Middleware.Relative.Type.AFTER, "Hearth::Middleware::Build"))
                .build();
```

This produces:
```ruby
stack.use(Hearth::Middleware::Validate,
        validator: Validators::CreateHighScoreInput,
        validate_input: @config.validate_input
      )
      stack.use(Hearth::Middleware::Middleware1)
      stack.use(Hearth::Middleware::Build,
        builder: Builders::CreateHighScore
      ) 
      stack.use(Hearth::Middleware::Middleware2)
      stack.use(Hearth::HTTP::Middleware::ContentLength)
```

There are a few error cases that will result in errors during codegen time:
* circular dependencies: These will result in a stack overflow.  Here we catch the error and raise a message.
* specifying a Relative to that doesn't exist (ie, it may have gotten filtered for a given operation) - here was also raise an error w/ a message.

There is one weakness - you can only specify one relative ordering and it doesn't guarantee the rest of the order.
eg:
A before B
B after C
This could produce:
A -> C -> B
C -> A -> B
(ie, either is perfectly valid, but A may care about its relation to both C and B).

### DIFF
--- a/codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/middleware/MiddlewareBuilder.java
+++ b/codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/middleware/MiddlewareBuilder.java
@@ -23,6 +23,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 import software.amazon.smithy.codegen.core.SymbolProvider;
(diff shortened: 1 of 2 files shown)


codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/middleware/MiddlewareBuilder.java
```diff
@@ -100,18 +101,65 @@ public void render(RubyCodeWriter writer, GenerationContext context,
...
+                .stream().filter((m) -> m.includeFor(model, service, operation))
+                .collect(Collectors.toList());
+
+        filteredMiddleware.forEach((m) -> klassToMiddlewareMap.put(m.getKlass(), m));
+
+        try {
+            for (Middleware middleware : filteredMiddleware) {
+                resolve(middleware, resolved, order, klassToMiddlewareMap);
+            }
+        } catch (StackOverflowError e) {
```
mullermp, line 129: It would be nice to detect a cycle intelligently without rescuing a stack overflow / out of memory

alextwoods: Yeah good idea - pushed a change to detect.

mullermp (approved):

PR #131, opened 2023-04-21
Fix bug in handling of renamed shapes

alextwoods:
*Description of changes:*
Fixes a bug in the handling or renamed shapes.  The TreeSet used to store/sort the shapes we generate (eg in Builders) had a comparator that only considered the shape's name.  When a shape has a duplicate name (but is renamed in the service), it looks like a duplicate and isn't added to the set.  

This change adds the shapeId to the end of the string used for comparison (with a space to ensure that sort order is maintained).

### DIFF
--- a/codegen/smithy-ruby-rails-codegen-test/model/protocol-test/unions.smithy
+++ b/codegen/smithy-ruby-rails-codegen-test/model/protocol-test/unions.smithy
@@ -37,6 +37,10 @@ union MyUnion {
     listValue: StringList,
     mapValue: StringMap,
     structureValue: GreetingStruct,
+
+    // Note that this uses a conflicting structure name with
+    // GreetingStruct, so it must be renamed in the service.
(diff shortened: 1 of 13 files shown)


codegen/smithy-ruby-rails-codegen-test/model/protocol-test/unions.smithy
```diff
@@ -248,6 +252,30 @@ apply JsonUnions @httpRequestTests([
             }
         }
     },
+    {
+        id: "RailsJsonSerializeRenamedStructureUnionValue",
```
mullermp, line 256: Is this something that we should contribute upstream?

mullermp (approved):

codegen/smithy-ruby-rails-codegen-test/model/protocol-test/unions.smithy
alextwoods: I ported these tests from the Smithy repo (I think they exist in most protocols).

PR #132, opened 2023-05-04
SRA: Retry Strategy classes

mullermp:
Implements Retry Strategy interface

### DIFF
--- a/codegen/projections/rails_json/lib/rails_json/config.rb
+++ b/codegen/projections/rails_json/lib/rails_json/config.rb
@@ -25,7 +25,9 @@ module RailsJson
   #     Logger to use for output
   #
   #   @option args [Hearth::Retry::Strategy] :retry_strategy (Hearth::Retry::Standard.new)
-  #     Specifies which retry strategu to use. Values are:
+  #     Specifies which retry strategy class to use. Strategy classes
+  #      may have additional options, such as max_retries and backoff strategies.
+  #      Available options are:
   #      * `Retry::Standard` - A standardized set of retry rules across the AWS SDKs. This includes support for retry quotas, which limit the number of unsuccessful retries a client can make.
   #      * `Retry::Adaptive` - An experimental retry mode that includes all the functionality of `standard` mode along with automatic client side throttling.  This is a provisional mode that may change behavior in the future.
   #
--- a/codegen/smithy-ruby-codegen-test/integration-specs/retry_spec.rb
+++ b/codegen/smithy-ruby-codegen-test/integration-specs/retry_spec.rb
@@ -15,8 +15,10 @@ module WhiteLabel
         # first return error, then some data
         client.stub_responses(:kitchen_sink, [error, { string: "ok" }])
 
+        expect_any_instance_of(Hearth::Retry::Standard).to receive(:acquire_initial_retry_token).and_call_original
+        expect_any_instance_of(Hearth::Retry::Standard).to receive(:refresh_retry_token).and_call_original
+        expect_any_instance_of(Hearth::Retry::Standard).to receive(:record_success).and_call_original
         expect(Kernel).to receive(:sleep).once
-        expect_any_instance_of(Hearth::Middleware::Retry).to receive(:call).and_call_original
 
         client.kitchen_sink
       end
--- a/codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/ApplicationTransport.java
+++ b/codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/ApplicationTransport.java
@@ -258,7 +258,7 @@ public final class ApplicationTransport {
      * @return the error inspector used for HTTP errors.
      */
     public String getErrorInspector() {
-        return "Hearth::HTTP::ErrorInspector";
+        return Hearth.HTTP_ERROR_INSPECTOR.toString();
     }
 
     /**
--- a/codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/middleware/MiddlewareBuilder.java
+++ b/codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/middleware/MiddlewareBuilder.java
@@ -236,7 +236,9 @@ public class MiddlewareBuilder {
                 .documentationDefaultValue("Hearth::Retry::Standard.new")
                 .defaultValue("Hearth::Retry::Standard.new")
                 .documentation(
-                        "Specifies which retry strategu to use. Values are: \n"
+                        "Specifies which retry strategy class to use. Strategy classes\n"
+                                + " may have additional options, such as max_retries and backoff strategies.\n"
+                                + " Available options are: \n"
                                 + " * `Retry::Standard` - A standardized set of retry rules across the AWS SDKs. "
                                 + "This includes support for retry quotas, which limit the number of"
                                 + " unsuccessful retries a client can make.\n"
--- a/codegen/projections/high_score_service/lib/high_score_service/config.rb
+++ b/codegen/projections/high_score_service/lib/high_score_service/config.rb
@@ -25,7 +25,9 @@ module HighScoreService
   #     Logger to use for output
   #
   #   @option args [Hearth::Retry::Strategy] :retry_strategy (Hearth::Retry::Standard.new)
-  #     Specifies which retry strategu to use. Values are:
+  #     Specifies which retry strategy class to use. Strategy classes
+  #      may have additional options, such as max_retries and backoff strategies.
+  #      Available options are:
   #      * `Retry::Standard` - A standardized set of retry rules across the AWS SDKs. This includes support for retry quotas, which limit the number of unsuccessful retries a client can make.
   #      * `Retry::Adaptive` - An experimental retry mode that includes all the functionality of `standard` mode along with automatic client side throttling.  This is a provisional mode that may change behavior in the future.
   #
--- a/codegen/projections/weather/lib/weather/config.rb
+++ b/codegen/projections/weather/lib/weather/config.rb
@@ -25,7 +25,9 @@ module Weather
   #     Logger to use for output
   #
   #   @option args [Hearth::Retry::Strategy] :retry_strategy (Hearth::Retry::Standard.new)
-  #     Specifies which retry strategu to use. Values are:
+  #     Specifies which retry strategy class to use. Strategy classes
+  #      may have additional options, such as max_retries and backoff strategies.
+  #      Available options are:
   #      * `Retry::Standard` - A standardized set of retry rules across the AWS SDKs. This includes support for retry quotas, which limit the number of unsuccessful retries a client can make.
   #      * `Retry::Adaptive` - An experimental retry mode that includes all the functionality of `standard` mode along with automatic client side throttling.  This is a provisional mode that may change behavior in the future.
   #
--- a/codegen/projections/white_label/lib/white_label/config.rb
+++ b/codegen/projections/white_label/lib/white_label/config.rb
@@ -25,7 +25,9 @@ module WhiteLabel
   #     Logger to use for output
   #
   #   @option args [Hearth::Retry::Strategy] :retry_strategy (Hearth::Retry::Standard.new)
-  #     Specifies which retry strategu to use. Values are:
+  #     Specifies which retry strategy class to use. Strategy classes
+  #      may have additional options, such as max_retries and backoff strategies.
+  #      Available options are:
   #      * `Retry::Standard` - A standardized set of retry rules across the AWS SDKs. This includes support for retry quotas, which limit the number of unsuccessful retries a client can make.
   #      * `Retry::Adaptive` - An experimental retry mode that includes all the functionality of `standard` mode along with automatic client side throttling.  This is a provisional mode that may change behavior in the future.
   #
--- a/codegen/projections/white_label/spec/retry_spec.rb
+++ b/codegen/projections/white_label/spec/retry_spec.rb
@@ -15,8 +15,10 @@ module WhiteLabel
         # first return error, then some data
         client.stub_responses(:kitchen_sink, [error, { string: "ok" }])
 
+        expect_any_instance_of(Hearth::Retry::Standard).to receive(:acquire_initial_retry_token).and_call_original
+        expect_any_instance_of(Hearth::Retry::Standard).to receive(:refresh_retry_token).and_call_original
+        expect_any_instance_of(Hearth::Retry::Standard).to receive(:record_success).and_call_original
         expect(Kernel).to receive(:sleep).once
-        expect_any_instance_of(Hearth::Middleware::Retry).to receive(:call).and_call_original
 
         client.kitchen_sink
       end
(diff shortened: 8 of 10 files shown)


hearth/lib/hearth/http/error_inspector.rb
```diff
@@ -0,0 +1,53 @@
+# frozen_string_literal: true
+
+require 'time'
+
+module Hearth
+  module HTTP
```
mullermp, line 6: Error inspector was moved to HTTP because it is protocol specific - it checks http response status and specific http error classes.

hearth/lib/hearth/http/error_inspector.rb
```diff
@@ -0,0 +1,53 @@
...
+        (500..599).cover?(@http_response.status)
+      end
+
+      def client?
+        (400..499).cover?(@http_response.status)
+      end
+
+      def hints
+        hints = {}
+        hints[:retry_after_hint] = retry_after_hint if retry_after_hint
```
mullermp, line 33: Deviates from SRA a bit - retry middleware will collect all hints and forward them to the strategy.

hearth/lib/hearth/http/response.rb
```diff
@@ -39,7 +39,7 @@ def reset
         @status = 0
         @reason = nil
         @fields.clear
-        @body.truncate(0)
+        @body.truncate(0) if @body.respond_to?(:truncate)
```
mullermp, line 42: Not sure about this one. Previous logic was checking truncate to determine if something is retryable. Now it's just always retryable, and we don't truncate if the body doesn't allow for it. But does this break?

hearth/lib/hearth/middleware/retry.rb
```diff
@@ -2,114 +2,93 @@
...
-      #   initial attempt.
-      # @param [Boolean] adaptive_retry_wait_to_fill Used only in `adaptive`
-      #   retry mode. When true, the request will sleep until there is
-      #   sufficient client side capacity to retry the request. When false,
-      #   the request will raise a `CapacityNotAvailableError` and will not
-      #   retry instead of sleeping.
-      def initialize(app, retry_mode:, max_attempts:,
-                     adaptive_retry_wait_to_fill:, error_inspector_class:,
-                     retry_quota:, client_rate_limiter:)
+      # @param [RetryStrategy] retry_strategy The retry strategy to use.
```
mullermp, line 10: Do we care to make an interface with not implemented errors?

hearth/lib/hearth/middleware/retry.rb
```diff
@@ -2,114 +2,93 @@
...
-        return output unless @capacity_amount.positive?
-
-        delay = [Kernel.rand * (2**@retries), MAX_BACKOFF].min
-        Kernel.sleep(delay)
-        retry_request(input, context, output)
+        token = @retry_strategy.acquire_initial_retry_token
+
+        raise 'Unable to make request, too many failures' if token.nil?
+
+        # TODO: determine if loop or nested function call is better here
```
mullermp, line 32: We used to invoke `call` over and over. Now it's a loop. A loop is less overhead. I'm open to either option.

hearth/lib/hearth/middleware/retry.rb
```diff
@@ -2,114 +2,93 @@
...
-        return if @retry_mode == 'standard'
-
-        # either fail fast or block until a token becomes available
-        # must be configurable
-        # need a maximum rate at which we can send requests (max_send_rate)
-        # is unset until a throttle is seen
-        @client_rate_limiter.token_bucket_acquire(
-          1, wait_to_fill: @adaptive_retry_wait_to_fill
-        )
+      def error_info(error, error_inspector)
```
mullermp, line 58: We could skip SRA error_info and instead use error_inspector and pass it into our strategy interface. I originally had it that way, and I think it might work out better, but I've done it this way for review first.

hearth/lib/hearth/middleware/retry.rb
```diff
@@ -2,114 +2,93 @@
...
-      # retry quota is updated if the request is successful (both modes)
-      def request_bookkeeping(output, error_inspector)
-        @retry_quota.release(@capacity_amount) unless output.error
-
-        return unless @retry_mode == 'adaptive'
-
-        @client_rate_limiter.update_sending_rate(
-          error_inspector.error_type == 'Throttling'
-        )
+      def error_type(error, error_inspector)
```
mullermp, line 65: Could instead belong back in error_inspector.

hearth/lib/hearth/middleware/retry.rb
```diff
@@ -2,114 +2,93 @@
...
 
-        error_inspector.retryable? &&
-          context.response.body.respond_to?(:truncate)
+      def modeled_throttling?(error)
+        modeled_retryable?(error) && error.throttling?
       end
 
-      def retry_request(input, context, output)
-        @retries += 1
+      # TODO: find a way to dup req/resp, to undo signing, rather than resetting?
```
mullermp, line 87: Not sure what's preferred here. We may want to keep a copy of the request before signing, and use that for retries. The object IDs might be different though. Right now we just reset everything ourselves.

hearth/lib/hearth/retry/retry_backoff_strategy.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module Hearth
+  module Retry
+    # @api private
+    # Computes the next backoff delay for a retry attempt.
+    class RetryBackoffStrategy
```
mullermp, line 7: I don't like this class. I'd rather have a "call" interface and make it a proc (or a class) like we have backoff currently.

hearth/lib/hearth/retry/retry_quota.rb
```diff
@@ -19,13 +19,14 @@ def initialize
       # Check if there is sufficient capacity to retry and return it.
       # If there is insufficient capacity, return 0
       # @return [Integer] The amount of capacity checked out
-      def checkout_capacity(error_inspector)
+      def checkout_capacity(error_type)
```
mullermp, line 22: Design doc says to take a response here, but we can't do that anymore if this is a detail of RetryStrategy. It could take error_inspector again if we do away with error_info.

hearth/lib/hearth/retry/standard_retry_strategy.rb
```diff
@@ -0,0 +1,55 @@
...
+
+      def refresh_retry_token(retry_token, error_info)
+        return unless retryable?(error_info)
+
+        return if @retries >= @max_attempts - 1
+
+        @capacity_amount = @retry_quota.checkout_capacity(error_info[:error_type])
+        return unless @capacity_amount.positive?
+
+        delay = error_info[:retry_after_hint]
```
mullermp, line 33: Make use of retry after (new behavior).

hearth/lib/hearth/retry/standard_retry_strategy.rb
```diff
@@ -0,0 +1,55 @@
...
+      end
+
+      def record_success(retry_token)
+        @retry_quota.release(@capacity_amount)
+        retry_token
+      end
+
+      private
+
+      def retryable?(error_info)
```
mullermp, line 48: Unsure about this method. The idea is that if it's modeled retryable or transient/throttling, it's ok to retry. Modeled retry is checked in the middleware and then set on error info. It did not make sense in error inspector if it were to be protocol specific.

hearth/lib/hearth/http/response.rb
mullermp: Undo this

hearth/lib/hearth/retry/adaptive_retry_strategy.rb
```diff
@@ -0,0 +1,73 @@
...
+        @retry_backoff_strategy = retry_backoff_strategy
+        @max_attempts = max_attempts
+        @wait_to_fill = wait_to_fill
+
+        # instance state
+        @client_rate_limiter = ClientRateLimiter.new
+        @retry_quota = RetryQuota.new
+        @capacity_amount = nil
+        # TODO: use token for this, and update testing
+        @retries = 0
```
mullermp, line 28: Move this to middleware

hearth/lib/hearth/middleware/retry.rb
alextwoods: Since there is no shared code and its ruby idiomatic to not have an "interface" class, lets skip it for now.

hearth/lib/hearth/middleware/retry.rb
alextwoods: Yeah - I think that makes sense to do.  Though I think it makes sense to change the name "error_inspector" to "error_info".  

I think we can then do away with the error_info method and move the error_type and modeled_retryable? onto the errorInfo class (yes, this will be transport specific.  Can be duplicated if needed in the future, or we can create a parent class and inherit... but I think it just makes sense as something an ErrorInfo should know).  

I think we leave the `retryable?` method in the stratagies though and allow them to check the type of error (I don't love that, but does give the strategy room to implement that differently I guess).

hearth/lib/hearth/middleware/retry.rb
```diff
@@ -2,114 +2,93 @@
...
-
-        @capacity_amount = @retry_quota.checkout_capacity(error_inspector)
-        return output unless @capacity_amount.positive?
-
-        delay = [Kernel.rand * (2**@retries), MAX_BACKOFF].min
-        Kernel.sleep(delay)
-        retry_request(input, context, output)
+        token = @retry_strategy.acquire_initial_retry_token
+
+        raise 'Unable to make request, too many failures' if token.nil?
```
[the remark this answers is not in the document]
alextwoods: This should probably be the same error I think.  And I dont see it getting raised anywhere - I think this is the only spot we raise.  

The documentation for adaptive does say we raise a CapacityNotAvailableError.  But we don't.

I think we should raise that error here (and update the docs for adaptive strategy).

hearth/lib/hearth/middleware/retry.rb
```diff
@@ -2,114 +2,63 @@
...
-      def retryable?(context, output, error_inspector)
-        return false unless output.error
-
-        error_inspector.retryable? &&
-          context.response.body.respond_to?(:truncate)
-      end
-
-      def retry_request(input, context, output)
+      def reset_request(context, output)
         @retries += 1
```
alextwoods, line 58: It doesn't feel like this belongs in "reset_request" (maybe just move it outside of that method).

hearth/lib/hearth/retry.rb
```diff
@@ -0,0 +1,13 @@
...
+
+require_relative 'retry/adaptive'
+require_relative 'retry/client_rate_limiter'
+require_relative 'retry/exponential_backoff'
+require_relative 'retry/retry_quota'
+require_relative 'retry/standard'
+
+module Hearth
+  module Retry
+    RetryToken = Struct.new(:retry_count, :retry_delay, keyword_init: true)
```
alextwoods, line 11: What about shortening this name to `Token`?  (since its in the Retry namespace already)

alextwoods (approved):

hearth/lib/hearth/middleware/retry.rb
mullermp: It was previously in retry_request. I can move it out. Ideally I think we shouldn't need it either..

hearth/lib/hearth/retry.rb
mullermp: Yeah I can do that

codegen/projections/rails_json/lib/rails_json/config.rb
```diff
@@ -27,23 +24,17 @@ module RailsJson
...
   #
-  #   @option args [Integer] :max_attempts (3)
-  #     An integer representing the maximum number of attempts that will be made for a single request, including the initial attempt.
-  #
-  #   @option args [String] :retry_mode ('standard')
-  #     Specifies which retry algorithm to use. Values are:
-  #      * `standard` - A standardized set of retry rules across the AWS SDKs. This includes support for retry quotas, which limit the number of unsuccessful retries a client can make.
-  #      * `adaptive` - An experimental retry mode that includes all the functionality of `standard` mode along with automatic client side throttling.  This is a provisional mode that may change behavior in the future.
+  #   @option args [Hearth::Retry::Strategy] :retry_strategy (Hearth::Retry::Standard.new)
+  #     Specifies which retry strategu to use. Values are:
```
alextwoods, line 28: Typo: `strategu`

codegen/smithy-ruby-codegen-test/integration-specs/retry_spec.rb
```diff
@@ -16,7 +16,7 @@ module WhiteLabel
         client.stub_responses(:kitchen_sink, [error, { string: "ok" }])
 
         expect(Kernel).to receive(:sleep).once
-        expect_any_instance_of(Hearth::Middleware::Retry).to receive(:call).twice.and_call_original
+        expect_any_instance_of(Hearth::Middleware::Retry).to receive(:call).and_call_original
```
alextwoods, line 19: This doesn't seem like a useful test anymore - can we assert something else?

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/middleware/MiddlewareBuilder.java
```diff
@@ -230,55 +230,28 @@ public void addDefaultMiddleware(GenerationContext context) {
...
-                .documentation(
-                        "Specifies which retry algorithm to use. Values are: \n"
-                                +
-                                " * `standard` - A standardized set of retry rules across the AWS SDKs. "
-                                +
-                                "This includes support"
-                                +
-                                " for retry quotas, which limit the number of unsuccessful retries a client can make.\n"
-                                + " * `adaptive` - An experimental retry mode that includes all the"
+                        "Specifies which retry strategu to use. Values are: \n"
```
alextwoods, line 239: Typo: `strategu`

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/middleware/MiddlewareBuilder.java
```diff
@@ -230,55 +230,28 @@ public void addDefaultMiddleware(GenerationContext context) {
...
+                                + " * `Retry::Standard` - A standardized set of retry rules across the AWS SDKs. "
```
alextwoods, line 240: Its a little confusing with "the values are" since these need to be instantiated.  Maybe "values are instances of: ".  And maybe mention that retry configuration such as max retries must be configured on the retry strategy.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/ApplicationTransport.java
```diff
@@ -254,6 +254,13 @@ public ClientFragment getTransportClient() {
         return transportClient;
     }
 
+    /**
+     * @return the error inspector used for HTTP errors.
+     */
+    public String getErrorInspector() {
+        return "Hearth::HTTP::ErrorInspector";
```
alextwoods, line 261: Add this to Hearth.java as a symbol

alextwoods (approved):

alextwoods (approved):

codegen/smithy-ruby-codegen-test/integration-specs/retry_spec.rb
mullermp: I tried to think of something, I'm not sure.

PR #133, opened 2023-06-06
Fix NetworkingError to be retryable

mullermp:
NetworkingError was not retryable because it was being raised by the HTTP client, but the retry logic was inspecting the output error. HTTP client will instead return an error or a response. Errors will be set on the output.


hearth/lib/hearth/middleware/send.rb
```diff
@@ -27,27 +27,29 @@ def initialize(_app, client:, stub_responses:,
...
           end
-          output
         else
-          @client.transmit(
+          resp_or_error = @client.transmit(
             request: context.request,
             response: context.response
           )
-          Output.new
+          if resp_or_error.is_a?(Hearth::NetworkingError)
```
mullermp, line 44: An alternative is rescuing here

alextwoods: Yeah I'd prefer rescuing here and having the http client throw the error rather than return either and error or response.

PR #134, opened 2023-06-08
Refactor http client and allow it on config

mullermp:
Pre-req for doing some client/connections in SRA - make http client a configurable option, and normalize the http option naming. Http client options will no longer be flat parameters on Config

### DIFF
--- a/hearth/lib/hearth/http/client.rb
+++ b/hearth/lib/hearth/http/client.rb
@@ -6,19 +6,34 @@ require 'openssl'
 
 module Hearth
   module HTTP
-    # @api private
+    # An HTTP client that uses Net::HTTP to send requests.
     class Client
       # Initialize an instance of this HTTP client.
       #
       # @param [Hash] options The options for this HTTP Client
       #
       # @option options [Boolean] :debug_output (false) When `true`,
-      #   HTTP debug output will be sent to the configured logger.
-      #
-      # @option options [Logger] :logger A logger where debug output is sent.
+      #   sets an output stream to the configured Logger for debugging.
       #
       # @option options [String] :proxy A proxy to send
       #   requests through. Formatted like 'http://proxy.com:123'.
+      #
+      # @option options [Float] :open_timeout Number of seconds to
+      #   wait for the connection to open.
+      #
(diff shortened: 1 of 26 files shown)


hearth/lib/hearth/http/client.rb
```diff
@@ -12,31 +12,31 @@ class Client
...
+      # @option options [String] :proxy A proxy to send
       #   requests through. Formatted like 'http://proxy.com:123'.
       #
-      # @option options [Boolean] :ssl_verify_peer (true) When `true`,
+      # @option options [Boolean] :verify_peer (true) When `true`,
       #   SSL peer certificates are verified when establishing a
       #   connection.
       #
-      # @option options [String] :ssl_ca_bundle Full path to the SSL
+      # @option options [String] :ca_file Full path to the SSL
```
alextwoods, line 27: the`ssl_` prefix feels a bit different than the `http` prefix - seems like these may still be useful in providing clarity on what its for?

hearth/lib/hearth/http/client.rb
```diff
@@ -47,22 +47,21 @@ class Client
...
       end
 
       # @param [Request] request
       # @param [Response] response
       # @return [Response]
-      def transmit(request:, response:)
+      def transmit(request:, response:, **options)
         http = create_http(request.uri)
-        http.set_debug_output(@logger) if @http_wire_trace
+        http.set_debug_output(options[:logger]) if @debug_output
```
alextwoods, line 64: What happens here if logger is nil?

alextwoods (approved):

hearth/lib/hearth/http/client.rb
mullermp: It shouldn't be, but I believe the method succeeds but no logging will occur.

hearth/lib/hearth/http/client.rb
mullermp: These were changed to exactly what Net::HTTP uses. I think the prefixes made sense when they were public configuration. Now Client is the only public config.

alextwoods (approved):
Updates LGTM.

PR #135, opened 2023-06-12
WIP Add basic connection pool

mullermp:
Adds a connection pool. It stores a map of pools as a class instance variable. It is keyed by config (a hash) and generic for any http client implementation. Each pool has an array of active connections, keyed by scheme+endpoint.


hearth/lib/hearth/connection_pool.rb
```diff
@@ -0,0 +1,103 @@
...
+
+module Hearth
+  # @api private
+  class ConnectionPool
+    @pools_mutex = Mutex.new
+    @pools = {}
+
+    class << self
+      # @return [ConnectionPool]
+      def for(config = {})
```
alextwoods, line 11: Why are `self.for` and `session_for` not combined as just `lease`?

hearth/lib/hearth/connection_pool.rb
```diff
@@ -0,0 +1,103 @@
...
+        @pools_mutex.synchronize do
+          puts "using connection pool for #{config}\n" if @pools[config]
+          puts "creating connection pool for #{config}\n" unless @pools[config]
+          @pools[config] ||= new
+        end
+      end
+
+      # @return [Array<ConnectionPool>] Returns a list of the
+      #   constructed connection pools.
+      def pools
```
alextwoods, line 21: Is there a reason to expose all the pools?

hearth/lib/hearth/connection_pool.rb
```diff
@@ -0,0 +1,103 @@
...
+        @pools_mutex.synchronize do
+          @pools.values
+        end
+      end
+    end
+
+    # @api private
+    def initialize
+      @pool_mutex = Mutex.new
+      @pool = {}
```
alextwoods, line 31: A ConnectionPool having a pool is a little weird to me - esp because a pool here is actually a map of endpoint to pool (so its multiple pools.  So a ConenctionPool has pools?)

hearth/lib/hearth/http/client.rb
```diff
@@ -81,28 +82,56 @@ def initialize(options = {})
...
+        # else
+        #   http.use_ssl = false
+        # end
+        endpoint = request.uri
+        pool = ConnectionPool.for(config)
+        session = pool.session_for(endpoint) do
+          start_session(endpoint)
         end
+        _transmit2(session, request, response)
+        pool.offer(endpoint, session)
```
alextwoods, line 100: Just confirming: this isn't done in a finally because we only `release` a connection when its successful right?

hearth/lib/hearth/connection_pool.rb
```diff
@@ -0,0 +1,103 @@
...
+    end
+
+    # @param [URI::HTTP, URI::HTTPS] endpoint The HTTP(S) endpoint
+    #    to connect to (e.g. 'https://domain.com').
+    # @param [Proc] block A block that returns a new session.
+    # @return [Object, nil]
+    def session_for(endpoint, &block)
```
alextwoods, line 38: nit - it feels weird for a connection pool to return `session`.  I'd lean towards using `connection` everywhere instead of session.

alextwoods (commented):
Overall looks good - I like the separation of the ConnectionPool from the http implementation.

hearth/lib/hearth/connection_pool.rb
```diff
@@ -0,0 +1,103 @@
...
+      session = nil
+      endpoint = remove_path_and_query(endpoint)
+      # attempt to recycle an already open session
+      @pool_mutex.synchronize do
+        _clean
+        session = @pool[endpoint].shift if @pool.key?(endpoint)
+        puts "using session from pool: #{session}\n" if session
+      end
+      session || block.call
```
alextwoods, line 47: Is there anything that keeps this from growing in an unbounded way?  (ie, a max connections or something like that).  Kicking off a ton of concurrent requests will create a ton of connections, which could be expensive

hearth/lib/hearth/connection_pool.rb
mullermp: It's used in empty_connection_pools! in v3. It's a public interface. We can start with it not included.

hearth/lib/hearth/http/client.rb
mullermp: We don't want to release/finish a connection when it's successful. It goes back into the pool. An ensure block would execute in any case.

hearth/lib/hearth/connection_pool.rb
mullermp: Connection is fine

hearth/lib/hearth/connection_pool.rb
mullermp: It's bounded by uniqueness of configuration and uniqueness of endpoint. I think practically speaking, it will not grow large, and any new request will clean old connections. The way the code is currently, a ton of concurrent requests to the same endpoint/service and with the same configuration will only have one connection which is re-used.

hearth/lib/hearth/connection_pool.rb
mullermp: pool is the internal data store of a ConnectionPool instance. The class method works across multiple pools, which makes sense IMO. I think it depends on how you define a pool. To me it's more than the array portion. It's every connection related to the client's configuration.

hearth/lib/hearth/connection_pool.rb
mullermp: I don't think we would want to create an instance of ConnectionPool because the pools need to maintain state across multiple clients potentially. So a lease method would need to be on the class, but then it would need to understand a particular instance. What would that look like?

hearth/lib/hearth/connection_pool.rb
alextwoods: I believe a large number of concurrent requests will each create a new endpoint.

And the number of endpoints may be very large - consider virtual hosting of s3 buckets - each bucket will end up with its own pool of connections.

hearth/lib/hearth/connection_pool.rb
alextwoods: Yeah, I'd argue for having it private and not exposing the pools publicly.

hearth/lib/hearth/http/client.rb
```diff
@@ -81,28 +82,56 @@ def initialize(options = {})
...
+        # http.set_debug_output(options[:logger]) if @debug_output
+        # configure_timeouts(http)
+        #
+        # if request.uri.scheme == 'https'
+        #   configure_ssl(http)
+        # else
+        #   http.use_ssl = false
+        # end
+        endpoint = request.uri
+        pool = ConnectionPool.for(config)
```
alextwoods, line 95: Rather than getting this pool each time we transmit, the `config` is per instance of the http client right?  Why not get the `pool` once?  That avoids getting two seperate mutex for each transmit.

hearth/lib/hearth/connection_pool.rb
alextwoods: I was suggesting a `self.lease` that would do somethin glike:
```
# skips mutex handling for brevity
def self.lease(config, endpoint, &block)
  pool = @pools[config] ||= new
  pool.session_for(endpoint, &block)
end
```

but with the optimization suggested below (moving getting the pool into the http client initialization, this doesn't make sense, so nvm)

hearth/lib/hearth/connection_pool.rb
alextwoods: To me its still just weird to have ConnectionPool class have a `pools` array of ConnectionPool.  Then each ConnectionPool has a hash or `pools` where the values are arrays.  `pools` is just being overloaded and its confusing to have `pools` on both the class and instance level to me.

hearth/lib/hearth/http/client.rb
mullermp: That could work

hearth/lib/hearth/connection_pool.rb
mullermp: Talked offline, it's not an issue in v3, inclined to leave it out for now.

hearth/lib/hearth/connection_pool.rb
mullermp: Moving pool into http client initialization would mean that it doesn't work for request level options

hearth/lib/hearth/connection_pool.rb
mullermp: From an active model-esque perspective, I think it's fairly idiomatic.. because ConnectionPool class methods interact with all pools, and instances interact with one pool. Each pool has 1 or many endpoints for one specific config. Multiple pools exist across other configs and potentially other http clients. I can see how it differs a bit if pool is just an array of connections.

hearth/lib/hearth/http/client.rb
mullermp: I've re-arranged things to make this easier to understand hopefully

mullermp:
Still needs tests if approach makes sense. Ignore puts statements, I'm still verifying connection re-use.

hearth/lib/hearth/connection_pool.rb
```diff
@@ -0,0 +1,92 @@
...
+    # @return [nil]
+    def offer(endpoint, connection)
+      endpoint = remove_path_and_query(endpoint)
+      @pool_mutex.synchronize do
+        @pool[endpoint] = [] unless @pool.key?(endpoint)
+        @pool[endpoint] << connection
+      end
+    end
+
+    # Closes and removes all sessions from the pool.
```
alextwoods, line 59: nit, sessions -> connections

alextwoods (approved):

hearth/spec/hearth/connection_pool_spec.rb
```diff
@@ -0,0 +1,151 @@
...
+
+    let(:endpoint) { URI('https://example.com') }
+    let(:endpoint2) { URI('https://example.org') }
+    let(:endpoint_path_query) { URI('https://example.com/path?query') }
+    let(:connection) { double('connection', stale?: false, finish: nil) }
+    let(:connection2) { double('connection2', stale?: false, finish: nil) }
+    let(:stale_connection) { double('stale_connection', stale?: true) }
+
+    shared_examples 'thread safety' do
+      it 'is thread safe' do
```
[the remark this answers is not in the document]
mullermp: It did once in many runs, but you're right, I don't think this is correct. I tried expecting Mutex too, but there's funkiness because there's two of them and they get nested. I think it's better to leave these tests out.

PR #137, opened 2023-06-26
Interceptors

alextwoods:
*Description of changes:*

Adds Interceptors to Hearth:
* Add interceptor context (and utilities for creating it from request context)
* Add Interceptor::List - similar to PluginList.  Also has an apply method for calling an interceptor hook (previously the `interceptor_hook` method).
* Add Initialize middleware (provide first/last hooks)
* Moved validate to a new middleware step (`VALIDATE`). Updated host_prefix to to use Relative to (before) Build.
* Codegen
  * Add Initialize middleware
  * Add interceptors to generated Configs
  * resolution during client initialization and operation invocation.

Future PRs:
* Convenience utilities for writing interceptors (TBD, needs some interface work) - will be done as separate PR.
* Remove middleware from public api (requires updates to protocol tests) - will be done as separate PR.


codegen/smithy-ruby-codegen-test/integration-specs/interceptor_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+require_relative 'spec_helper'
+
+module WhiteLabel
+  describe Client do
+    let(:test_interceptor) do
+      Class.new do
+        def read_before_execution(context); end
+      end.new
+    end
```
mullermp, line 11: newline

codegen/smithy-ruby-codegen-test/integration-specs/interceptor_spec.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+require_relative 'spec_helper'
+
+module WhiteLabel
+  describe Client do
+    let(:test_interceptor) do
```
mullermp, line 7: This creates an anonymous class right? I'm not sure what rspec and rubocop recommend - I thought it complains about this, but I've been doing it like this:

```
let(:interceptor_class) do
  class TestInterceptor
    # normal class stuff
  end
end
```

and then creating with `interceptor_class.new`

hearth/lib/hearth/interceptor.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+require_relative 'interceptor/context'
+require_relative 'interceptor/list'
```
mullermp, line 4: I don't think it makes sense to module-ize interceptors. I think like PluginList, this should be InterceptorList (flat) and also InterceptorContext.

hearth/lib/hearth/interceptor/context.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+module Hearth
+  module Interceptor
+    # Context provided to interceptor hooks methods.
```
mullermp, line 5: These should have real yardocs eventually

hearth/lib/hearth/context.rb
```diff
@@ -34,5 +36,18 @@ def initialize(options = {})
 
     # @return [Hash]
     attr_reader :metadata
+
+    # @return [Array] An ordered list of interceptors
+    attr_reader :interceptors
+
+    def interceptor_context(input, output)
```
mullermp, line 43: docs

hearth/lib/hearth/interceptor/list.rb
```diff
@@ -0,0 +1,91 @@
...
+      def set_output_error(last_error, output)
+        return unless last_error && output
+
+        output.data = nil
+        output.error = last_error
+      end
+
+      # TODO: What makes a valid interceptor? Options:
+      # 1. Check for responds_to for all known hooks
+      # 2. Require all interceptors to include an Interceptor module
```
mullermp, line 84: Could it be an inheritable Interceptor class that raises no method errors (interface)? If not defined, we could rescue/skip those?

hearth/lib/hearth/middleware/initialize.rb
```diff
@@ -0,0 +1,44 @@
...
+      end
+
+      # @param input
+      # @param context
+      # @return [Output]
+      def call(input, context)
+        # if there are exceptions, execution proceeds to before_completion hooks
+        before_execution_error = context.interceptors.apply(
+          hook: :read_before_execution, input: input,
+          context: context, output: nil, aggregate_errors: true
```
mullermp, line 21: nit: can we multiline all of these for readability?

hearth/lib/hearth/middleware/initialize.rb
```diff
@@ -0,0 +1,44 @@
...
+        )
+
+        output = if before_execution_error
```
mullermp, line 24: also a nit, If we're going all in on Ruby 3, we could use new rightward assignment syntax ... otherwise I think better style would be:
```
output =
  if before_execution_error
    # ...
  end
```

mullermp (commented):
Nice. Looks good so far.

codegen/smithy-ruby-codegen-test/integration-specs/interceptor_spec.rb
alextwoods: I believe anonymous classes are a preferred style.  See: https://github.com/rubocop/rspec-style-guide#declare-constants.

I think the code you had may leak the `TestInterceptor` name and should instead be:

```
let(:interceptor_class) do
  Class.new do
    # normal class stuff
  end
end
```

hearth/lib/hearth/interceptor/context.rb
alextwoods: Yeah - will spend time on documentation as a finishing touch once other parts are finialized.

hearth/lib/hearth/middleware/initialize.rb
alextwoods: Was trying to avoid these methods getting too much longer.... but I can multiline.

codegen/smithy-ruby-codegen-test/integration-specs/interceptor_spec.rb
mullermp: Ah ok. I think `Class.new('TestInterceptor')` makes sense, then interceptor_class.new in the test. I should see if we have other examples in Hearth that should probably be that way.

alextwoods: I think you don't want the name pass to `Class.new` - Thats usually used for extending a class.  Eg: `Class.new(ActiveRecord)` to extend ActiveRecord.

mullermp: Oh, I see. I thought that was for the class name. You're right.

mullermp: I do prefer `let()` always defining one thing (a class) and not a class and then an instance of it.

mullermp (approved):

hearth/lib/hearth/interceptor/hooks.rb
```diff
@@ -0,0 +1,38 @@
...
+      READ_AFTER_TRANSMIT = :read_after_transmit
+
+      MODIFY_BEFORE_DESERIALIZATION = :modify_before_deserialization
+      READ_BEFORE_DESERIALIZATION = :read_before_deserialization
+      READ_AFTER_DESERIALIZATION = :read_after_deserialization
+
+      MODIFY_BEFORE_COMPLETION = :modify_before_completion
+      READ_AFTER_EXECUTION = :read_after_execution
+
+      @hooks = constants.map { |c| const_get(c) }
```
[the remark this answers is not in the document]
alextwoods: Moved to a single list definition.

codegen/projections/white_label/lib/white_label/plugins/test_plugin.rb
```diff
@@ -1,12 +1,18 @@
 module WhiteLabel
   module Plugins
     class TestPlugin
+
+      TEST_CLASS_INTERCEPTOR = Class.new do
```
mullermp, line 5: Why is this a constant? I think you can just define a normal class here, and then call `config.interceptors = TestInterceptor.new`

codegen/projections/white_label/spec/interceptor_spec.rb
```diff
@@ -4,25 +4,89 @@
...
 
         client.kitchen_sink()
       end
     end
+
+    describe 'operation configured interceptor' do
+      it 'calls interceptor hook' do
+        config = Config.new(
+          stub_responses: true
+        )
```
mullermp, line 38: nit - single line!

codegen/projections/white_label/spec/interceptor_spec.rb
```diff
@@ -4,25 +4,89 @@
...
+        client = Client.new(config)
+
+        expect(interceptor).to receive(hook)
+                                 .with(instance_of(Hearth::InterceptorContext))
```
mullermp, line 42: RubyMine :(

codegen/projections/white_label/spec/interceptor_spec.rb
```diff
@@ -4,25 +4,89 @@
...
+          stub_responses: true,
+          plugins: Hearth::PluginList.new([config_plugin]),
+          interceptors: Hearth::InterceptorList.new([config_interceptor])
+        )
+      end
+
+      let(:client) { Client.new(config) }
+
+      it 'applies interceptors in expected order' do
+        expect(Plugins::TestPlugin::TEST_CLASS_INTERCEPTOR).to receive(hook).ordered
```
mullermp, line 79: I think you can use any_instance_of with the above change.

codegen/projections/white_label/spec/interceptor_spec.rb
```diff
@@ -4,25 +4,89 @@
...
+      # * Interceptors registered via operation-level plugins.
+      # * Interceptors registered via operation-level options (passed directly).
+
+      let(:config_plugin_interceptor) { interceptor_class.new('cfg_plg') }
+      let(:config_interceptor) { interceptor_class.new('cfg_int') }
+      let(:operation_plugin_interceptor) { interceptor_class.new('op_plg') }
+      let(:operation_interceptor) { interceptor_class.new('op') }
+
+      let(:config_plugin) do
+        proc { |cfg| cfg.interceptors << config_plugin_interceptor}
```
mullermp, line 61: Space before `}` on this and below - we should enable rubocop styling/auto correct on spec files.

codegen/projections/white_label/spec/interceptor_spec.rb
```diff
@@ -4,25 +4,89 @@
...
+        expect(Plugins::TestPlugin::TEST_CLASS_INTERCEPTOR).to receive(hook).ordered
+        expect(config_plugin_interceptor).to receive(hook).ordered
+        expect(config_interceptor).to receive(hook).ordered
+        expect(operation_plugin_interceptor).to receive(hook).ordered
+        expect(operation_interceptor).to receive(hook).ordered
+
+        client.kitchen_sink(
+          {},
+          plugins: [operation_plugin],
+          interceptors: [operation_interceptor])
```
mullermp, line 88: Not sure how rubocop isn't failing on the ending parens!

codegen/smithy-ruby-codegen-test-utils/build.gradle.kts
```diff
@@ -31,6 +31,9 @@ java {
 repositories {
     mavenLocal()
     mavenCentral()
+    flatDir {
+        dirs("middleware", "plugins")
```
mullermp, line 35: What's this for?

codegen/smithy-ruby-codegen-test-utils/plugins/test_plugin.rb
```diff
@@ -1,11 +1,17 @@
 module Plugins
   class TestPlugin
+
+    TEST_CLASS_INTERCEPTOR = Class.new do
```
mullermp, line 4: Ditto

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/middleware/MiddlewareBuilder.java
```diff
@@ -204,7 +210,10 @@ public void addDefaultMiddleware(GenerationContext context) {
 
         Middleware hostPrefix = (new Middleware.Builder())
                 .klass("Hearth::Middleware::HostPrefix")
-                .step(MiddlewareStackStep.INITIALIZE)
+                .step(MiddlewareStackStep.BUILD)
+                .relative(new Middleware.Relative(
+                        Middleware.Relative.Type.BEFORE, "Hearth::Middleware::Build")
```
mullermp, line 215: Is there a symbol for middleware here? Or do we have to pass "Hearth::Middleware::Build" as a string?

hearth/lib/hearth/context.rb
mullermp: Should this method be documented? Is it private?

hearth/lib/hearth/interceptor_context.rb
```diff
@@ -0,0 +1,30 @@
...
+      @request = request
+      @response = response
+      @output = output
+      @attributes = attributes
+    end
+
+    # Params object, eg Params::<Operation>Input
+    attr_reader :input
+
+    # Hearth::Request (eg HTTPRequest)
```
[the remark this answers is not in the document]
mullermp: These readers should have `@return` with appropriate types.

hearth/lib/hearth/interceptor_list.rb
```diff
@@ -0,0 +1,88 @@
...
+          break unless aggregate_errors
+        end
+      end
+
+      set_output_error(last_error, context, output)
+
+      last_error
+    end
+
+    def dup
```
mullermp, line 67: minimal `@return` docs for these

mullermp (approved):
LGTM. Just minor nit picky stuff

hearth/lib/hearth/context.rb
alextwoods: Added @api private - I dont think this should be considered part of our interface (though it needs to be public to be called from the middelware).

codegen/projections/white_label/lib/white_label/plugins/test_plugin.rb
alextwoods: I wanted to avoid using `any_instance_of` - this lets the test be a bit more specific.

codegen/projections/white_label/spec/interceptor_spec.rb
alextwoods: Wanted to be specific rather than use any_instance_of.

codegen/projections/white_label/spec/interceptor_spec.rb
alextwoods: ran rubocop on all of these.  But we should long term find a way to run it on all of these integration specs, but not on generated gems....

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/middleware/MiddlewareBuilder.java
alextwoods: We don't have symbols for any of these..... But we should.  I'll update all of our middleware to have symbol constants.

PR #139, opened 2023-06-30
Upgrade Smithy to 1.33.0

jterapin:
Upgrading the Smithy dependency to 1.33.0

Also made a minor fix so that we can successfully build this dependency with v4


alextwoods (approved):
LGTM!

PR #140, opened 2023-07-05
Add Config to Context

alextwoods:
*Description of changes:*

This adds config to context and ensures that the operation overrides are correctly applied to it.


mullermp (approved):
I think something like this makes sense. I think middleware should remove keyword args that use config (larger change) and it should only take keyword args for codegen classes or special unconfigurable things. Probably a hot debate.

PR #141, opened 2023-07-05
Add context to output

alextwoods:
*Description of changes:*
Add the context to output.


mullermp:
I really think we need this but I understand it's contentious. Especially with middleware as private, we need a way to inspect request context not just for testing but for debugging. I think context can be also response context so we want to make sure it round trips. It also I think shouldn't be documented on output, we should make it API private maybe. I'm not certain.

mullermp:
Probably should close this for now if we aren't doing config context.

PR #142, opened 2023-07-05
Service Integrations (specify integrations in smithy-build.json)

alextwoods:
*Description of changes:*
Allows integrations to be specified in the smithy-build.json using a new `integration` config.  This allows service specific integrations - but doesn't limit the flexibility of the `includeFor` for plugins that need to inspect the model (eg, integrations that check for presence of a certain trait on the model). 


codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/DirectedRubyCodegen.java
```diff
@@ -75,7 +75,8 @@ public GenerationContext createContext(CreateContextDirective<RubySettings, Ruby
         Model model = directive.model();
         List<RubyIntegration> integrations = directive.integrations().stream()
                 .filter((integration) -> integration
-                        .includeFor(service, model))
+                        .includeFor(service, model)
```
mullermp, line 78: I think it should be one or the other. If we still need the other approach then maybe we shouldn't do this one.

PR #143, opened 2023-07-11
Remove middleware from public interface

alextwoods:
*Description of changes:*
Interceptors are the public interface for injecting logic into the internals of SDK execution going forward.  We are removing middleware from the public interface so that there is only a single way to modify execution.  Middleware remains the internal framework that the SDK is built on. 


hearth/lib/hearth/http/response.rb
```diff
@@ -33,14 +33,28 @@ def initialize(status: 0, reason: nil, fields: Fields.new, **kwargs)
       # @return [Fields::Proxy]
       attr_reader :trailers
 
+      # Replace attributes from other response
+      # @param [Response] other
+      # @return [Response]
+      def replace(other)
```
mullermp, line 39: I think `update` might be the better name..

mullermp (approved):

PR #144, opened 2023-07-12
Request Compression

jterapin:
*Description of changes:*
Implementing Compression middleware - a middleware that compresses the request body and adds the `Content-Encoding` header


codegen/smithy-ruby-codegen-test/model/component-test/compression.smithy
```diff
@@ -0,0 +1,35 @@
+$version: "1.0"
+namespace smithy.ruby.tests
+
+@requestCompression(
+    encodings: ["gzip"]
+)
+
+@http(method: "POST", uri: "/some_operation")
+operation SomeOperation {
```
alextwoods, line 9: Rather than calling this SomeOperation we should make it a bit more specific/descriptive. Maybe requestCompressuonOperation or something - since there are a lot of operations in this model testing different features.

codegen/smithy-ruby-codegen-test/model/component-test/compression.smithy
```diff
@@ -0,0 +1,35 @@
...
+}
+
+@http(method: "POST", uri: "/some_streaming_operation")
+operation SomeStreamingOperation {
+    input: SomeStreamingInput
+}
+
+@input
+structure SomeOperationInput {
+    @httpPayload
```
alextwoods, line 20: Nice - this is the trait we will want to look for in the fakeProtocol builderGenerator and then put this member on the request body.

alextwoods (commented):
Looking good so far!

jterapin:
I'm failing the **Hearth CI / rubocop** check since line 20 is just two characters extra long. I don't think I can change the config naming as it is strictly outlined in the [Smithy Docs](https://smithy.io/2.0/spec/behavior-traits.html?highlight=compression#client-implementation). Am I able to exclude this specific file under rubocop's Layout/LineLength? LMK your thoughts. Thanks in advance for looking into the structure! 😄

alextwoods:
You can either rename the internal variable to be shorter or split the line after =

codegen/smithy-ruby-codegen-test/model/component-test/compression.smithy
```diff
@@ -0,0 +1,35 @@
...
+
+@http(method: "POST", uri: "/request_compress_streaming_operation")
+operation RequestCompressionStreamingOperation {
+    input: RequestCompressionStreamingInput
+}
+
+@input
+structure RequestCompressionInput {
+    @httpPayload
+    body: SomeBody
```
mullermp, line 21: I think a `Blob` shape is already defined somewhere - shouldn't need SomeBody

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/integrations/RequestCompression.java
```diff
@@ -0,0 +1,101 @@
...
+    @Override
+    public void modifyClientMiddleware(MiddlewareBuilder middlewareBuilder, GenerationContext context) {
+        ClientConfig disableRequestCompression = (new ClientConfig.Builder())
+                .name("disable_request_compression")
+                .type("Boolean")
+                .defaultValue("false")
+                .documentation("When set to 'true' the request body will not be compressed for supported operations.")
+                .allowOperationOverride()
+                .defaults(new ConfigProviderChain.Builder()
+                        .envProvider("DISABLE_REQUEST_COMPRESSION", "Boolean")
```
mullermp, line 49: This env is supposed to be prefixed by AWS_ but only in AWS SDKs. The compression trait doesn't mention environment variables: https://smithy.io/2.0/spec/behavior-traits.html so I would actually leave this off. I would also removed REQUEST_MIN_COMPRESSION_SIZE_BYTES too.

mullermp (commented):
Looks good so far.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/integrations/RequestCompression.java
alextwoods: I think they are specified here: https://smithy.io/2.0/spec/behavior-traits.html#compression-client-implementation

mullermp: Yeah. There's no ENV requirement. It says:

```
 In order of precedence from highest to lowest, the setting shall be exposed:

Per-request configuration (SHOULD, if request-level configuration is supported by that client)
Service client configuration - RequestMinCompressionSizeBytes (MUST, with idiomatic casing/spacing for that client programming language)
```

mullermp: I don't know why it reads "Smithy clients MUST expose a setting REQUEST_MIN_COMPRESSION_SIZE_BYTES" - it should probably read as RequestMinCompressionSizeBytes

codegen/smithy-ruby-codegen-test/model/component-test/compression.smithy
jterapin: Great, I think I found that one and another one for streaming `blob` - thanks!

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/integrations/RequestCompression.java
jterapin: Confirmed with `RequestCompression` author and dev on Smithy team that there's no ENV requirements for generic smithy clients.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/test/protocol/fakeprotocol/generators/BuilderGenerator.java
```diff
@@ -39,12 +42,24 @@ public BuilderGenerator(GenerationContext context) {
...
+        // checks for Payload member
+        Optional<MemberShape> httpPayloadMember = inputShape.members()
+                .stream()
+                .filter((m) -> m.hasTrait(HttpPayloadTrait.class))
+                .findFirst();
+        if (httpPayloadMember.isPresent()) {
+            if (Streaming.isStreaming(model, inputShape)) {
+                renderStreamingBodyBuilder(inputShape);
+            } else {
+                writer.write("http_req.body = StringIO.new(input[:$L] || '')",
```
alextwoods, line 55: We should add a comment that this will only work for String/Blob/Number data types which is sufficent for testing but not complete.  (the full solution woudl be to create a body serializer visitor here and provide implementations for all of the smithy types as we do in the RestBuilderGeneratorBase).

alextwoods (approved):
Nice - Looks good!!

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/test/protocol/fakeprotocol/generators/BuilderGenerator.java
mullermp: Are we coming back to do this later?

codegen/smithy-ruby-codegen-test/integration-specs/compression_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+          .to raise_error(
+            ArgumentError,
+            'Expected config[:disable_request_compression] to be in [TrueClass, FalseClass], got String.'
+          )
+      end
+    end
+
+    context 'request_min_compression_size_bytes' do
+      it 'raises error when given invalid integer' do
+        # TODO after we implement constraints on configs
```
mullermp, line 19: This should be merged now

codegen/smithy-ruby-codegen-test/integration-specs/compression_spec.rb
```diff
@@ -0,0 +1,83 @@
+# frozen_string_literal: true
```
mullermp, line 1: Can you run rubocop -A on this file?

mullermp (approved):

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/test/protocol/fakeprotocol/generators/BuilderGenerator.java
alextwoods: I dont think theres any reason for now to do so since this is just used as a minimal test protocol (ie, we don't actually serialize anything normally.... but its useful for testing some features.)

codegen/smithy-ruby-codegen-test/integration-specs/compression_spec.rb
alextwoods: I dont think we ever added the ability to do arbitrary constraints, such as enforcing the range of a number.

codegen/smithy-ruby-codegen-test/integration-specs/compression_spec.rb
jterapin: I actually haven't - doing it now, thanks for the call out!

codegen/smithy-ruby-codegen-test/integration-specs/compression_spec.rb
jterapin: My current plan is as follows:
- merge this PR now
- make a separate PR to address the constraints
- another PR for updating the middleware relative ordering
- a quick PR to wrap up testing

I would like to keep the PRs separately. They're all about different features and it'd probably be easier to review and adding each one independently.

PR #146, opened 2023-07-17
Stub improvements + other paper cuts

mullermp:
* Change HTTP client to raise networking error
* Refactor send.rb stub response logic
* Support Hearth::Structure
* Allow stubbing of error or data (interface change)
* Generated error stub classes
* Remove http pool empty! method


hearth/lib/hearth/middleware/send.rb
```diff
@@ -79,15 +67,31 @@ def call(input, context)
...
+
       def apply_stub(stub, input, context, output)
         case stub
         when Proc
           stub = stub.call(input, context)
           apply_stub(stub, input, context, output) if stub
-        when Exception
+        when Exception, ApiError
           output.error = stub
-        when Class
```
alextwoods, line 89: We should update the docuemtnation in client_stubs to match whatever we decide here (just a note for future us).

alextwoods (commented):
LGTM

mullermp:
Changes are not ready for review / production ready / no clean ups / no test changes. Just wanted feedback on approach.

Main highlights:
* Stubbing errors will only work for modeled errors.
* Protocol implementations will probably need some way to generate the http response necessary to parse into the error being stubbed. For Rails JSON, the status code is sufficient, so I didn't do this change here.
* Stubs must have data or error key. Data for normal case, error for error case. Error must be an array of 1 or 2 items. First is the class, second is any params.

```
irb(main):010:0> c.stub_responses(:update_high_score, { error: [HighScoreService::Errors::UnprocessableEntityError, { errors: { thing: ['foo', 'bar'] } } ] }, nil) 
=> [{:error=>[HighScoreService::Errors::UnprocessableEntityError, {:errors=>{:thing=>["foo", "bar"]}}]}, nil]
irb(main):011:0> c.update_high_score(id: '1')
/Users/mamuller/workplace/smithy-ruby/codegen/projections/high_score_service/lib/high_score_service/client.rb:395:in `update_high_score': UnprocessableEntityError (HighScoreService::Errors::UnprocessableEntityError)                                                                                                                                                              
        from (irb):11:in `<main>'                                                                                                                                                       
        from /Users/mamuller/.rubies/ruby-3.2.2/lib/ruby/gems/3.2.0/gems/irb-1.6.2/exe/irb:11:in `<top (required)>'                                                                     
        from /Users/mamuller/.rubies/ruby-3.2.2/bin/irb:25:in `load'                                                                                                                    
        from /Users/mamuller/.rubies/ruby-3.2.2/bin/irb:25:in `<main>'                                                                                                                  
irb(main):012:0> c.update_high_score(id: '1')
=> 
#<Hearth::Output:0x00000001109ff748
 @data=
  #<struct HighScoreService::Types::UpdateHighScoreOutput
   high_score=#<struct HighScoreService::Types::HighScoreAttributes id="id", game="game", score=1, created_at=2023-07-21 20:54:34.358 UTC, updated_at=2023-07-21 20:54:34.358 UTC>>,
 @error=nil,
 @metadata={:request_id=>nil}>
```

hearth/lib/hearth/http/fields.rb
```diff
@@ -103,6 +103,10 @@ def each(&block)
                  .each(&block)
         end
         alias each_pair each
+
+        def inspect
+          "#<#{self.class.name}>"
```
alextwoods, line 108: Why do we do this?

hearth/lib/hearth/middleware/send.rb
```diff
@@ -86,30 +91,50 @@ def send_request(context, output)
...
           apply_stub(stub, input, context, output) if stub
         when Exception, ApiError
           output.error = stub
         when Hash
-          @stub_class.stub(
-            context.response,
-            stub: @params_class.build(
-              stub,
-              context: 'stub'
+          if stub.key?(:error)
```
alextwoods, line 103: I'd still argue for either:
1) Check here for something like `stub.key?(:error) and stub[:error].is_a?(Class)`.  And then avoid the data key.  (since an output w/ a class value isn't valid - it avoids any chance of an api overlapping with this).
2) Creating a `ResponseError` class and adding a `when ResponseError` to this `case`.

I just think errors are a less common stubbing case and I want to avoid friction in upgrading from V3 to V4 and avoid interface changes where its reasonable....

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/BuilderGeneratorBase.java
```diff
@@ -280,17 +280,17 @@ protected void renderBuildersForOperation(OperationShape operation, Shape inputS
...
                 .closeBlock("end");
 
-        LOGGER.finer("Generated operation builder for: " + operation.getId().getName());
+        LOGGER.finer("Generated builder for operation " + operation.getId().getName());
     }
 
     /**
      * @param inputShape inputShape from a streaming operation to render for.
      */
+    // TODO: should probably be in Rest Base
```
alextwoods, line 293: Yeah, this is where this breaks down a little bit.  We do rely on this being in all of our existing, HTTP protocols.  We could add another level:  Rest -> Http -> Base -> Interface.

codegen/projections/high_score_service/lib/high_score_service/stubs.rb
```diff
@@ -43,7 +60,25 @@ def self.stub(http_resp, stub:)
       end
     end
 
-    # Operation Stubber for GetHighScore
+    class ErrorMessages
+      def self.default(visited=[])
```
alextwoods, line 64: Unrelated to this change: Should we be using a `Set` for visited?  Probably yes....

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/RestStubsGeneratorBase.java
```diff
@@ -130,32 +131,43 @@ protected abstract void renderPayloadBodyStub(OperationShape operation, Shape ou
...
-                .call(() -> renderOperationBodyStubber(operation, outputShape))
+                .call(() -> renderStatusCodeDataStubber(operation))
+                .call(() -> renderHeaderStubbers(outputShape))
+                .call(() -> renderPrefixHeadersStubbers(outputShape))
+                .call(() -> renderResponseCodeStubber(outputShape))
+                .call(() -> renderBodyStubber(outputShape))
+                .closeBlock("end");
+    }
+
+    // TODO: this is not entirely correct and will need to change later.
```
alextwoods, line 151: What part isn't correct?

hearth/lib/hearth/http/fields.rb
mullermp: Response output is very verbose because Headers/Trailers will show the value of Fields too. It's duplicate information.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/BuilderGeneratorBase.java
mullermp: Yeah. I think this is a bigger refactor and it's not just Builder generator but also Errors, Stubs, Parsers, etc. For now I just put a comment but we should revisit especially when doing mqtt.

codegen/projections/high_score_service/lib/high_score_service/stubs.rb
mullermp: With the logic here, I don't think it's possible for the same key to exist twice, right? Because we check if it includes. What's the argument for Set?

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/RestStubsGeneratorBase.java
mullermp: I've re-used the normal stub methods but I think we need error stub method equivalents because they will need additional abstract methods to render error code from body, header, or status.

codegen/projections/high_score_service/lib/high_score_service/stubs.rb
alextwoods: Just an argument for efficiency and using a data structure that matches the problem.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/RestStubsGeneratorBase.java
alextwoods: Ah gotcha - maybe just update the comment to include that.

codegen/projections/high_score_service/lib/high_score_service/stubs.rb
mullermp: Yeah, set can make sense. It would have to be required in this file.

hearth/lib/hearth/middleware/send.rb
mullermp: We've already made a lot of friction such as Config object, namespace, return Output instead of data, etc. I agree with the sentiment. I actually think since hash is ambiguous, using data/error key is the correct call. It also allows us to maybe stub metadata in the future? The other thing I thought of was maybe just an array is assumed to be an error but that is kinda wonky.

hearth/lib/hearth/middleware/send.rb
```diff
@@ -79,37 +69,99 @@ def call(input, context)
...
           raise ArgumentError, 'Unsupported stub type'
         end
       end
+
+      def apply_stub_hash(stub, context)
+        if stub.key?(:error)
+          apply_stub_hash_error(stub, context)
+        elsif stub.key?(:data)
+          apply_stub_hash_data(stub, context)
+        else
```
alextwoods, line 115: Should we validate in the above cases that it contains ONLY error or data? Right now you could pass {error: ..., data: ...}.  Or something like {error: MyError, message: 'I think this will work but it gets ignored'}

hearth/lib/hearth/middleware/send.rb
```diff
@@ -79,37 +69,99 @@ def call(input, context)
...
         when NilClass
-          @stub_class.stub(
-            context.response,
-            stub: @params_class.build(
-              @stub_class.default,
-              context: 'stub'
-            )
-          )
+          apply_stub_nil(context)
+        when Hearth::Structure
```
alextwoods, line 101: I think we may want to validate that the structure passed is actually the right one for the operation.  Though I realize we don't actually have that type here - but I think like we add PARAMS_CLASS to the stub, I think its reasonable to add the type class somewhere (either in the stub or in the params class).

hearth/lib/hearth/middleware/send.rb
alextwoods: Additionally - do we want validation when adding the stubs on the client? That puts the error close to when its created.  But obviously we'd still need the same validation here since you can stub with a proc and that could return an invalid stub.  So to avoid repeating.... maybe we only do validation at this point?

alextwoods (commented):
Looks good overall.

hearth/lib/hearth/middleware/send.rb
mullermp: Validated it's either one or the other. I think we don't want to validate when adding stubs because of procs - late as possible is fine I think.

hearth/lib/hearth/middleware/send.rb
mullermp: I added a TYPES_CLASS to data stubs.

alextwoods (approved):
Looks good for Part 1.

PR #148, opened 2023-07-24
Benchmark

alextwoods:
*Description of changes:*
Follow approach for benchmarking from v3 sdk.

This also adds generation of the `VERSION` file - following the practice we've done in other gems (including the new Hearth). 


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,55 @@
...
+  id-token: write # required for OIDC
+  contents: read
+
+jobs:
+  benchmark:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby: [3.2]
```
mullermp, line 22: We may as well run this on all 3.0+ based ruby versions

benchmark/gems/hearth.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module Benchmark
+  module Gems
+    class Core < Benchmark::Gem
```
mullermp, line 5: I feel like hearth benchmark is pretty useless. It's also not aws/core ..

benchmark/metrics.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+module Benchmark
+  module Metrics
+    # put metrics generated from `run_benchmarks` to cloudwatch
+    def self.put_metric(client, dims, timestamp, k, v)
```
mullermp, line 6: Positional args for this is kind of weird .. I know it's like this in v3

mullermp (approved):

benchmark/gems/hearth.rb
alextwoods: Updated the name to be Hearth instead. I do think its useful to track the gem size and require time over time.

benchmark/metrics.rb
alextwoods: Updated to kwargs

.github/workflows/benchmark.yml
alextwoods: Added

PR #149, opened 2023-07-25
Update CI with SimpleCov

jterapin:
*Description of changes:*
- Updated `RequestCompression` specs for full coverage
- Update CI with Simple Cov for Hearth library 

### DIFF
--- a/hearth/spec/hearth/http/middleware/request_compression_spec.rb
+++ b/hearth/spec/hearth/http/middleware/request_compression_spec.rb
@@ -116,15 +116,15 @@ module Hearth
                 end.to raise_error(StandardError)
               end
 
-              it 'raises a standard error in streaming compression' do
-                expect do
-                  subject.send(
-                    :process_streaming_compression,
-                    'custom',
-                    request
-                  )
-                end.to raise_error(StandardError)
-              end
+              # it 'raises a standard error in streaming compression' do
+              #   expect do
+              #     subject.send(
+              #      :process_streaming_compression,
+              #      'custom',
+              #      request
+              #     )
+              #   end.to raise_error(StandardError)
+              # end
             end
           end
 
--- a/hearth/spec/hearth/http/middleware/request_compression_spec.rb
+++ b/hearth/spec/hearth/http/middleware/request_compression_spec.rb
@@ -116,15 +116,15 @@ module Hearth
                 end.to raise_error(StandardError)
               end
 
-              # it 'raises a standard error in streaming compression' do
-              #   expect do
-              #     subject.send(
-              #      :process_streaming_compression,
-              #      'custom',
-              #      request
-              #     )
-              #   end.to raise_error(StandardError)
-              # end
+              it 'raises a standard error in streaming compression' do
+                expect do
+                  subject.send(
+                   :process_streaming_compression,
+                   'custom',
+                   request
+                  )
+                end.to raise_error(StandardError)
+              end
             end
           end
 
--- a/hearth/spec/hearth/http/middleware/request_compression_spec.rb
+++ b/hearth/spec/hearth/http/middleware/request_compression_spec.rb
@@ -81,24 +81,55 @@ module Hearth
             end
           end
 
-          context 'no supported encodings' do
-            let(:encodings) { ['custom'] }
+          context 'encodings' do
+            context 'none supported by client' do
+              let(:encodings) { ['custom'] }
 
-            it 'skips compression' do
-              expect(app).to receive(:call).with(input, context)
-              resp = subject.call(input, context)
-              expect_uncompressed_body(request, body)
-              expect(resp).to be output
+              it 'skips compression' do
+                expect(app).to receive(:call).with(input, context)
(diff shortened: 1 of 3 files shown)


hearth/spec/hearth/http/middleware/request_compression_spec.rb
```diff
@@ -81,24 +81,55 @@ def expect_uncompressed_body(request, body)
...
+                resp = subject.call(input, context)
+                expect(request.headers['Content-Encoding']).to eq('gzip')
+                expect(resp).to be output
+              end
+            end
+
+            context 'unsupported encoding is given' do
+              it 'raises a standard error in compression' do
+                expect do
+                  subject.send(:process_compression,
```
mullermp, line 111: nitpick on styling - multiline these indented under subject.

subject.send(
  foo
  bar
)

jterapin: Looks much better!, I will advise - thanks!

jterapin:
All set to review! I commented out some parts of my spec to test to see if the CI fails. Working as intended. 👍

alextwoods (approved):
Nice!! LGTM

hearth/spec/hearth/http/middleware/request_compression_spec.rb
```diff
@@ -81,24 +81,59 @@ def expect_uncompressed_body(request, body)
...
+                resp = subject.call(input, context)
+                expect(request.headers['Content-Encoding']).to eq('gzip')
+                expect(resp).to be output
+              end
+            end
+
+            context 'unsupported encoding is given' do
+              it 'raises a standard error in compression' do
+                expect do
+                  subject.send(
```
mullermp, line 111: Rather than testing private methods (using subject.send - usually not great to do because it's an implementation detail) can you instead test the behavior using subject.call?

hearth/spec/hearth/http/middleware/request_compression_spec.rb
```diff
@@ -81,24 +81,59 @@ def expect_uncompressed_body(request, body)
...
+                    :process_compression,
+                    'custom',
+                    request
+                  )
+                end.to raise_error(StandardError)
```
mullermp, line 116: I think these errors should have been RuntimeErrors - I see it's StandardError in the implementation though.

hearth/spec/hearth/http/middleware/request_compression_spec.rb
jterapin: If I were to test via using `subject.call`, I would have to change the [`SUPPORTED_ENCODING` constant](https://github.com/smithy-lang/smithy-ruby/blob/a30fa264f24147e058a28fd27e1b9829dc600fe4/hearth/lib/hearth/http/middleware/request_compression.rb#L10) (by using `.replace` since it is frozen). Since, changing the constants are highly discouraged - I opted for using `subject.send` instead. 

Let me know if there's another avenue that I can look at!

hearth/spec/hearth/http/middleware/request_compression_spec.rb
jterapin: I just did some digging on this to understand better. Makes sense to use `RuntimeError` as it is a reasonable default rather than using the parent class, `StandardError`. 

I'll change it now.

hearth/spec/hearth/http/middleware/request_compression_spec.rb
mullermp: Are you not able to use stub_const?

hearth/spec/hearth/http/middleware/request_compression_spec.rb
mullermp: Nit pick on the error message too.. Not sure if we should use language like "We do not" - try to be objective - something like "Encoding #{encoding} is not supported"

hearth/spec/hearth/http/middleware/request_compression_spec.rb
mullermp: `stub_const('Hearth::HTTP::Middleware::SUPPORTED_ENCODINGS', ['gzip', 'custom'])`

hearth/spec/hearth/http/middleware/request_compression_spec.rb
alextwoods: This error should never actually reach end users - the modeling validation ensures that only a set of supported encodings is applied and the logic above in selecting a supported algorithm ensures we would never get here... hence the difficulty in testing it directly.

We could just remove this case completely.

jterapin: Per our discussion, I will be removing this case and any specs related to it completely

hearth/spec/hearth/http/middleware/request_compression_spec.rb
jterapin: Good to know for future! thanks

mullermp (approved):

PR #150, opened 2023-07-27
Codegen Output Demo - Config Constraints

jterapin:
*Description of changes:*
Before I dive into the design/implementation work, I wanted to visualize what the end result _could_ look like in Hearth and the `config.rb` file in the projected service. 
In this possible implementation, I'm thinking of adding a `validate_range` method to the Hearth's `Validator` module

Please ignore the inconsistency of naming as this is just a rough draft.

### DIFF
--- a/codegen/projections/white_label/lib/white_label/config.rb
+++ b/codegen/projections/white_label/lib/white_label/config.rb
@@ -128,6 +128,13 @@ module WhiteLabel
       Hearth::Validator.validate_types!(validate_input, TrueClass, FalseClass, context: 'config[:validate_input]')
     end
 
+    def validate_range!
+      Hearth::Validator.validate_range(
+        request_min_compression_size_bytes,
+        { :min_value => 0, :max_value => 10485760 },
+        context: 'config[:request_min_compression_size_bytes]')
+    end
+
     def self.defaults
       @defaults ||= {
         disable_host_prefix: [false],
--- a/hearth/lib/hearth/validator.rb
+++ b/hearth/lib/hearth/validator.rb
@@ -41,5 +41,20 @@ module Hearth
       raise ArgumentError,
             "Unexpected members: [#{unknown.join(', ')}]"
     end
+
+    # draft of what validate_range *could* look like
+    # valid_range param consists of min and max keys
+    def self.validate_range(value, valid_range, context:)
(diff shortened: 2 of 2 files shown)


codegen/projections/white_label/lib/white_label/config.rb
```diff
@@ -128,6 +128,13 @@ def validate_types!
       Hearth::Validator.validate_types!(validate_input, TrueClass, FalseClass, context: 'config[:validate_input]')
     end
 
+    def validate_range!
```
alextwoods, line 131: I think I might argue for changing the name of the `validate_types!` method to be just `validate!` (requires changing Hearth::Configuration) and then adding the validations for each member together.  Eg:  
```
validate_type(config_1)
validate_range(config_1)
some_other_future_validation(config_1)

validate_type(config_2)
validate_tyupe(config_3)
```

hearth/lib/hearth/validator.rb
```diff
@@ -41,5 +41,20 @@ def self.validate_unknown!(struct, params, context:)
       raise ArgumentError,
             "Unexpected members: [#{unknown.join(', ')}]"
     end
+
+    # draft of what validate_range *could* look like
+    # valid_range param consists of min and max keys
+    def self.validate_range(value, valid_range, context:)
```
alextwoods, line 47: What about using kwargs for min and max?
```
self.validate_range(value, min:, max:, context:)
```
Makes it easy to use like: 
```
validate_range(value, min: 0, max: 10000)
```

hearth/lib/hearth/validator.rb
```diff
@@ -41,5 +41,20 @@ def self.validate_unknown!(struct, params, context:)
...
+      # need to validate the valid_range object before proceeding
+      # does it make sense us to use existing method - validate_type!
+      valid_range.each { |context, value| self.validate_type!(value, Integer, context: context) }
```
alextwoods, line 50: I think we can probably skip validating that the inputs for min/max are correct since these aren't intended as customer input.

hearth/lib/hearth/validator.rb
jterapin: Ahh, yes. That would make sure we have `:min` and `:max` passed in.

PR #151, opened 2023-07-27
Stub errors codegen for protocols

mullermp:
Provide methods to override for stubbing errors

### DIFF
--- a/codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/StubsGeneratorBase.java
+++ b/codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/StubsGeneratorBase.java
@@ -49,7 +49,6 @@ import software.amazon.smithy.model.shapes.StructureShape;
 import software.amazon.smithy.model.shapes.TimestampShape;
 import software.amazon.smithy.model.shapes.UnionShape;
 import software.amazon.smithy.model.traits.ErrorTrait;
-import software.amazon.smithy.model.traits.StreamingTrait;
 import software.amazon.smithy.ruby.codegen.CodegenUtils;
 import software.amazon.smithy.ruby.codegen.GenerationContext;
 import software.amazon.smithy.ruby.codegen.RubyCodeWriter;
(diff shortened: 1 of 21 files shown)


codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/StubsGeneratorBase.java
```diff
@@ -353,16 +352,6 @@ protected void renderMemberDefaults(Shape s) {
         writer.closeBlock("}");
     }
 
-    // TODO: should probably be in Rest Base
```
alextwoods, line 356: I'm not quite convinced this should be in Rest Base (though this is HTTP specific, I believe it applies to every protocol we know right now.  But I guess this is where our only 2 levels of base classes falls short.  But since this would need to be duplicated in alot of the v4 protocols, I'd potentially lean towards leaving it here but making the todo more descriptive)

codegen/smithy-ruby-rails-codegen/src/main/java/software/amazon/smithy/ruby/codegen/protocol/railsjson/generators/StubsGenerator.java
```diff
@@ -128,6 +130,20 @@ protected void renderPayloadBodyStub(Shape outputShape, MemberShape payloadMembe
         }
     }
 
+    @Override
+    protected void renderHeaderStubbers(Shape shape) {
+        if (shape.hasTrait(ErrorTrait.class)) {
+            RailsJsonTrait railsJsonTrait = context.service().getTrait(RailsJsonTrait.class).get();
+            String errorLocation = railsJsonTrait.getErrorLocation().orElse("status_code");
```
alextwoods, line 137: How do we handle the status_code case?

alextwoods (commented):
Approach looks good

codegen/smithy-ruby-rails-codegen/src/main/java/software/amazon/smithy/ruby/codegen/protocol/railsjson/generators/StubsGenerator.java
mullermp: It's handled already in RestBase by looking at the status code on the error and setting it on the stub.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/StubsGeneratorBase.java
mullermp: It's used in restxml, restjson, fakeprotocol, and railsjson, all of which inherit from rest base. I think it belongs in an http specific base but rest base I think is a better home for it now. I don't believe this applies to other protocols like json 1.0/1.1, ec2, and query.

alextwoods (approved):
Looks good!

PR #154, opened 2023-08-02
Identity and Auth

mullermp:
Done:
* Identity classes
* Auth schemes
* Auth option and auth resolver
* Refreshing identity resolver module
* Basic and refreshing identity resolver class
* Auth middleware
* Auth / white label integration specs
* Identity specs
* Refreshing identity module specs
* Basic identity resolver spec
* Auth schemes specs
* Auth option specs
* Auth middleware spec

Signers and signing auth will be done in a different PR?


codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/middleware/factories/AuthMiddlewareFactory.java
```diff
@@ -0,0 +1,133 @@
...
+                    String authParams = "Auth::Params.new(operation_name: :%s)".formatted(operationName);
+                    params.put("auth_params", authParams);
+                    return params;
+                });
+        clientConfigSet.forEach(authBuilder::addConfig);
+
+        return authBuilder.build();
+    }
+
+    private static String identityResolverConfigNameForAuthTrait(Trait trait) {
```
alextwoods, line 107: I think we could clean up a lot of these If/else with some sort of AuthScheme Codegen class(es).

hearth/lib/hearth/auth_schemes.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+module Hearth
+  module AuthSchemes
+    # Base class for all AuthScheme classes.
+    class Base
+      def initialize(scheme_id:)
```
alextwoods, line 7: Why not also take identity_type and signer as parameters here?

hearth/lib/hearth/auth_schemes.rb
```diff
@@ -0,0 +1,29 @@
...
+      # @return [String]
+      attr_reader :scheme_id
+
+      # @return [IdentityResolver, nil]
+      def identity_resolver(identity_resolvers = {})
+        identity_resolvers[@identity_type]
+      end
+
+      # @return [Signers::Base]
+      attr_reader :signer
```
alextwoods, line 20: Is there a reason we don't have an attr_reader for identity_type?

hearth/lib/hearth/middleware/sign.rb
```diff
@@ -0,0 +1,21 @@
...
+      # @param [Class] app The next middleware in the stack.
+      def initialize(app)
+        @app = app
+      end
+
+      # @param input
+      # @param context
+      # @return [Output]
+      def call(input, context)
+        @app.call(input, context)
```
alextwoods, line 17: Should this apply the signer? (or add a todo for later?)

hearth/lib/hearth/refreshing_identity.rb
```diff
@@ -0,0 +1,61 @@
+# frozen_string_literal: true
+
+module Hearth
+  # Resolves an identity from identity properties.
+  module RefreshingIdentity
```
alextwoods, line 5: The naming here is a little confusing I think.  This seems like it should be the `RefreshingIdentityResolver`

hearth/lib/hearth/refreshing_identity_resolver.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module Hearth
+  # Refreshing Identity resolver that uses a proc to resolve an Identity.
+  # Automatically refreshes the identity if it is near expiration.
+  class RefreshingIdentityResolver < IdentityResolver
```
alextwoods, line 6: This seems like it could be something like.... ProcIdentityResolver or GenericIdentityResolver.  (naming is hard)

Do we actually need this?  I assume most IdentityResolvers will include the RefreshingIdentity module, and provide an implementation of refresh themselves (as we do in V3).

codegen/smithy-ruby-codegen/build.gradle.kts
```diff
@@ -37,3 +37,7 @@ dependencies {
     implementation("software.amazon.smithy:smithy-waiters:${rootProject.extra["smithyVersion"]}")
     implementation("software.amazon.smithy:smithy-protocol-test-traits:${rootProject.extra["smithyVersion"]}")
 }
+
+java.sourceSets["main"].java {
+    srcDirs("model", "src/main/smithy")
```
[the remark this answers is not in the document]
mullermp: For the Smithy plugin in IntelliJ, it adds syntax highlighting etc. Does it work for you without this?

hearth/lib/hearth/auth_schemes.rb
mullermp: Signer is no longer a thing that needs to be configurable - the signing properties are used to modify signing behavior instead. Identity type will always be static. I guess both can be passed in here, but it was not called out.

hearth/lib/hearth/auth_schemes.rb
mullermp: It's never accessed publicly. It's only referenced internally when calling identity_resolver with a hash of possible identity resolvers.

hearth/lib/hearth/middleware/sign.rb
mullermp: Yeah. It's a TODO.

hearth/lib/hearth/refreshing_identity.rb
mullermp: Refreshing Identity Resolver is already a class that we have now.

hearth/lib/hearth/refreshing_identity_resolver.rb
mullermp: We have IdentityResolver and RefreshingIdentityResolver. IdentityResolver doesn't care for expiration (calls a proc to return a thing) and RefreshingIdentityResolver has a slightly different interface (implement refresh method) and checks expiration. I think it's valid to have both types.. The non-aws identities (login, bearer, api) are probably things that won't need to refresh and probably should not implement "refresh" method, it's weird.

mullermp: Removing RefreshingIdentityResolver and renaming RefreshingIdentity to RefreshingIdentityResolver

hearth/spec/hearth/refreshing_identity_resolver_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+      end
+
+      context 'near sync expiration' do
+        let(:near_sync_expiration) { Time.now + 200 }
+        let(:near_sync_expiration_identity) do
+          Identities::Base.new(expiration: near_sync_expiration)
+        end
+
+        it 'refreshes synchronously' do
+          expect(proc).to receive(:call)
```
alextwoods, line 48: Maybe add an explicit check for Thread.new not being called?

alextwoods (approved):
Nice - LGTM!

PR #155, opened 2023-08-03
Implement config constraints

jterapin:
_Description of some notable changes:_
* Added a `ConfigConstraint` interface to keep all the constraint-related classes in check 
* Added `RangeConstraint` and `TypeConstraint` classes to render different constraint validators
* Removed `type` from `ClientConfig` constructor as it is redundant since we have `documentationType`
* `Validator` module in hearth will now validate other constraints such as range


codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/config/ConfigConstraint.java
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.smithy.ruby.codegen.config;
+
+public interface ConfigConstraint {
```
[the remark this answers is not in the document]
mullermp: Can types also fit as a constraint here?

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/config/ClientConfig.java
```diff
@@ -186,6 +195,7 @@ public Builder name(String name) {
          */
         public Builder type(String type) {
             this.type = type;
```
alextwoods, line 197: I think we should remove `type` from the ClientConfig (but leave it in the Builder).  When we construct a `ClientConfig`, we check the builder for `documentationType` and use that if set, otherwise we set `documentationType` to the `builder.type`.  Eg:

```
public ClientConfig(Builder builder) {
  this.documentationType = builder.documentationType != null ? builder.documentationType : builder.type;
```

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/config/ClientConfig.java
```diff
@@ -114,7 +102,7 @@ public String renderDefaults(GenerationContext context) {
     }
 
     public List<ConfigConstraint> getConstraints() {
-        return List.copyOf(this.constraints);
+        return constraints;
```
alextwoods, line 105: I think in the constructor we still need: `this.constraints = List.copyOf(builder.constraints)`;

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/ConfigGenerator.java
```diff
@@ -100,7 +100,7 @@ private void renderConfigDocumentation(RubyCodeWriter writer) {
         writer.writeYardMethod("initialize(*options)", () -> {
             clientConfigList.forEach((clientConfig) -> {
                 String member = RubyFormatter.asSymbol(RubySymbolProvider.toMemberName(clientConfig.getName()));
-                String returnType = clientConfig.getType();
+                String returnType = clientConfig.getDocumentationType();
```
mullermp, line 103: It'd be nice if we can keep using type here but special case Boolean instead of using documentation type, because we may want to have a responds to documentation in the future.

mullermp (approved):

hearth/lib/hearth/validator.rb
```diff
@@ -13,7 +13,7 @@ module Validator
     # @param context [String] The context of the value being validated.
     # @raise [ArgumentError] Raises when the value is not one of given type(s).
     def self.validate_types!(value, *types, context:)
-      return if !value || types.any? { |type| value.is_a?(type) }
+      return if value.nil? || types.any? { |type| value.is_a?(type) }
```
mullermp, line 16: Is this because of Boolean? We should make sure this is tested.

jterapin: Yup, won't raise when the input value is boolean `false` with a `String` as type

I'll add a test to cover this! 👍

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/config/TypeConstraint.java
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Type Constraint for config value.
+ */
+public class TypeConstraint implements ConfigConstraint {
+
+    private final String type;
+
+    /**
+     * @param type ruby type for the config. Used for validation, must be a valid Ruby class.
```
alextwoods, line 26: This is actually not exactly true - since Boolean isn't a Ruby class..... It certainly throws a wrench in how we do all of these.

I think lets just add "must be a valid Ruby class or `Boolean`".

hearth/lib/hearth/validator.rb
```diff
@@ -41,5 +41,19 @@ def self.validate_unknown!(struct, params, context:)
       raise ArgumentError,
             "Unexpected members: [#{unknown.join(', ')}]"
     end
+
+    # Validate the given value is within the expected range (inclusive).
+    # @param value [Object] The value to validate.
+    # @param min [Integer] The minimum that the given value should be.
```
alextwoods, line 47: I think this should work for any `Numeric`, so I think we could update the doc type here.  (Our RangeConstraint in Java codegen currently only applies for integer types, but this code is more flexible).

alextwoods (approved):
Nice - Looks good!

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/config/TypeConstraint.java
jterapin: haha, that is very true. I'll update.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/ConfigGenerator.java
jterapin: Initially, I was looking for areas where we used `getType()` method to see how it was used throughout the codegen. Then, came across this piece. I thought updating to  using `getDocumentationType()` was appropriate since the method, `renderConfigDocumentation`  is to render documentation. 

> because we may want to have a responds to documentation in the future

Could you expand on this? I don't follow (might be due to lack of context)

jterapin: While I did remove `type` from ClientConfig constructor, the `type` method still exists on the Builder for convenience and to avoid breaking existing code (also creates type constraint and takes care of `documentationType` if unset). You could still skip using the `type` builder method to just set the `documentationType` like below:
```
ClientConfig.builder()
  .constraint(new RespondsToConstraint(":call"))
  .documentationType("Callable")
  .build()
```
I believe this will address the above.

PR #156, opened 2023-08-15
Factories for Middleware (and clean ups)

mullermp:
Add static factories for middleware building, to clean up application transport and middleware builder. Also makes request compression a factory instead of integration.


codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/middleware/MiddlewareFactory.java
```diff
@@ -0,0 +1,24 @@
...
+ * permissions and limitations under the License.
+ */
+
+package software.amazon.smithy.ruby.codegen.middleware;
+
+import software.amazon.smithy.ruby.codegen.GenerationContext;
+
+public interface MiddlewareFactory {
+    static Middleware build(GenerationContext context) {
+        return null;
```
alextwoods, line 22: Shouldn't have a default implementation.

mullermp: It needs one because java

alextwoods (approved):

PR #157, opened 2023-08-21
Add rbs and touch up public methods

mullermp:
Also flattens some stuff


hearth/lib/hearth/config/env_provider.rb
```diff
@@ -1,9 +1,9 @@
 # frozen_string_literal: true
 
 module Hearth
+  # @api private
```
alextwoods, line 4: Should these be private? We'd want external users who are extending smithy-ruby to use these implementations right?  But I guess thats different than SDK users, who likely wouldn't be using these directly....

mullermp: Yeah. I think they're supposed to be private in runtime but public as a Hearth constant in codegen? Most likely. We probably also want to make sure Hearth.java doesn't include constants for API private code

alextwoods (approved):

PR #158, opened 2023-08-21
Add sign middleware

mullermp:
Implements Sign middleware and fixes auth scheme generation for some auth cases. Implemented signers and tests, except digest auth.

Signers have a way to reset themselves now.

Minor changes to Field enumeration, context (auth), retry, related tests.


codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/middleware/factories/AuthMiddlewareFactory.java
```diff
@@ -130,4 +134,16 @@ private static String defaultIdentityResolverForAuthTrait(Trait trait) {
         return "proc { |cfg| cfg[:stub_responses] ? %s.new(proc { %s }) : nil }"
                 .formatted(Hearth.IDENTITY_RESOLVER, identity);
     }
+
+    private static String hearthIdentityForAuthTrait(Trait trait) {
```
alextwoods, line 138: Can fix this later as part of larger refactoring of codegen to address extensibility of auth schemes (possibly adding the `AuthScheme` interface + an `additionalAuthSchemes` method in interceptors) - but this will need to be changed as well and made non hearth specific.

hearth/lib/hearth/request.rb
```diff
@@ -19,5 +18,11 @@ def initialize(uri: URI(''), body: StringIO.new)
 
     # @return [IO]
     attr_accessor :body
+
+    # @api private
+    def initialize_copy(other)
+      @uri = other.uri.dup
+      # body should not be copied
```
alextwoods, line 25: But should it be rewound (if respond_to rewind)?

hearth/lib/hearth/signers/http_digest.rb
```diff
@@ -4,8 +4,10 @@ module Hearth
   module Signers
     # A signer that signs requests using the HTTP Digest Auth scheme.
     class HTTPDigest < Signers::Base
-      def sign(request:, identity:, properties: {})
-        # TODO
+      def sign(request:, identity:, properties:)
+        # TODO: requires a nonce from the server - this cannot
```
alextwoods, line 8: Should these raise a non implemented error in the mean time?

hearth/lib/hearth/context.rb
```diff
@@ -13,17 +13,16 @@ def initialize(options = {})
...
     # @return [Symbol] Name of the API operation called.
     attr_reader :operation_name
 
     # @return [Hearth::HTTP::Request]
-    attr_reader :request
+    attr_accessor :request
 
     # @return [Hearth::HTTP::Response]
-    attr_reader :response
+    attr_accessor :response
```
alextwoods, line 25: This can still be a reader right? since we are just calling `reset` no the existing one.

hearth/lib/hearth/request.rb
mullermp: That's done in retry. This is for dup / copy. But I removed this in favor of signer reset.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/middleware/factories/AuthMiddlewareFactory.java
mullermp: Yeah I think we would have getters on auth scheme that map to hearth classes.

hearth/lib/hearth/context.rb
mullermp: For stub response, we should make it an accessor. I think it makes sense to replace in that case, so that the test has access to response for assertions.

alextwoods: I'm not sure I understand, stub responses can modify the request, which allows tests to have access to responses for assertions.

And if the response is replaced, any wrappers put on the response body (eg, a verify checksum) will be lost.

mullermp: Fair point. I think both need to be readers now, and I'll bring back the replace method. Since we are modifying the request (reset signing) instead of dup.

hearth/lib/hearth/middleware/auth.rb
```diff
@@ -28,17 +26,19 @@ def initialize(app, auth_resolver:, auth_params:,
...
       end
 
       private
 
-      SelectedAuthScheme = Struct.new(:identity, :signer, :auth_option)
-
-      def anonymous_identity_resolver
-        Hearth::IdentityResolver.new(proc { Identities::Anonymous.new })
-      end
+      ResolvedAuth = Struct.new(
```
alextwoods, line 35: Nice, I like this flatter approach.

hearth/lib/hearth/signers.rb
```diff
@@ -1,11 +1,16 @@
...
     # Base class for all Signer classes.
     class Base
-      def sign(request:, identity:, properties: {})
-        # do nothing
+      def sign(request:, identity:, properties:)
+        # Do nothing by default
+      end
+
+      def reset(request:, properties:)
+        # Do nothing by default
```
alextwoods, line 13: Do we want the base signer to do nothing here or should it raise a not implemented error - to suggest that you should provide an implementation (even if its empty)?

alextwoods (approved):
Nice, I like the signer reset approach - LGTM!

PR #160, opened 2023-08-25
Improve rbs and docs

mullermp:
Done:
* waiters
* paginators
* client
* errors
* types
* auth
* client
* minimal set of hearth signatures
* generated config types


alextwoods (approved):
Nice - docs seem a lot more useful

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/RubyCodeWriter.java
```diff
@@ -377,7 +408,7 @@ public RubyCodeWriter apply(String filename, String namespace) {
     }
 
     /**
-     * Adds TypeScript symbols for the "$T" formatter.
+     * Adds Ruby symbols for the "$T" formatter.
```
alextwoods, line 411: Haha, whoops.  I wonder where this code came from....

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/ClientGenerator.java
```diff
@@ -196,27 +192,32 @@ private void renderRbsOperations(RubyCodeWriter writer) {
...
-        String operationName =
-                RubyFormatter.toSnakeCase(symbol.getName());
-
-        String documentation = new ShapeDocumentationGenerator(model, symbolProvider, operation).render();
+        String operationName = RubyFormatter.toSnakeCase(symbolProvider.toSymbol(operation).getName());
+        boolean isStreaming = outputShape.members().stream()
+                .anyMatch((m) -> m.getMemberTrait(model, StreamingTrait.class).isPresent());
+        // optimization for client generation
+        if (isStreaming) {
+            hasStreamingOperation = true;
```
alextwoods, line 202: Whats this doing? Doesn't quite seem right to set this here

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/UnionGenerator.java
```diff
@@ -86,27 +82,23 @@ public void render() {
 
             for (MemberShape memberShape : shape.members()) {
                 writer
-                    .openBlock("class $L < $T",
-                            symbolProvider.toMemberName(memberShape), symbol)
-                    .write("def to_h: () -> { $L: Hash[Symbol,$T] }",
-                            RubyFormatter.toSnakeCase(symbolProvider.toMemberName(memberShape)),
-                            symbol)
+                    .openBlock("class $L < $T", symbolProvider.toMemberName(memberShape), symbol)
+                    .write("def to_h: () -> { $L: Hash[Symbol, $T] }",
```
alextwoods, line 86: What does this type signature mean? I thought the to_h should potentally be Hash[Symbol, untyped]?

hearth/lib/hearth/http/fields.rb
```diff
@@ -47,7 +46,7 @@ def delete(key)
         @entries.delete(key.downcase)
       end
 
-      # @return [Enumerable<String,Field>]
+      # @return [Enumerable<Field>]
       def each(&block)
         @entries.values.each(&block)
```
alextwoods, line 51: Is this the right implementation for `each`? or would we want it to be over <String, Field> (as a pair?)

hearth/lib/hearth/interceptor_list.rb
```diff
@@ -35,6 +35,8 @@ def add(interceptor)
 
     alias << add
 
+    # TODO: remove this from public api
```
alextwoods, line 38: Why? I think it must be part of the public API if its part of Config? (unless we move away from using the List classes and uses some of List of Type validation?)

alextwoods (approved):
Nice!

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/RubyCodeWriter.java
mullermp: Haha I wonder...

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/ClientGenerator.java
mullermp: We iterate operations for streaming and if we see one, we know to generate the block io response body method. I'm also using it to have block in the method signature.

hearth/lib/hearth/http/fields.rb
mullermp: Fields will be an array of Field. That was a recent change. It's more usable that way.

mullermp: Headers and trailers iterate keys and values. Field has all the properties. It's slightly different.

hearth/lib/hearth/interceptor_list.rb
mullermp: Open to ideas but I don't think this should be publicly callable in a public API (List)

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/UnionGenerator.java
mullermp: This is what it was before. It is the union parent class. It ends up being something like: 

```
    class Union < Hearth::Union
      class String < Union
        def to_h: () -> { string: Hash[Symbol, Union] }
        def to_s: () -> String
      end

      class Struct < Union
        def to_h: () -> { struct: Hash[Symbol, Union] }
        def to_s: () -> String
      end

      class Unknown < Union
        def to_h: () -> { unknown: Hash[Symbol, Union] }
        def to_s: () -> String
      end
    end
```

mullermp: I think actually it needs to be:

```
def to_h: () -> { string: String }
def to_h: () -> { struct: Hash[Symbol, untyped] }
```

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/ClientGenerator.java
alextwoods: While that does work (and I probably wrote that code originally....) - I dont think we should rely on setting a member in each of these method calls.  Would be better to set `hasStreamingOperation` once by inspecting all operations at that time.  (can still use the `isStreaming` locally, I just dont think we should be setting the instance variable here).

hearth/lib/hearth/interceptor_list.rb
alextwoods: Ah yeah - I see what you mean.  The functionality could be moved into a utilility in... middleware maybe?

I'm also still a bit on the fence about having a class for this list in the ifrst place.  It seems like we should just have a List and add better validation utilities (which we now have the ability to do in codegen).

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/middleware/factories/HostPrefixMiddlewareFactory.java
```diff
@@ -37,6 +37,7 @@ public static Middleware build(GenerationContext context) {
         ClientConfig disableHostPrefix = ClientConfig.builder()
                 .name("disable_host_prefix")
                 .type("Boolean")
+                .rbsType("bool")
```
alextwoods, line 40: We have some special logic around the "Boolean" "type" already - I wonder if its worth adding a special case in the builder for this to just automatically set the rbsType?

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/middleware/MiddlewareBuilder.java
```diff
@@ -186,6 +186,7 @@ private Collection<? extends ClientConfig> getDefaultClientConfig() {
         ClientConfig logger = ClientConfig.builder()
                 .name("logger")
                 .type("Logger")
+                .rbsType("untyped") // TODO: find out why Logger does not resolve
```
alextwoods, line 189: Rbs seems to be missing some core classes..... my guess is its missing this one :-(

alextwoods (approved):

codegen/smithy-ruby-codegen-test/model/component-test/paginators.smithy
```diff
@@ -44,6 +44,11 @@ structure ResultWrapper {
     __123nextToken: String
 }
 
-list Items {
-    member: String
+//list Items {
+//    member: String
+//}
```
jterapin, line 49: These are commented out - what's the reason to keep them? Just curious about the intent!

jterapin (approved):
LGTM! 👍

codegen/smithy-ruby-codegen-test/model/component-test/paginators.smithy
mullermp: Good catch. It was from testing different types.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/middleware/MiddlewareBuilder.java
mullermp: Created issues in rbs

hearth/lib/hearth/interceptor_list.rb
mullermp: I think the InterceptorList makes sense. We could possible do this code in the Interceptor module (api private). I will investigate in a follow up PR - I have a list of small follow ups that I found during this.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/middleware/factories/HostPrefixMiddlewareFactory.java
mullermp: Maybe? Can think about other improvements too.. we definitely need an override for this anyway. Arguably we should have .type("TrueClass, FalseClass") (it could even be documented this way in Yard) and also "Boolean" as docType.

PR #161, opened 2023-08-28
Update relative middleware support

jterapin:
*Description of changes:*
Current implementation of the Relative middleware raises an error when a middleware specifies a Relative that does not exist. The behavior that we want to adjust is that if that Relative is not a **required** middleware, the middleware stack ordering should proceed as normal. This PR will address this modification.

In addition to the above, the following has been added/revised:
* Documentation for relative middleware class
* Added specs to cover relative middleware ordering
* Added a `CodegenException` to cover cases when a streaming body has a `httpChecksumRequired` trait
* Update existing compression specs to cover a streaming pipe body
* Changed `FINALIZE` step to `SIGN` to match with Hearth.


codegen/projections/white_label/lib/white_label/client.rb
```diff
@@ -137,9 +140,16 @@ def defaults_test(params = {}, options = {}, &block)
         validator: Validators::DefaultsTestInput,
         validate_input: config.validate_input
       )
+      stack.use(Middleware::BeforeMiddleware)
```
mullermp, line 143: I don't think we intended for middleware to be relative in runtime, but it was supposed to be relative in code generation?

codegen/projections/white_label/lib/white_label/client.rb
```diff
@@ -9,6 +9,9 @@
 
 require 'stringio'
 
+require_relative 'middleware/test_middleware'
```
alextwoods, line 12: We should add something to the way we write out the dependent ruby files (eg when doing `.rubySource("smithy-ruby-codegen-test-utils/middleware/test_middleware.rb")` to remove duplicates - IE, probably just using a Set is sufficent.

codegen/smithy-ruby-codegen-test-utils/middleware/test_middleware.rb
```diff
@@ -15,4 +15,48 @@ def call(input, context)
...
+      context.metadata[:before_middleware] = true
+      @app.call(input, context)
+    end
+  end
+
+  # Middleware used to test relative middleware ordering -
+  # checks the 'BEFORE' middleware and an optional case
+  # when a relative middleware is not required
+  class MidMiddleware
+    def initialize(app, verify_before_middleware:)
```
alextwoods, line 36: This is a little confusing to have the MidMiddlware take a verify_before config (I know it gets used to verify the Before middleware, but from the middleware POV, its only associated with the Mid Middleware.  So maybe something like `verify_in_mid`?

codegen/smithy-ruby-codegen-test/integration-specs/relative_middleware_spec.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+require_relative 'spec_helper'
+
+module WhiteLabel
+  describe Client do
+    context 'relative middleware ordering' do
```
alextwoods, line 7: I do wish we had a better way to link the various parts of these tests together better.... but haven't been able to think of one.

I think maybe we should at least add a comment here that says something like "Tests the order of middleware added by the WhiteLabelTestIntegration" or something like that.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/middleware/MiddlewareBuilder.java
```diff
@@ -147,17 +146,19 @@ private void resolve(Middleware middleware, Set<Middleware> resolved, Set<Middle
...
+                    resolve(relativeTo, resolved, visiting, order, klassToMiddlewareMap);
+                    // the order of relativeTo should now be set
+                    if (relative.getType().equals(Middleware.Relative.Type.BEFORE)) {
+                        order.put(middleware, order.get(relativeTo) - 1);
+                    } else {
+                        order.put(middleware, order.get(relativeTo) + 1);
+                    }
                 } else {
-                    order.put(middleware, order.get(relativeTo) + 1);
+                    resolveNullRelativeTo(relative, middleware, order);
```
alextwoods, line 161: Rather than "Null" in the name, maybe "Missing"

alextwoods (commented):
Nice! looking good, just some minor comments.

codegen/smithy-ruby-codegen-test-utils/middleware/test_middleware.rb
mullermp: I think you can remove all of these configs and simply expect certain middleware to exist in app. It's an "onion" of next middleware.

codegen/projections/white_label/lib/white_label/client.rb
mullermp, line 143: Rather than adding these to every operation, can we have a single RelativeMiddleware operation and only add these to it?

codegen/projections/white_label/spec/relative_middleware_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    context 'relative middleware ordering' do
+      it 'allows a middleware to be set before their relative' do
+        config = Config.new(
+          stub_responses: true,
+          verify_before_middleware: proc { |metadata|
+            expect(metadata[:before_middleware]).to be(true)
+          }
+        )
+        client = Client.new(config)
+        client.request_compression_operation({ body: 'foo' })
```
mullermp, line 16: I think request compression operation should only be for testing that - I would use a new test middleware operation here.

codegen/smithy-ruby-codegen-test/model/component-test/compression.smithy
```diff
@@ -1,12 +1,13 @@
 $version: "1.0"
 namespace smithy.ruby.tests
 
+@httpChecksumRequired
```
mullermp, line 4: Formatting of this file got funky

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/middleware/factories/RequestCompressionMiddlewareFactory.java
```diff
@@ -59,31 +61,31 @@ public static Middleware build(GenerationContext context) {
...
                 .build();
 
+        Middleware.Relative compressionRelative = Middleware.Relative.builder()
+                .before("Hearth::HTTP::Middleware::ContentMD5")
+                .optional()
+                .build();
+
         Middleware.Builder compressionBuilder = Middleware.builder()
+                .klass("Hearth::HTTP::Middleware::RequestCompression")
+                .step(MiddlewareStackStep.AFTER_BUILD)
```
mullermp, line 71: Totally optional, but if you're up to it, I think we should also improve middleware stack steps to match 1:1 with our hearth middleware steps. We chose names similar to Go but it's kind of just confusing the way it is now.

codegen/smithy-ruby-codegen-test-utils/middleware/test_middleware.rb
```diff
@@ -15,4 +15,48 @@ def call(input, context)
...
+
+  # Middleware used to test relative middleware ordering -
+  # this is for 'BEFORE' type testing
+  class BeforeMiddleware
+    def initialize(app)
+      @app = app
+    end
+
+    def call(input, context)
+      context.metadata[:before_middleware] = true
```
mullermp, line 27: I believe you can write Rspec expectations here that check if `@app` contains mid and after middleware. And in mid middleware you can check for after and not before, etc. Then you don't need any of the Config values or to check context metadata. You simply just execute the operation and it would fail if it were wrong.

mullermp: Another (possibly better) option is to make these middleware no-op (just @app.call). And in the test, do something like `expect_any_instance_of(MidMiddleware).to receive(:call).ordered` (use ordered to make sure the middleware is called in an order you expect).

codegen/smithy-ruby-codegen-test-utils/middleware/test_middleware.rb
jterapin: Thanks to y'all - getting rid of the messy configs on these middlewares! 😃

codegen/smithy-ruby-codegen-test/integration-specs/relative_middleware_spec.rb
jterapin: Agreed, added comments.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/middleware/MiddlewareBuilder.java
jterapin: Done. Thank you!

codegen/projections/white_label/lib/white_label/client.rb
jterapin: Good call. We have a handy `.appliesOnlyToOperations` to do just exact thing.

codegen/projections/white_label/spec/relative_middleware_spec.rb
jterapin: Added!

codegen/smithy-ruby-codegen-test/model/component-test/compression.smithy
jterapin: Oof, fix'd.

codegen/smithy-ruby-codegen-test-utils/middleware/test_middleware.rb
jterapin: Updated the test setups so we can expect on output's metadata. (Yay! and thank you for the ideas)

codegen/projections/white_label/sig/white_label/types.rbs
```diff
@@ -126,6 +126,14 @@ module WhiteLabel
       include Hearth::Structure
     end
 
+    class RelativeOperationInput < ::Struct[untyped]
+      include Hearth::Structure
```
mullermp, line 130: I think you may have to merge main and rebuild.

mullermp (approved):
Nice!

alextwoods (approved):
Nice - LGTM!

codegen/smithy-ruby-codegen-test/model/component-test/relative.smithy
```diff
@@ -0,0 +1,6 @@
+$version: "1.0"
+
+namespace smithy.ruby.tests
+
+@http(method: "POST", uri: "/relative_operation")
+operation RelativeOperation {}
```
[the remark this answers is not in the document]
jterapin: Will update~

codegen/projections/white_label/sig/white_label/types.rbs
jterapin: !!! yes, thanks for reminder :)

PR #162, opened 2023-08-29
Misc improvements

mullermp:
RBS struct members
Refactor interceptor apply out of interceptor list
Rename yardType to docType
Extra RBS in hearth + some minor tweaks

other minor stuff.


codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/RubySymbolProvider.java
```diff
@@ -279,9 +279,9 @@ public Symbol timestampShape(TimestampShape shape) {
     @Override
     public Symbol listShape(ListShape shape) {
         Symbol member = toSymbol(model.expectShape(shape.getMember().getTarget()));
-        String rbsType = "Array[" + member.getProperty("rbsType").get() + "]";
-        String yardType = "Array<" + member.getProperty("yardType").get() + ">";
-        return createSymbolBuilder(shape, getDefaultShapeName(shape, "List__"), rbsType, yardType, moduleName)
+        String rbsType = "::Array[" + member.getProperty("rbsType").get() + "]";
```
alextwoods, line 282: Out of scope of this PR - but I think related to my comment from previous PR - I think maybe we shoudl define a RubyType class and create instances for things like Array<Type>. that will know how to do the doc types, rbs, contarints, ect....  I think this applies here and in the middleware and possibly in other places as well.

hearth/lib/hearth/middleware/build.rb
```diff
@@ -18,7 +18,8 @@ def initialize(app, builder:)
       # @param context
       # @return [Output]
       def call(input, context)
-        interceptor_error = context.interceptors.apply(
+        interceptor_error = Interceptor.apply(
+          interceptors: context.interceptors,
```
alextwoods, line 22: Will the interceptors we pass ever differ from whats on context?  I think they never would.  Should we drop this parameter from the method and just use the interceptors that are on the context internally?

alextwoods (approved):
LGTM - just minor comments

hearth/lib/hearth/middleware/build.rb
mullermp: Ah yeah I think we could.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/RubySymbolProvider.java
mullermp: I'm not sure what that would look like... But yeah we can potentially do that outside this change.

PR #163, opened 2023-08-30
Interceptor improvements

mullermp:
Adds interceptor class with usage documentation (pending location). Interceptor class allows for generic one-off interceptor creation using callables (useful for tests or quick debug insertions). It can still be a class with defined method.

Audit of interceptor usage - found two hooks not implemented.

Fixed tests to be stronger (assert when it should return Output or when it assigns output.error).

Change how interceptor apply works - do not mutate output object. It is now under Interceptors as an invoke method.

Interceptor::Context and Hooks got re-homed.

Updates RBS for all of the above.


hearth/lib/hearth/interceptor_list.rb
```diff
@@ -1,38 +1,35 @@
...
+              "#{Interceptor.hooks.join(', ')}"
       end
+
+      @interceptors << interceptor
+    end
+    alias << append
+
+    # @param [InterceptorList] other
+    # @return [InterceptorList] self
+    def concat(other)
```
alextwoods, line 29: Why this change? To make it behave more like Ruby lists where you can't use `<<` to concat another list?

hearth/lib/hearth/interceptor.rb
```diff
@@ -1,66 +1,464 @@
...
+
+        @callbacks[hook] = callbacks[hook]
+        define_singleton_method(hook) do |context|
+          @callbacks[hook].call(context)
+        end
+      end
     end
 
+    # @return [Hash<Symbol, Proc>]
+    attr_reader :callbacks
```
alextwoods, line 84: Should we have this exposed?

hearth/lib/hearth/interceptor.rb
```diff
@@ -1,66 +1,464 @@
...
 
-      last_error
+        unless valid_callback?(callbacks[hook])
+          raise ArgumentError,
+                "#{hook} must be a callable with arity 1 (context)"
+        end
+
+        @callbacks[hook] = callbacks[hook]
+        define_singleton_method(hook) do |context|
+          @callbacks[hook].call(context)
```
alextwoods, line 78: I think we can simplify this even more.  We can just pass the callback to the method and then we don't even need to store them at all in `@callbacks`:

```
define_singleton_method(hook, callbacks[hook])
```

alextwoods (commented):
Nice - I like this approach.  As far as I can tell, all of the interceptor logic in middleware is correct

hearth/lib/hearth/interceptor_list.rb
mullermp: Yeah this follows the enumerable interface.

hearth/lib/hearth/interceptor.rb
mullermp: I wasn't sure. Maybe not?

hearth/lib/hearth/interceptor.rb
mullermp: I'll try this.

mullermp: It works - however - it does not work for capturing input in our poller. I'm not entirely sure on the reason, but it looks to register differently than when I pass the proc as a block - maybe because it's double wrapped?

codegen/projections/weather/lib/weather/waiters.rb
```diff
@@ -78,7 +78,7 @@ def initialize(client, options = {})
       attr_reader :tags
 
       # @param (see Client#get_city)
-      # @return [true, Hearth::Waiters::WaiterFailed]
+      # @return [true]
```
alextwoods, line 81: Maybe I've forgotten how our waiters work.  Do they raise an exception instead of returning false? IF so, maybe that should be documented on the wait method?

hearth/spec/hearth/interceptor_spec.rb
```diff
@@ -28,34 +28,22 @@ module Hearth
...
+        it "defines a singleton method for #{hook}" do
+          expect(subject).to respond_to(hook)
+        end
 
-      it 'ignores unknown hooks' do
-        interceptor = Interceptor.new(unknown: callable)
-        expect(interceptor.callbacks).not_to have_key(:unknown)
+        it "calls the callback for #{hook}" do
+          output = subject.send(hook, context)
+          expect(output).to be context
```
alextwoods, line 45: This seems like it could happen in a few ways - maybe should assert the callback actually receives call

hearth/spec/hearth/interceptor_spec.rb
```diff
@@ -28,34 +28,22 @@ module Hearth
...
-        expect(subject.callbacks).to be_a(Hash)
-        expect(subject.callbacks.keys).to include(*Interceptor.hooks)
-      end
+    describe 'singleton methods' do
+      Interceptor.hooks.each do |hook|
+        it "defines a singleton method for #{hook}" do
+          expect(subject).to respond_to(hook)
+        end
 
-      it 'ignores unknown hooks' do
```
alextwoods: Why remove this test?

alextwoods (approved):
LGTM - minor comments

hearth/spec/hearth/interceptor_spec.rb
mullermp: I had it that way before, but I changed it when trying to change the define Singleton method approach. I can change this back.

codegen/projections/weather/lib/weather/waiters.rb
mullermp: They raise an exception or return true. So this part was not correct (encountered this trying to test the Singleton method change). I think I can add a raises yard tag here.

PR #164, opened 2023-09-06
Add Auth Scheme classes and integrations

mullermp:
Adds an AuthScheme class that can be implemented and then added to integrations.


alextwoods (commented):
General approach looks good - aside from moving AuthScheme to a concrete class and using a builder to define implementations instead of interface + implementations.

codegen/projections/high_score_service/lib/high_score_service/client.rb
```diff
@@ -63,10 +63,11 @@ def api_key_auth(params = {}, options = {})
         error_inspector_class: Hearth::HTTP::ErrorInspector
       )
       stack.use(Hearth::Middleware::Auth,
-        auth_schemes: options.fetch(:auth_schemes, config.auth_schemes),
         auth_params: Auth::Params.new(operation_name: :api_key_auth),
         http_api_key_identity_resolver: options.fetch(:http_api_key_identity_resolver, config.http_api_key_identity_resolver),
         auth_resolver: options.fetch(:auth_resolver, config.auth_resolver),
+        identity_resolver_map: { http_login_identity_resolver: Hearth::Identities::HTTPLogin, http_bearer_identity_resolver: Hearth::Identities::HTTPBearer, http_api_key_identity_resolver: Hearth::Identities::HTTPApiKey },
+        auth_schemes: options.fetch(:auth_schemes, config.auth_schemes),
```
alextwoods, line 70: should we be concating the auth_schemes rather than entirely replacing them?  (I think replacing them is probably fine, but worth double checking)

codegen/smithy-ruby-codegen-test-utils/src/main/resources/META-INF/services/software.amazon.smithy.ruby.codegen.RubyIntegration
```diff
@@ -1 +1,2 @@
+software.amazon.smithy.ruby.codegen.integrations.TestIntegration
```
alextwoods, line 1: I feel like TestIntegration may no longer be the right name for this.  It only provides the "FakeProtocol" - maybe we should call it the FakeProtocolIntegration?

We may also want to rename FakeProtocol to something more descriptive like, NoOpProtocol or NullProtocol or something like that?

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/auth/AuthScheme.java
```diff
@@ -0,0 +1,222 @@
...
+     */
+    public List<String> writeAdditionalFiles(GenerationContext context) {
+        return writeAdditionalFiles.writeAdditionalFiles(context);
+    }
+
+    @FunctionalInterface
+    /**
+     * Called to write out additional files needed by this auth scheme.
+     */
+    public interface WriteAdditionalFiles {
```
alextwoods, line 103: We use this interface in a number of places, maybe worth centralizing?

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/auth/AuthScheme.java
```diff
@@ -0,0 +1,222 @@
...
+    /**
+     * Write additional files required by this auth scheme.
+     *
+     * @param context generation context
+     * @return List of additional files written out
+     */
+    public List<String> writeAdditionalFiles(GenerationContext context) {
+        return writeAdditionalFiles.writeAdditionalFiles(context);
+    }
+
```
alextwoods, line 98: I believe we need to add an equals and hash implementation here for use in Sets/HashSet to work correctly

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/AuthGenerator.java
```diff
@@ -91,6 +115,13 @@ public void renderRbs() {
         LOGGER.fine("Wrote auth rbs module to " + rbsFile());
     }
 
+    private Object removeRbExtension(String s) {
```
alextwoods, line 118: We do this elsewhere as well - I think we may want to create some sort of additional files utility

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/AuthGenerator.java
```diff
@@ -91,6 +115,13 @@ public void renderRbs() {
...
+        if (s != null && s.endsWith(".rb")) {
+            return s.split(".rb")[0];
+        }
+        return s;
+    }
+
     private void renderAuthParamsClass(RubyCodeWriter writer) {
         // TODO: this should have more params when hooked up with endpoint/auth rules?
```
alextwoods, line 126: Should we be adding the additionalAuthParams from the authSchemes here?

codegen/projections/high_score_service/lib/high_score_service/client.rb
```diff
@@ -63,10 +63,11 @@ def api_key_auth(params = {}, options = {})
         error_inspector_class: Hearth::HTTP::ErrorInspector
       )
       stack.use(Hearth::Middleware::Auth,
-        auth_schemes: options.fetch(:auth_schemes, config.auth_schemes),
         auth_params: Auth::Params.new(operation_name: :api_key_auth),
         http_api_key_identity_resolver: options.fetch(:http_api_key_identity_resolver, config.http_api_key_identity_resolver),
         auth_resolver: options.fetch(:auth_resolver, config.auth_resolver),
+        identity_resolver_map: { http_login_identity_resolver: Hearth::Identities::HTTPLogin, http_bearer_identity_resolver: Hearth::Identities::HTTPBearer, http_api_key_identity_resolver: Hearth::Identities::HTTPApiKey },
```
alextwoods, line 69: Needing this map on every use of middleware seems wrong.  Why doesn't the auth scheme itself know the type and the symbol? 

Or why doesn't the Identity itself know the symbol?

Thinking about passing in a custom authScheme to an operation, how would this work?  It seems like the auth scheme would need to know how to get its identity if its not in that mapping?

codegen/smithy-ruby-codegen-test-utils/src/main/resources/META-INF/services/software.amazon.smithy.ruby.codegen.RubyIntegration
mullermp: Can do this in a different PR

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/auth/AuthScheme.java
mullermp: Where and how?

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/auth/AuthScheme.java
mullermp: What do you mean?

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/AuthGenerator.java
mullermp: I was thinking of adding a method to the writer, given a list of files, to write require relative.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/AuthGenerator.java
mullermp: Yeah. It's on my plan. I also need signer properties somehow.

codegen/projections/high_score_service/lib/high_score_service/client.rb
mullermp: I think it's a replacement but I can check.

codegen/projections/high_score_service/lib/high_score_service/client.rb
mullermp: I think the map is always needed because it's used to find the identity resolver based on the identity, for what is only valid on the operation. We previously constructed this in middleware by collecting the identity resolver keywords. A custom auth scheme wouldn't work unless it is one of the available ones for the operation (it's a way to force a single type of auth or auth resolution chain). The auth scheme does know the type but there is no mapping from type to config, which is arbitrary.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/middleware/factories/AuthMiddlewareFactory.java
```diff
@@ -44,22 +40,27 @@ public static Middleware build(GenerationContext context) {
         Map<ShapeId, Trait> serviceAuthSchemes =
                 ServiceIndex.of(context.model()).getAuthSchemes(context.service());
 
+        // get additional auth schemes
```
alextwoods, line 43: Should this be done once and set on the context?

codegen/projections/high_score_service/lib/high_score_service/client.rb
alextwoods: But the AuthScheme itself is what defines the configuration.  I think the example code has `identityResolver = authScheme.identityResolver(identityResolvers);`. It does seem like its reasonable for an authScheme to be able to say, given config, return the correct identity provider.  

But if we want to stick to auth schemes in ruby only defining their identity type and keeping this type to config map - I think it can be a constant on the generated middleware.  The type map shouldn't be specific to the operation (because config is typed at the client level.  The type of a certain config property cant change in different operations, nor can we add new ones in an operation).

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/auth/AuthScheme.java
alextwoods: Java's default for these won't work the way we want with Sets, so we need to define them and include the properties we consider important (maybe just shapeId?)

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/auth/AuthScheme.java
alextwoods: You can move this definition to its own file and then replace it here, RubyRuntimePlugin and Middleware.

codegen/projections/high_score_service/lib/high_score_service/client.rb
mullermp: Yeah, I think it's a replacement. It's supposed to be a map but I chose a list for simplicity. I think this forces an operation to only use certain auth schemes.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/auth/AuthScheme.java
mullermp: Added equals and hashCode for shapeId

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/middleware/factories/AuthMiddlewareFactory.java
mullermp: Ideally. But I also need a way to link it up to the model because I need to extract properties from the traits...

codegen/.rubocop.yml
```diff
@@ -18,6 +18,9 @@ Style/Documentation:
   Exclude:
     - '**/*_spec.rb'
 
+Style/FrozenStringLiteralComment:
+  Enabled: false
```
alextwoods, line 22: Why are we changing this?

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/GenerationContext.java
```diff
@@ -161,4 +164,25 @@ public List<RubyRuntimePlugin> getRuntimePlugins() {
                 .flatMap(List::stream)
                 .collect(Collectors.toList());
     }
+
+    /**
+     * @return list of all AuthSchemes from all integrations and the application transport.
+     */
+    public List<AuthScheme> getAuthSchemes() {
+        List<AuthScheme> authSchemes = applicationTransport().defaultAuthSchemes();
```
alextwoods, line 172: `defaultAuthSchemes` shouldn't return a modifiable list.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/GenerationContext.java
```diff
@@ -161,4 +164,25 @@ public List<RubyRuntimePlugin> getRuntimePlugins() {
...
+
+        ServiceIndex.of(model()).getAuthSchemes(service()).forEach((shapeId, trait) -> {
+            AuthScheme authScheme = authSchemes.stream()
+                    .filter(s -> s.getShapeId().equals(shapeId))
+                    .findFirst()
+                    .orElseThrow(() -> new IllegalStateException("No auth scheme found for " + shapeId));
+            authScheme.extractSignerProperties(trait);
+            authScheme.extractIdentityProperties(trait);
+        });
+        return authSchemes;
```
alextwoods, line 186: Ditto - this shouldn't be modifiable.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/GenerationContext.java
```diff
@@ -161,4 +164,25 @@ public List<RubyRuntimePlugin> getRuntimePlugins() {
...
+        integrations().forEach((i) -> {
+            i.getAdditionalAuthSchemes(this).forEach((s) -> authSchemes.add(s));
+        });
+
+        ServiceIndex.of(model()).getAuthSchemes(service()).forEach((shapeId, trait) -> {
+            AuthScheme authScheme = authSchemes.stream()
+                    .filter(s -> s.getShapeId().equals(shapeId))
+                    .findFirst()
+                    .orElseThrow(() -> new IllegalStateException("No auth scheme found for " + shapeId));
+            authScheme.extractSignerProperties(trait);
```
alextwoods, line 183: Why is this done here?  Is this modifying the auth scheme?  It seems like the authScheme itself shouldn't be modifiable once returned.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/RubyCodeWriter.java
```diff
@@ -139,6 +140,29 @@ public RubyCodeWriter includeRequires() {
         return this;
     }
 
+    /**
+     * Writes a require relative statement for an additional file.
```
alextwoods, line 144: Comment doesn't match the properties - "for additional files".

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/RubyRuntimePlugin.java
```diff
@@ -77,63 +72,24 @@ public interface RenderAdd {
...
-
-                    context.fileManifest().writeFile(fileName, writer.toString());
-                    return Collections.singletonList(relativeName);
-                } catch (IOException e) {
-                    throw new CodegenException(
-                            "Error reading rubySource file: " + rubyFileName,
-                            e);
-                }
-            };
+            this.writeAdditionalFiles = RubySource.rubySource(rubyFileName, "plugins/");
```
alextwoods, line 92: nice!

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/RubyRuntimePlugin.java
```diff
@@ -150,29 +106,6 @@ public Builder pluginClass(String pluginClass) {
...
         }
 
-        /**
-         * Used to completely override and fully customize the rendering of
-         * adding this plugin to the list of client class plugins.
-         *
-         * @param r Called to render code to add this plugin to the list of client class plugins
-         * @return Returns the Builder.
-         */
-        public Builder renderAdd(RenderAdd r) {
```
alextwoods, line 160: Why are these being removed?

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/auth/AuthScheme.java
```diff
@@ -0,0 +1,278 @@
...
+    public void extractSignerProperties(Trait trait) {
+         Map<String, String> signerProperties = extractSignerProperties.extractProperties(trait);
+         signerProperties.forEach((key, value) -> this.signerProperties.put(key, value));
+    }
+
+    /**
+     * @return the Signer properties for the auth scheme.
+     */
+    public Map<String, String> getSignerProperties() {
+        return signerProperties;
```
alextwoods, line 118: It seems like this method should take a Trait and pass it to the extractSignerProperties to avoid mutating authScheme.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/auth/AuthScheme.java
alextwoods: I still don't see equals and hashCode here?

mullermp: I removed it because I stopped using Set.

codegen/.rubocop.yml
mullermp: Because rubySource is an inner write. The outside will have it.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/RubyRuntimePlugin.java
mullermp: They weren't used

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/GenerationContext.java
mullermp: getRuntimePlugins is modifiable. Why can't this?

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/GenerationContext.java
mullermp: Yeah, it was modifying auth scheme. I added trait to the getter as you suggested, and I need to merge the static properties from the builder each time the getter is called.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/GenerationContext.java
alextwoods: `getRuntimePlugins` shouldn't be modifiable either.  It looks like its using `.collect(Collectors.toList());` which I had assumed created an unmodifiable list, but it does not.  

Can you change this to `.collect(Collectors.toUnmodifiableList());`

PR #165, opened 2023-09-12
Clean ups

mullermp:
Clean ups from the previous identity PRs.

* Rename test integration
* Ruby dependencies takes context
* Integrations return List only
* Remove identity_providers_map by using classes as keys in middleware

### DIFF
--- a/codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/middleware/factories/AuthMiddlewareFactory.java
+++ b/codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/middleware/factories/AuthMiddlewareFactory.java
@@ -19,6 +19,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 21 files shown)


codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/middleware/factories/AuthMiddlewareFactory.java
```diff
@@ -112,18 +89,65 @@ public static Middleware build(GenerationContext context) {
...
-                                    .map(s -> " " + s + " ")
-                                    .orElse(""));
-                    params.put("identity_resolver_map", identityResolverMapHash);
+                    String staticParamsBlock = params
+                            .entrySet()
+                            .stream()
+                            .map(entry -> entry.getKey() + ": " + entry.getValue())
+                            .collect(Collectors.joining(",\n"));
+
+                    // Dynamic params - identity resolvers as Class => config entries
```
alextwoods, line 98: Nice - I think I like this better than the old map.

alextwoods (approved):

PR #166, opened 2023-09-13
Remove Config from client initialize

mullermp:
Removes Config from Client initialize, requiring options hash instead. Config is still a public type and is returned on plugins or when inspecting the client.


codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/config/ClientConfig.java
alextwoods, line 1: With this change we are now merging ALL operation options into Config - that means we should remove the `allowOperationOverride` and we should remove the `options.fetch(:config, config.config)` in the client.rb

hearth/spec/hearth/configuration_spec.rb
```diff
@@ -31,49 +24,34 @@ def self.defaults
...
-          complex_option: complex_option
-        )
-        expect(simple_option).to receive(:dup)
-        expect(complex_option).to receive(:dup)
-        config.dup
+    describe '#merge' do
+      it 'merges the configuration' do
+        config = config_class.new(simple_option: 'test')
+        merged = config.merge(config_class.new(simple_option: 'test2'))
+        expect(merged.simple_option).to eq('test2')
```
alextwoods, line 54: Should we test invalid options in the merge?

hearth/spec/hearth/configuration_spec.rb
```diff
@@ -31,49 +24,34 @@ def self.defaults
...
-      it 'raises on invalid types' do
-        expect do
-          config_class.new(simple_option: 1)
-        end.to raise_error(
-          ArgumentError,
-          'Expected config[:simple_option] to be in [String], got Integer.'
-        )
-      end
-
-      it 'raises on unknown options' do
```
alextwoods, line 55: Why remove this test? Do we not validate this anymore?

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/config/ClientConfig.java
```diff
@@ -273,9 +273,7 @@ public Builder defaultDynamicValue(String rubyDefaultBlock) {
          */
         public Builder defaultValue(String value) {
             validateDefaultNotSet();
-            this.defaults = ConfigProviderChain.builder()
-                    .dynamicProvider("proc { " + value + " }")
-                    .build();
+            this.defaults = ConfigProviderChain.builder().dynamicProvider(value).build();
```
alextwoods, line 276: This isn't a dynamic config value.  Use the static provider here.

alextwoods (commented):
Looks good overall -  few minor comments.

hearth/spec/hearth/configuration_spec.rb
mullermp: Merge doesn't validate.

hearth/spec/hearth/configuration_spec.rb
mullermp: It's validated implicitly because of struct call, so I figure we don't need to test code we don't control.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/config/ClientConfig.java
mullermp: Ah yeah good point.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/config/ClientConfig.java
mullermp: There doesn't really seem to be a difference between the two. Maybe dynamic provider should provide the proc and |cfg| wrapping.

alextwoods: The most important difference is semantic - its worth conveying the intent.  But there is a behavior difference as well - the dynamic provider allows you to use a `ClientFragment`, which lets you explicitly specify a config dependency.  

I'm not sure about providing the proc and cfg wrapping - I think having the " { |cfg| .... }" is more explicit and readable.

hearth/spec/hearth/configuration_spec.rb
alextwoods: Was just thinking about if decide to change away form using a struct to implement and whether we consider raising on unknowns an important part of the interface or not.  I'd argue it is since we want to give users information about having potentially used the wrong options.

hearth/spec/hearth/configuration_spec.rb
alextwoods: I meant unknown rather than invalid - and I think that may be worth testing as well - I'd consider it a useful part of the interface for users - if they pass something wrong to the options on an operation, I think it should fail rather than silently ignore it.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/config/ClientConfig.java
```diff
@@ -258,7 +234,7 @@ public Builder defaultPrimitiveValue(String value) {
         }
 
         /**
-         * @param rubyDefaultBlock a single, dynamic default value to use.
+         * @param rubyDefaultBlock a single, dynamic default value to use. Assumes a proc with a cfg parameter.
```
alextwoods, line 237: This doc is a little confusing.  Instead of "Assumes" be more specific about exactly what you must provide (a callable that accepts a single value).

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/config/ConfigProviderChain.java
```diff
@@ -109,7 +109,7 @@ public Builder staticProvider(String value) {
          * @return this builder
          */
         public Builder dynamicProvider(String rubyDefaultBlock) {
```
alextwoods, line 111: Add documentation here to make it explicit that this will wrap the provided code in a proc that takes a single argument named "cfg" (and maybe even provide an example).  Otherwise I think this magic behavior would be easy to miss.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/config/ConfigProviderChain.java
```diff
@@ -109,7 +109,7 @@ public Builder staticProvider(String value) {
...
-            this.providers.add(new DynamicConfigProvider(rubyDefaultBlock));
+            this.providers.add(new DynamicConfigProvider("proc { |cfg| " + rubyDefaultBlock + " }"));
```
alextwoods, line 112: Is this a block at this point?  Or is it a ruby code fragment or something like that?

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/config/ClientConfig.java
```diff
@@ -69,6 +67,13 @@ public String getName() {
         return name;
     }
 
+    /**
+     * @return The value of the config using the config's name.
+     */
+    public String getValue() {
```
alextwoods, line 73: This method name is confusing - it does not return value.  It returns ruby code that calls the config's accessor for this config.

alextwoods (approved):

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/config/ClientConfig.java
mullermp: It's not a callable. It's the proc body

PR #167, opened 2023-10-31
Add debug logging to all middleware

mullermp:


hearth/lib/hearth/http/middleware/content_md5.rb
```diff
@@ -16,8 +16,14 @@ def initialize(app, _ = {})
         def call(input, context)
           request = context.request
           unless request.headers.key?('Content-MD5')
+            context.logger.debug(
```
alextwoods, line 19: This information is basically the same in every middleware (and is pretty minimal) - instead of needing to add log statements to every one, could we do it through the middleware stack (eg: the app gets wrapped with a logger that knows the class, and which would just log the start/end but otherwise pass through).  Or alternatively, a middleware base class?

hearth/lib/hearth/middleware/parse.rb
```diff
@@ -66,11 +66,15 @@ def call(input, context)
       private
 
       def parse_error(context, output)
+        context.logger.debug('[Middleware::Parse] Started parsing error')
         output.error = @error_parser.parse(context.response, output.metadata)
+        context.logger.debug('[Middleware::Parse] Finished parsing error')
```
alextwoods, line 71: For cases like this - should we log whether an error was parsed?

hearth/lib/hearth/middleware/retry.rb
```diff
@@ -73,20 +73,24 @@ def call(input, context)
           output.error = interceptor_error if interceptor_error
 
           if (error = output.error)
+            context.logger.debug('[Middleware::Retry] Request errored')
             error_info = @error_inspector_class.new(error, context.response)
             token = @retry_strategy.refresh_retry_token(token, error_info)
             break unless token
 
+            context.logger.debug('[Middleware::Retry] Sleeping before retry')
```
alextwoods, line 81: worth adding the amount of time we plan to sleep?

hearth/lib/hearth/middleware/retry.rb
```diff
@@ -73,20 +73,24 @@ def call(input, context)
           output.error = interceptor_error if interceptor_error
 
           if (error = output.error)
+            context.logger.debug('[Middleware::Retry] Request errored')
```
alextwoods, line 76: should we have info about the error?

hearth/lib/hearth/http/middleware/content_md5.rb
alextwoods: Edit:  in some of the middlewares there is more detailed/specific logging, but it does seem like the start/end of the middleware would be easy to have the stack log instead.

mullermp: I think it's different per middleware. I am essentially wrapping the meat of the middleware, like for Build I am wrapping the actual build part, not the start and end of middleware execution. It's also conscious of interceptor invokes. We could do one or the other.

hearth/lib/hearth/middleware/retry.rb
mullermp: Yeah, if it's not raised, we should probably log details.

hearth/lib/hearth/middleware/retry.rb
mullermp: We could yeah

hearth/lib/hearth/http/middleware/content_md5.rb
alextwoods: Yeah - thats fair.  The specialized per middleware is likely more useful.

Another thought - we only log in interceptor invoke when actual interceptors are run, not at the hook point.  Maybe worth logging that as well since its the core lifecycle of an SRA request?

alextwoods (approved):

hearth/lib/hearth/http/middleware/content_md5.rb
mullermp: I'm not sure either .. I am leaning towards keeping it as is, because the debug logging is really useful for logic that's actually executing.

PR #170, opened 2024-02-14
Rules Engine / Endpoints

alextwoods:
### Summary
Add support for generating EndpointProviders from rulesets and for using endpoint providers in the generated SDK to resolve and set endpoints on the request.

### Extension Points
The Smithy rules engine is extensible through the [EndpointRuleSetExtension](https://smithy.io/javadoc/1.44.0/software/amazon/smithy/rulesengine/language/EndpointRuleSetExtension.html) SPI - and allows for extension of Built-ins and Functions.  The smithy-ruby codegen supports this through `BuiltInBindings` and `FunctionBindings` which any RubyIntegration can register.  

#### Implementation Notes
Unlike in the C2J based V3, smithy's Rules Engine classes provide a number of abstractions for code generation.  In particular for Expressions, Templates and some standard functions.  Expression visitors have specific methods for getAttr, set, not, booleanEquals and stringEquals.  Rather than relying on stdlib ruby functions for these cases - we code generate simple ruby code to implement them.  This avoids the over head of calling an extra function when simple ruby code can be used inline.  In particular, this is beneficial for getAttr where the path is known at codegen time and smithy's path parsing utilities can be used - this avoids us needing to write/maintain the getAttr ruby implementation.

### Open questions
* What files should the endpoint generator produce? Currently it renders only an `endpoints.rb` that contains the Endpoint parameters, EndpointProvider and all of the operation specific endpoint parameter builders.  These could be separated instead into individual files, eg: `endpoints/params.rb`, `endpoints/provider.rb`, ect. 


hearth/lib/hearth.rb
```diff
@@ -38,6 +38,7 @@
 require_relative 'hearth/query/param_list'
 require_relative 'hearth/refreshing_identity_resolver'
 require_relative 'hearth/retry'
+require_relative 'hearth/rules_engine'
```
mullermp, line 41: Maybe this should actually be called endpoint_rules? It's not clear what rules_engine refers to (we know it's endpoints, though)

hearth/lib/hearth/middleware/endpoint.rb
```diff
@@ -0,0 +1,61 @@
...
+        @endpoint_provider = endpoint_provider
+        @config = kwargs
+      end
+
+      def call(input, context)
+        params = @param_builder.build(@config, input, context)
+        endpoint = @endpoint_provider.resolve_endpoint(params)
+        update_request(context, endpoint)
+
+        # TODO: combine with auth middleware
```
mullermp, line 32: We set an auth on context I believe. Maybe schemes also belongs under there somehow.

mullermp (commented):
Nice!

codegen/projections/high_score_service/lib/high_score_service/endpoint.rb
```diff
@@ -0,0 +1,108 @@
...
+    class Provider
+      def resolve_endpoint(params)
+        endpoint = params.endpoint
+
+        if (endpoint != nil)
+          return Hearth::RulesEngine::Endpoint.new(uri: endpoint)
+        end
+        raise ArgumentError, "Endpoint is not set - you must configure an endpoint."
+
+        raise ArgumentError, 'No endpoint could be resolved'
```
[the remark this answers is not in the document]
alextwoods: Yeah - in this case the defined rules have an error rule that will always be hit.  Generically however, they do not and the Smithy [specification](https://smithy.io/2.0/additional-specs/rules-engine/specification.html) on the rules engine has:

> If the list of rules is exhausted with none evaluated successfully, then an implementation of the rule set engine MUST stop and return an error indicating that rule exhaustion has occurred. Rule authors SHOULD use either an [endpoint rule](https://smithy.io/2.0/additional-specs/rules-engine/specification.html#rules-engine-endpoint-rule-set-endpoint-rule) or [error rules,](https://smithy.io/2.0/additional-specs/rules-engine/specification.html#rules-engine-endpoint-rule-set-error-rule) with an empty set of conditions to provide a more meaningful default or error depending on the scenario.

There are a few ways we could approach this:
1. [Current approach] Always add the default error case last in codegen.  It may be unreachable.
2. Attempt to detect if the ruleset always has a terminal case.  (Generically this is difficult but technically do-able)
3. Always add a default rule to endpoint ruleset.  Then codegen will take care of adding a case for it.  It may be unreachable.

We could also have the Endpoint middleware handle this case - if it returns nil, raise an exception.  However, I think the default error behavior should probably be part of the Endpoint Provider.

codegen/projections/high_score_service/lib/high_score_service/endpoint.rb
```diff
@@ -0,0 +1,108 @@
...
+          params
+        end
+      end
+
+      class UpdateHighScore
+        def self.build(config, input, context)
+          params = Params.new
+          params.endpoint = config[:endpoint]
+          params
+        end
```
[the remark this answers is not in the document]
alextwoods: In general they will be identical aside from operation specific parameters (staticContextParams and contextParams only).  

I could optimize this code and create a generic method that is called first, and then render operation specific things.  That would reduce our code size a little bit.  Thoughts?

codegen/projections/high_score_service/sig/high_score_service/endpoint.rbs
```diff
@@ -0,0 +1,20 @@
...
+# WARNING ABOUT GENERATED CODE
+
+module HighScoreService
+  module Endpoint
+    class Params < ::Struct[untyped]
+      attr_accessor endpoint: ::String
+    end
+
+    class Provider
+      def resolve_endpoint: (Params params) -> Hearth::RulesEngine::Endpoint
```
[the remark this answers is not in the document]
alextwoods: I'm a little uncertain about making this Hearth generic since its a definition thats specific to the Rules Engine.   Esp the `AuthScheme` part - these are specific to RulesEngine view of AuthSchemes rather than our generic SRA defined auth schemes....

codegen/smithy-ruby-codegen-test/model/test-services/default-values.smithy
```diff
@@ -0,0 +1,111 @@
+$version: "2.0"
+
+namespace rulesengine.defaultvalues
```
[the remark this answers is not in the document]
alextwoods: yup - coming soon!  Wanted to just add enough to ensure generation/testing was working to start.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/EndpointGenerator.java
```diff
@@ -0,0 +1,684 @@
...
+                    .write("")
+                    .call(() -> renderRbsEndpointProvider(writer))
+                    .closeAllModules();
+        });
+        LOGGER.fine("Wrote auth rbs module to " + rbsFile());
+    }
+
+    private void renderEndpointParamsClass(RubyCodeWriter writer) {
+        List<String> params = new ArrayList<>();
+        Map<String, String> defaultParams = new LinkedHashMap<>();
```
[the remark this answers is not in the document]
alextwoods: This ensures a stable ordering and prevents issues with the order of parameters jumping around in diffs - which causes a bunch of churn on diffs that make it harder to see real changes.

hearth/lib/hearth.rb
alextwoods: Yeah - this was one of my open questions on the PR description.

I started with `RulesEngine` to match the naming used by Smithy for all of these.  But I agree - it is all endpoints so could be an Endpoint module instead.

hearth/lib/hearth/middleware/endpoint.rb
alextwoods: Yeah, somehow the properties of matchign auth schemes need to "override" or be merged onto the already resovled auth on context (eg: for sigv4, replacing the region with signignRegion).

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/GenerationContext.java
```diff
@@ -83,9 +104,55 @@ public GenerationContext(RubySettings rubySettings,
         this.protocolGenerator = protocolGenerator;
         this.applicationTransport = applicationTransport;
         this.symbolProvider = symbolProvider;
+        this.rulesEngineBuiltIns = rulesEngineBuiltIns.stream().collect(Collectors.toMap(
+                        (b) -> b.getBuiltIn().getBuiltIn().get(),
+                        (b) -> b
```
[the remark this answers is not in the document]
alextwoods: Its creating a map from the list - we could isntead of course multiline this and not use streams with something like:
```java
this.rulesEngineBuiltIns = new HashMap<>()
for (val b : rulesEngineBuildIns) {
  this.rulesEngineBuiltIns.put( b.getBuiltIn().getBuiltIn().get(), b)
}
```

I'm fine either way if you have a preference.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/GenerationContext.java
```diff
@@ -83,9 +104,55 @@ public GenerationContext(RubySettings rubySettings,
...
+        this.modeledClientConfig = new ArrayList<>();
+        service.getTrait(ClientContextParamsTrait.class).ifPresent((clientContext -> {
+            clientContext.getParameters().forEach((name, param) -> {
+                ClientConfig.Builder builder = ClientConfig.builder()
+                        .name(RubyFormatter.toSnakeCase(name));
+
+                param.getDocumentation().ifPresent((d) -> builder.documentation(d));
+                builder.type(symbolProvider.toSymbol(
+                                param.getType().createBuilderForType()
+                                        .id("smithy#temp")
```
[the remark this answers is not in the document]
alextwoods: Extracted this to a method and added comments.  Maps a client context param to the ruby type that we need to use for the generated client config (for documentatation + validation)

codegen/projections/high_score_service/lib/high_score_service/endpoint.rb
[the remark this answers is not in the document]
alextwoods: For now I've gone ahead with a variation of 2 above - specifically if the last rule does not have any conditions, then we know for sure we do not need the fallback error (Because that rule is guaranteed to be hit).   This prevents unreachable fallback error in a lot of cases, but of course not generically (there are complex rulesets that will always reach a leaf, but which don't have a last rule w. empty conditions).

codegen/projections/high_score_service/lib/high_score_service/endpoint.rb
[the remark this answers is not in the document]
mullermp: It's fine as is, assuming it could be different.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/GenerationContext.java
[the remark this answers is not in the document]
mullermp: For Java code, I don't. It just looked funky. I'm more concerned about generated output.

hearth/lib/hearth.rb
mullermp: Yeah. I vote EndpointRules or even Endpoints (preferred?) as a module name. Maybe evaluate what that would look like globally and see if it makes sense.

jterapin: 1+ on renaming to Endpoint-something. As someone somewhat new to the codebase, I would not have known that this file is all for endpoints

hearth/lib/hearth/rules_engine.rb
```diff
@@ -0,0 +1,133 @@
...
+
+require_relative 'structure'
+
+module Hearth
+  # Functions in the Smithy rules engine are named routines that
+  # operate on a finite set of specified inputs, returning an output.
+  # The rules engine has a set of included functions that can be
+  # invoked without additional dependencies, called the standard library.
+  # @api private
+  module RulesEngine
```
jterapin, line 15: Re: What files should the endpoint generator produce?

In spirit of following [Ruby Style Guide](https://github.com/rubocop/ruby-style-guide?tab=readme-ov-file#one-class-per-file), I think it is worth exploring having separate files for each classes. A single file is manageable right now but could this grow in the future? If so, we could keep them separated.

hearth/lib/hearth.rb
alextwoods: See: https://github.com/smithy-lang/smithy-ruby/pull/170/commits/e7342a768cdf0176374ce7d2019067ac8fe0c874 for what the rename looks like.

I'm fine with using `Endpoints` as the module name since its descriptive of what its actually used for - I was mostly initially trying to match Smithy's namespace.

hearth/lib/hearth/rules_engine.rb
mullermp: It depends. Sometimes anonymous/api private classes within another class or module are perfectly fine.

codegen/projections/high_score_service/sig/high_score_service/endpoint.rbs
```diff
@@ -14,7 +14,7 @@ module HighScoreService
     end
 
     class Provider
-      def resolve_endpoint: (Params params) -> Hearth::RulesEngine::Endpoint
+      def resolve_endpoint: (Params params) -> Hearth::Endpoints::Endpoint
```
mullermp, line 17: After seeing this, `Hearth::Endpoints::Endpoint` seems weird. I think `Hearth::EndpointRules::Endpoint` to make clear that it's an endpoint object as defined in the endpoint ruels, but I'll defer to you.

codegen/projections/white_label/spec/client_spec.rb
```diff
@@ -146,12 +146,12 @@ module WhiteLabel
         end
 
         it 'uses endpoint from options' do
-          proc = proc do |context|
+          verify = proc do |context|
```
mullermp, line 149: Was this the case? If so, can you (in a separate PR I guess?) update all occurrences?

codegen/projections/white_label/spec/endpoint_spec.rb
```diff
@@ -0,0 +1,204 @@
...
+          client.dataplane_operation({}, interceptors: [interceptor])
+        end
+
+        it 'produces the correct output from the client when calling endpoint_with_host_label_operation' do
+          config = {}
+          config[:stub_responses] = true
+          config[:endpoint] = nil
+
+          client = Client.new(config)
+          interceptor = Hearth::Interceptor.new(read_before_transmit: proc do |context|
```
mullermp, line 189: I prefer multi-step variables for cleanliness here but I'll defer to you.

mullermp (approved):

codegen/projections/white_label/spec/client_spec.rb
alextwoods: It doesn't actually break anything - I'll switch it back to `proc` since thats what we've used everywhere.

codegen/projections/white_label/spec/endpoint_spec.rb
alextwoods: yeah, we use an intermediate elsewhere, I'll update to follow that pattern.

codegen/projections/white_label/spec/client_spec.rb
mullermp: Ok. It could have been my use of `proc = Proc.new`. I don't remember the exact issue I had.

PR #174, opened 2024-03-07
RBS improvements

### LINKED ISSUE #172: RBS improvements
RE: @ksss

# First Impressions

## Steepfile

There is no `#check` in the Steepfile.
Running `steep check` in this state doesn't seem much different from `rbs validate`. Is this intentional?

If the intention is only to validate the `sig` directory, I recommend using `rbs validate`.

## Top-level Type Aliases

The top-level type aliases in `hearth/sig/lib/hearth/aliases.rbs`, such as `type document`, are very likely to conflict with names in other libraries. They should be namespaced.

## client.rbs

In client.rbs, the type is defined as `(?::Hash[::Symbol, untyped] params, ?::Hash[::Symbol, untyped] options)-> Hearth::Output`, but using the overload of keyword arguments and `Hash[::Symbol, untyped]` as done in aws-sdk-ruby could safely increase the information about the arguments.

```rbs
class RailsJson::Client
  def all_query_string_types: (query_string: String, query_string_list: Array[String], ...snip..., ?::Hash[::Symbol, untyped] options)-> Hearth::Output
         | (?::Hash[::Symbol, untyped] params, ?::Hash[::Symbol, untyped] options)-> Hearth::Output

### PR DESCRIPTION
**Pull request by mullermp**
Fixes #172

### CODE REVIEW

File: codegen/projections/high_score_service/sig/high_score_service/client.rbs [hunk 1]
```diff
@@ -17,14 +17,14 @@ module HighScoreService
...
-    def api_key_auth: (?::Hash[::Symbol, untyped] params, ?::Hash[::Symbol, untyped] options)-> Hearth::Output
-    def basic_auth: (?::Hash[::Symbol, untyped] params, ?::Hash[::Symbol, untyped] options)-> Hearth::Output
-    def bearer_auth: (?::Hash[::Symbol, untyped] params, ?::Hash[::Symbol, untyped] options)-> Hearth::Output
-    def create_high_score: (?::Hash[::Symbol, untyped] params, ?::Hash[::Symbol, untyped] options)-> Hearth::Output
-    def delete_high_score: (?::Hash[::Symbol, untyped] params, ?::Hash[::Symbol, untyped] options)-> Hearth::Output
-    def digest_auth: (?::Hash[::Symbol, untyped] params, ?::Hash[::Symbol, untyped] options)-> Hearth::Output
-    def get_high_score: (?::Hash[::Symbol, untyped] params, ?::Hash[::Symbol, untyped] options)-> Hearth::Output
-    def list_high_scores: (?::Hash[::Symbol, untyped] params, ?::Hash[::Symbol, untyped] options)-> Hearth::Output
-    def update_high_score: (?::Hash[::Symbol, untyped] params, ?::Hash[::Symbol, untyped] options)-> Hearth::Output
+    def api_key_auth: (?::Hash[::Symbol, untyped] params, ?::Hash[::Symbol, untyped] options)-> Hearth::Output[Types::ApiKeyAuthOutput]
```
mullermp, line 20: Is this correct? Output can have either an error or a type, under the accessor `data` or `error` - I think this should be as is?

hearth/sig/lib/hearth/configuration.rbs
```diff
@@ -2,6 +2,7 @@ module Hearth
   module Configuration
     def initialize: (**Hash[Symbol, untyped] options) -> void
 
-    def dup: () -> Hearth::Configuration
+    def merge: (Hash[Symbol, untyped] configuration) -> Configuration
+    def dup: () -> Configuration
```
mullermp, line 6: I don't think `dup` is a method anymore for this

hearth/sig/lib/hearth/interfaces.rbs
```diff
@@ -62,4 +62,8 @@ interface _ErrorInspector
   def error_type: () -> String
 
   def hints: () -> Hash[Symbol, untyped]
+end
+
+interface _WritableIO
```
mullermp, line 67: I think we need to namespace all of these

hearth/sig/lib/hearth/configuration.rbs
alextwoods: yeah, good call, we did remove that.

Related - I think the `-> Configuration` isn't actually quite correct here.  It returns an object of the class that includes this module.  I'm not quite sure how to correctly express that?

codegen/projections/high_score_service/sig/high_score_service/client.rbs
alextwoods: I think this is correct - Output is defined with a paramtraized type (`DATA`) which is the type of the `data` accessor - or more specially: `attr_accessor data: (DATA? | nil)` which is true.  The output's data method will always either return nil or the operation's output type.

mullermp: Hmm ok I couldn't remember if that change was still in the branch. Also I think DATA? Is sufficient because it's optional.

mullermp: I am thinking that parameterized is wrong though because we don't delegate to DATA like we do in v3. I think it should be reverted still. It depends on what the output looks like.

alextwoods: I think parametrized is correct - delegation doesn't matter for that.    Hearth::Output is generic.  Each client method returns Hearth::Output with a concrete known type for the `data` accessor:

```
client.get_high_score(...).data&.high_score # we know that if data is not nil, then it has the high_score method
```

codegen/projections/rails_json/Steepfile
```diff
@@ -3,8 +3,8 @@
 D = Steep::Diagnostic
 
 target :app do
-  signature 'sig'
   library 'hearth'
+  signature 'sig'
```
alextwoods, line 7: As I was working on various fixes I was trying to find a good set of type checks - I think we may want to do something like what we do in v3 using rbs:test. I think that gives the closest to validating what we care about - validation of the end user public types. 

I did run into issues with double that I can't figure out how to solve without just not using doubles (ie creating a client and mocking vs passing a double).

hearth/sig/lib/hearth/configuration.rbs
```diff
@@ -1,7 +1,7 @@
 module Hearth
-  module Configuration[T]
+  module Configuration
     def initialize: (**Hash[Symbol, untyped] options) -> void
 
-    def merge: (Hash[Symbol, untyped] configuration) -> T
+    def merge: (Hash[Symbol, untyped] configuration) -> Configuration
```
alextwoods, line 5: I'm pretty sure this still needs to be parameterized - this does not return a Configuration - it returns the service's Config class.

hearth/sig/lib/hearth/interfaces.rbs
```diff
@@ -10,52 +10,53 @@ end
 class Logger
 end
 
-interface _Plugin[T]
-  def call: (Hearth::Configuration[T]) -> void
-end
+module Hearth
+  interface _Plugin
+    def call: (Configuration) -> void
```
alextwoods, line 15: Ditto here - this takes a service's Config - so I think it should be paramterized.

hearth/sig/lib/hearth/plugin_list.rbs
```diff
@@ -1,15 +1,15 @@
 module Hearth
-  class PluginList[T]
-    def initialize: (?Array[_Plugin[T]] plugins) -> void
+  class PluginList
```
alextwoods, line 2: PluginLists for a given service must have plugins that operate on the service's Config object, so I think this should also be parameterized to specify that constraint.

codegen/projections/rails_json/Steepfile
alextwoods: Update for posterity: move to a v3 like rbs:test in CI.

alextwoods (approved):

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/RubySymbolProvider.java
```diff
@@ -318,14 +318,14 @@ public Symbol serviceShape(ServiceShape shape) {
...
-        return createSymbolBuilder(shape, name, name, name, moduleName)
+        return createSymbolBuilder(shape, name, "Types::" + name, name, moduleName)
                 .definitionFile(DEFAULT_DEFINITION_FILE).build();
     }
 
     @Override
     public Symbol unionShape(UnionShape shape) {
         String name = getDefaultShapeName(shape, "Union__");
-        return createSymbolBuilder(shape, name, name, name, moduleName)
+        return createSymbolBuilder(shape, name, "Types::" + name, name, moduleName)
```
mullermp, line 328: Is this correct? Some times we may not want to print out a full qualified name. I would assume it's relative to where it's being written.

codegen/projections/white_label/Rakefile
```diff
@@ -0,0 +1,30 @@
+require 'rubygems'
```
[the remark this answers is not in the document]
alextwoods: I'm not sure - its specific to whitelabel (in namespace)

codegen/projections/white_label/spec/errors_spec.rb
```diff
@@ -35,7 +35,7 @@ module Errors
       it 'inherits the base client api error' do
         http_resp = Hearth::HTTP::Response.new
         error = ApiRedirectError.new(
-          location: nil,
+          location: 'location',
```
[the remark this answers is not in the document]
alextwoods: I'm not positive -  https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Location states:
> All responses with one of these status codes send a Location header.

So I had assumed that location must actually be set on these.

codegen/projections/white_label/spec/types_spec.rb
```diff
@@ -51,10 +51,10 @@ module Types
       it 'implements to_h for delegation' do
         string_union = Union::String.new('simple string')
         struct_union = Union::Struct.new(struct)
-        unknown_union = Union::Unknown.new({ key: 'key', value: 'value' })
+        unknown_union = Union::Unknown.new({ name: 'name', value: 'value' })
```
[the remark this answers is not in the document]
alextwoods: Key was wrong from the POV of how our code actually creates these (we always create them using name and value) - but nothing in the generated Unknown class actually restricts the data that it holds.  

We could change how we generate the `Unknown` union member.  Right now it is:

```ruby
      class Unknown < Union
        def to_h
          { unknown: super(__getobj__) }
        end

        def to_s
          "#<WhiteLabel::Types::Unknown #{__getobj__ || 'nil'}>"
        end
      end
      
      # and the Hearth::Union is just a delegator
module Hearth
  # Top level class for all Union types
  class Union < ::SimpleDelegator
    include Hearth::Structure
  end
end
```

Instead of just allowing whatever data, we could write a class that intializes with kwargs that only allow name and value?

codegen/projections/white_label/spec/errors_spec.rb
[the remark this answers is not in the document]
mullermp: Makes sense, so probably shouldn't be optional.

codegen/smithy-ruby-codegen-test/Rakefile
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+# Used to test endpoint and other models from Smithy
```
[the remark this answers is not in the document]
alextwoods: I dont think so - this works on a different context.  We could jam it together, we just would then need to make sure to never run these tests from the whitelabel.  And vice-versa.

codegen/smithy-ruby-codegen-test/Rakefile_typecheck
```diff
@@ -0,0 +1,30 @@
+require 'rubygems'
```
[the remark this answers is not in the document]
alextwoods: We do just copy this rakefile?

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/RubySymbolProvider.java
alextwoods: Yeah - I was debating this.  The `rbsType` property doesn't have the same namespace sensitive ability as the actual symbol name.  I think `Types::<name>` will always be correct (it just may be more information than necessary if you use it within the types module)

codegen/smithy-ruby-codegen-test/Rakefile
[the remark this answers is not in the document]
alextwoods: RE: All of the rakefile comments - I'll take a broader look at what we can do here in a new PR.  I think we need to throw all of our existing CI away and redo it if we want a more unified, single Rakefile

codegen/projections/high_score_service/sig/high_score_service/client.rbs
```diff
@@ -11,20 +11,88 @@ module HighScoreService
...
+        ?http_api_key_identity_resolver: Hearth::IdentityResolver,
+        ?http_bearer_identity_resolver: Hearth::IdentityResolver,
+        ?http_client: Hearth::HTTP::Client,
+        ?http_login_identity_resolver: Hearth::IdentityResolver,
+        ?interceptors: Hearth::InterceptorList,
+        ?logger: Logger,
+        ?plugins: Hearth::PluginList[Config],
+        ?retry_strategy: Hearth::_RetryStrategy,
+        ?stub_responses: bool,
+        ?validate_input: bool      ) -> void
```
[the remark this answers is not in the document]
alextwoods: Fixed!

codegen/projections/high_score_service/sig/high_score_service/client.rbs
```diff
@@ -11,20 +11,88 @@ module HighScoreService
...
-    def delete_high_score: (?::Hash[::Symbol, untyped] params, ?::Hash[::Symbol, untyped] options)-> Hearth::Output
-    def digest_auth: (?::Hash[::Symbol, untyped] params, ?::Hash[::Symbol, untyped] options)-> Hearth::Output
-    def get_high_score: (?::Hash[::Symbol, untyped] params, ?::Hash[::Symbol, untyped] options)-> Hearth::Output
-    def list_high_scores: (?::Hash[::Symbol, untyped] params, ?::Hash[::Symbol, untyped] options)-> Hearth::Output
-    def update_high_score: (?::Hash[::Symbol, untyped] params, ?::Hash[::Symbol, untyped] options)-> Hearth::Output
+    def api_key_auth: (?::Hash[::Symbol, untyped] params, ?::Hash[::Symbol, untyped] options)  -> Hearth::Output[Types::ApiKeyAuthOutput] |
+      (?Types::ApiKeyAuthInput params, ?::Hash[::Symbol, untyped] options)  -> Hearth::Output[Types::ApiKeyAuthOutput] |
+      (
+
+      )  -> Hearth::Output[Types::ApiKeyAuthOutput]
```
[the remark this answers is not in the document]
alextwoods: Removed

codegen/projections/white_label/spec/types_spec.rb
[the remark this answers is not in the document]
alextwoods: Updated Union Unknown to constraint to name/value only.

PR #175, opened 2024-03-14
Refactor client middleware into a module

mullermp:


codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/MiddlewareGenerator.java
```diff
@@ -0,0 +1,100 @@
...
+                    .includeRequires()
+                    .writeRequireRelativeAdditionalFiles(additionalFiles)
+                    .addModule(settings.getModule())
+                    .addModule("Middleware")
+                    .call(() -> renderOperations(writer))
+                    .write("")
+                    .closeAllModules();
+        });
+
+        LOGGER.fine("Wrote client to " + rbFile());
```
alextwoods, line 76: We can probably remove all of these loggers (or fix client -> middleware).

alextwoods (approved):

PR #176, opened 2024-03-18
Add support for parsing lists of strings/timestamps in headers

alextwoods:
*Description of changes:*
Adds support for parsing lists of strings/timestamps in headers. 

Forward port of: https://github.com/aws/aws-sdk-ruby/pull/2988/commits/bc243c432748d52747cf2e282434b063921f494f


codegen/projections/rails_json/lib/rails_json/parsers.rb
```diff
@@ -402,14 +398,11 @@ def self.parse(http_resp)
         end
         unless http_resp.headers['X-TimestampList'].nil? || http_resp.headers['X-TimestampList'].empty?
           data.header_timestamp_list = http_resp.headers['X-TimestampList']
-            .split(', ')
-            .map { |s| Time.parse(s) }
+            .split(',').each_slice(2).map { |v| Time.parse(v[0] + v[1])}
```
mullermp, line 401: Header list parser should maybe handle Time case too? It should probably also call methods in TimeHelper

codegen/smithy-ruby-rails-codegen-test/model/protocol-test/http-headers.smithy
```diff
@@ -109,13 +108,6 @@ apply InputAndOutputWithHeaders @httpRequestTests([
     },
 ])
 
-apply InputAndOutputWithHeaders @skipTests([
```
mullermp, line 112: I thought there were two tests for this one, I could be wrong.

hearth/lib/hearth/http/header_list_parser.rb
```diff
@@ -0,0 +1,62 @@
...
+      def self.parse_string_list(value)
+        buffer = StringScanner.new(value)
+        parsed = []
+
+        parsed << read_value(buffer) until buffer.eos?
+
+        parsed
+      end
+
+      def self.read_value(buffer)
```
mullermp, line 21: Should all of these be private?

codegen/smithy-ruby-rails-codegen-test/model/protocol-test/http-headers.smithy
alextwoods: You're right - I think we just never ported it over.  I'll add it.

hearth/lib/hearth/http/header_list_parser.rb
alextwoods: Yeah - updated. Rubocop had removed the private for those (since private doesn't actually work on def self.method).  Moved to class << self style.

codegen/projections/rails_json/lib/rails_json/parsers.rb
alextwoods: Yeah, move this case to a method in the header list parser since its pretty weird logic.

And used the TimeformatHelper for the other cases.

mullermp (approved):

PR #178, opened 2024-03-22
Responds to constraint + refactors

mullermp:
* Add responds to constraint
* Naming refactors
* Client config refactoring
* more to come


codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/config/RespondsToConstraint.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+package software.amazon.smithy.ruby.codegen.config;
+
+import java.util.List;
+import java.util.stream.Collectors;
+import software.amazon.smithy.ruby.codegen.RubyFormatter;
+
+/**
+ * Range Constraint for config value.
```
alextwoods, line 23: update docs

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/middleware/factories/AuthMiddlewareFactory.java
```diff
@@ -127,9 +130,11 @@ private static ClientConfig buildAuthResolverConfig() {
                 """.formatted(Hearth.AUTH_OPTION);
         return ClientConfig.builder()
                 .name("auth_resolver")
-                .type("Auth::Resolver")
-                .documentation(authResolverDocumentation)
+                .type("#resolve(params)")
```
alextwoods, line 133: should be documentationType

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/config/ClientConfig.java
```diff
@@ -181,7 +181,6 @@ public Builder name(String name) {
          */
         public Builder type(String type) {
```
alextwoods, line 182: After seeing the changes - I would recommend:
- change this method to validateType and have it add a TypeConstraint (instead of needing to do `.constraint(new TypeConstraint)`).  
- remove `type` from the builder.  We shouldn't have it - its confusing.  There is no "type" for config, theres multiple different contexts for type.
- Remove the inference logic in the constructor and require that documetnationType AND rbsType are specified.  

this should create a consistent, explicit, easy to read/use builder.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/config/ClientConfig.java
```diff
@@ -42,8 +42,8 @@ public class ClientConfig {
     public ClientConfig(Builder builder) {
         this.name = builder.name;
         this.documentation = builder.documentation;
-        this.documentationType = builder.documentationType != null ? builder.documentationType : builder.type;
-        this.rbsType = builder.rbsType != null ? builder.rbsType : builder.type;
+        this.documentationType = builder.documentationType;
+        this.rbsType = builder.rbsType != null ? builder.rbsType : builder.documentationType;
```
alextwoods, line 46: I'm thinking we should do away with this as well

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/config/ClientConfig.java
```diff
@@ -195,21 +184,21 @@ public Builder documentation(String documentation) {
         }
 
         /**
-         * @param defaultValue an optional default value to be use in documentation.
+         * @param type an optional type to use in documentation (defaults to the type).
```
alextwoods, line 187: Documentation nit: does not default to type anymore

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/auth/factories/HttpBasicAuthSchemeFactory.java
```diff
@@ -37,13 +38,14 @@ public static AuthScheme build() {
...
                 .name("http_login_identity_resolver")
-                .type(Hearth.IDENTITY_RESOLVER.toString())
                 .documentation(
                         identityResolverDocumentation.formatted(
                                 Hearth.IDENTITY_RESOLVER,
                                 identityType,
                                 HttpBasicAuthTrait.ID))
+                .documentationType(Hearth.IDENTITY_RESOLVER.toString())
                 .defaultDynamicValue(defaultConfigValue)
+                .constraint(new TypeConstraint(Hearth.IDENTITY_RESOLVER.toString()))
```
alextwoods, line 48: I'd still lean towards just adding a convince builder method for `validateType`. 

Another convenience builder method we could add: `validationDocumentationAndRbsType(oneTypeForAll)`

Most config we have follows this - and its better in those cases to ensure that they all match and are correct.

PR #180, opened 2024-03-28
Fix retrying with IO objects and use JSON.parse

mullermp:
Fix retrying with IO objects and correctly rewind them

Use JSON.parse instead of load


alextwoods (approved):
Should we add some integration tests into whitelabel (or into rails_json?) that let us cover these cases e2e?

PR #187, opened 2024-04-22
Remove identity from identity provider config names

alextwoods:
*Description of changes:*
Remove identity from identity provider config names.


codegen/smithy-ruby-codegen-test/integration-specs/auth_spec.rb
```diff
@@ -267,7 +267,7 @@ module Auth
 
     describe '#http_bearer_auth' do
       let(:config_hash) do
-        { http_bearer_identity_resolver: identity_resolver }
+        { http_bearer_provider: identity_resolver }
```
mullermp, line 270: Do a generic search for identity_resolver too

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/auth/factories/HttpBasicAuthSchemeFactory.java
```diff
@@ -21,35 +21,29 @@
 import software.amazon.smithy.ruby.codegen.config.ClientConfig;
 
 public final class HttpBasicAuthSchemeFactory {
+
+    public static final ClientConfig HTTP_LOGIN_PROVIDER = ClientConfig.builder()
```
mullermp, line 25: Maybe there should be factories for all of the providers?

codegen/projections/white_label/lib/white_label/config.rb
```diff
@@ -174,7 +174,7 @@ def defaults
         http_api_key_provider: [proc { |cfg| cfg[:stub_responses] ? Hearth::IdentityProvider.new(proc { Hearth::Identities::HTTPApiKey.new(key: 'stubbed api key') }) : nil }],
         http_bearer_provider: [proc { |cfg| cfg[:stub_responses] ? Hearth::IdentityProvider.new(proc { Hearth::Identities::HTTPBearer.new(token: 'stubbed bearer') }) : nil }],
         http_client: [proc { |cfg| Hearth::HTTP::Client.new(logger: cfg[:logger]) }],
-        http_custom_auth_provider: [proc { |cfg| cfg[:stub_responses] ? Hearth::IdentityProvider.new(proc { Auth::HTTPCustomAuthIdentity.new(key: 'key') }) : nil }],
+        http_custom_auth_identity_provider: [proc { |cfg| cfg[:stub_responses] ? Hearth::IdentityProvider.new(proc { Auth::HTTPCustomAuthIdentity.new(key: 'key') }) : nil }],
```
mullermp, line 177: Revert?

codegen/smithy-ruby-codegen-test/integration-specs/auth_spec.rb
alextwoods: yea, fixed.  Was only in this test

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/auth/factories/HttpBasicAuthSchemeFactory.java
alextwoods: I'm not sure its worth creating separate classes for all for all of the client configs since they aren't generally shared.  This was a bit of a special case.  But maybe it should be its own file?

codegen/projections/white_label/lib/white_label/config.rb
alextwoods: The name of the CustomAuthIdentity has identity in it and I think it should, so I included it here since it would otherwise be "auth_provider" which sounds like its providing auth.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/auth/factories/HttpBasicAuthSchemeFactory.java
mullermp: It's probably fine as is

mullermp (approved):

PR #190, opened 2024-04-23
Move stubs to Config + Hearth::ClientBase

alextwoods:
*Description of changes:*
Moves stubs from a client instance variable to Config.  This removes special casing around the handling of `stubs` in middleware builders.   In addition, it allows for specifying Stubs when creating a Client.  This requires making `Stubs` a public interface.  Adds special case logic to `operation_config` to prevent `stub_responses` or `stubs` from being overridden on operation calls.    Renames `add_stubs` to `set_stubs` to make it clear that setting stubs overwrites existing stubs and is NOT additive.

Adds a new Hearth::Client with the common generated methods, including operation_config.  This removes duplication of code across all generated clients.


hearth/lib/hearth/stubs.rb
```diff
@@ -3,14 +3,13 @@
 module Hearth
   # Provides a thread safe data structure for adding and getting stubs
   # per operation.
-  # @api private
   class Stubs
```
mullermp, line 6: Do we want a private Stub class to wrap the stubs? That way our `send` middleware is less exhaustive with setting stubs and possibly with testing too? The Stub class could probably validate that stubs are a proc, or Response, or hash, or error, etc?

hearth/spec/hearth/stubbing/client_stubs_spec.rb
```diff
@@ -2,18 +2,20 @@
...
 
       def initialize(config = Config.new)
         @config = config
-        @stubs = Hearth::Stubs.new
+        @config.stubs = Hearth::Stubs.new
       end
 
       # for testing
-      attr_reader :stubs
+      def stubs
```
mullermp, line 16: Probably makes sense to actually put the attr_reader :config here instead (our clients have this anyway, so no comment needed "for testing" and can just retrieve stubs that way)

mullermp (commented):
I think it makes sense. Let's chat offline

hearth/lib/hearth/client_base.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    private
+
+    def initialize_config(options, client_class)
+      client_interceptors = options.delete(:interceptors)
+      config = client_class.new(**options)
+      config.validate!
+      self.class.plugins.each { |p| p.call(config) }
+      config.plugins.each { |p| p.call(config) }
+      config.interceptors.concat(client_interceptors) if client_interceptors
```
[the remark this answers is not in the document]
alextwoods: None of these can be empty (except the client interceptors from above) so I'd prefer to not use safe nav in those cases.  What do you think?

hearth/lib/hearth/stubs.rb
alextwoods: Yeah, a stub class could potentially do validation here, but lets consider in a separate PR since this is already a pretty big change.

hearth/lib/hearth/client.rb
```diff
@@ -0,0 +1,64 @@
...
+              'operations is not allowed'
+        raise ArgumentError, msg
+      end
+
+      operation_plugins = options.delete(:plugins)
+      operation_interceptors = options.delete(:interceptors)
+      config = @config.merge(options)
+      config.validate!
+      operation_plugins&.each { |p| p.call(config) }
+      if operation_interceptors
```
mullermp, line 50: I suppose we could default interceptors here too, so concat always succeeds.

hearth/sig/lib/hearth/client.rbs
```diff
@@ -0,0 +1,9 @@
+module Hearth
+  class Client
+    def initialize: (Hash[Symbol, untyped], Class) -> void
+
+    attr_reader config: untyped
```
mullermp, line 5: I think this would be parameterized right? Same for PluginList below?

mullermp (approved):
Nice! I think a component test for stubs in white label for all the modes (client, instance, and operation) are needed. You have most of those pieces already I think..

codegen/projections/rails_json/lib/rails_json/client.rb
```diff
@@ -25,8 +24,7 @@ def self.plugins
     # @param [Hash] options
     #   Options used to construct an instance of {Config}
     def initialize(options = {})
```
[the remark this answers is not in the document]
alextwoods: Ah right - I needed to keep the definition in the RBS file for it (to get the correct typing), but you are right it can be deleted from here.

hearth/sig/lib/hearth/client.rbs
alextwoods: I couldn't figure out how to correctly parameterize it on the Class level.  Open to ideas.... The actual generated client classes all end up with correct typing.

codegen/projections/rails_json/lib/rails_json/client.rb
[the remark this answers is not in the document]
mullermp: hmmm.. RBS doesn't follow the inheritance in IDE? Eventually that will be the case maybe ..

hearth/sig/lib/hearth/client.rbs
mullermp: If they have the correct typing this this, then I think that's fine as is.

codegen/projections/rails_json/spec/stubs_spec.rb
```diff
@@ -1,101 +1,101 @@
-# # frozen_string_literal: true
```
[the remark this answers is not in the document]
alextwoods: You did that already I thought?

hearth/lib/hearth/client.rb
alextwoods: Yeah, I think that makes sense.

PR #191, opened 2024-04-24
Refactor logic for building/parsing header lists to hearth

jterapin:
*Issue #, if available:*
During Smithy-Ruby Wiki Traits review, I came across the following documentation for the `@httpHeader` trait:
> When the trait is applied to a list or set, the protocol builder will append the values to the header with commas.
```ruby
class Operation
  def self.build(http_req, input:)
    ...
    http_req.headers['X-List'] = input[:foo].join(',')
  end
end
```

However - in the rails_json projection, the building of header lists are handled like follows:
```ruby
class InputAndOutputWithHeaders
  def self.build(http_req, input:)
    ...
    http_req.headers['X-StringList'] = input[:header_string_list] unless input[:header_string_list].nil? || input[:header_string_list].empty?
  end
end
```
[Full code snippet are here.](https://github.com/smithy-lang/smithy-ruby/blob/c02978090ca4cd862c4baf7cab92a115ec1bb517/codegen/projections/rails_json/lib/rails_json/builders.rb#L518)

The handling of the header lists are handled in the [core](https://github.com/smithy-lang/smithy-ruby/blob/c02978090ca4cd862c4baf7cab92a115ec1bb517/hearth/lib/hearth/http/field.rb#L31-L36).

I believe this was changed to be aligned with the guidance set with Smithy-based SDKs.

However, one could argue that the building of header lists should be handled by the RestBuilderGenerator:
* Builder has a full context of the shapes and therefore, exactly know how it should be serialized (i.e. list of string vs list of timestamps).
* Core should remain generic. However, the serialization of header lists follows the [HTTP binding serialization](https://smithy.io/2.0/aws/protocols/aws-restjson1-protocol.html#http-binding-serialization) and therefore should be moved to the Builder.

This PR showcases scenario where we relocate the building of header lists to the RestBuilderGenerator

--- 


codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/RestBuilderGeneratorBase.java
```diff
@@ -433,16 +433,26 @@ protected HeaderListMemberSerializer(String inputGetter, String dataSetter, Memb
 
         @Override
         protected Void getDefault(Shape shape) {
```
mullermp, line 435: This change should be affecting rest json too right? Are there any cases of that?

hearth/lib/hearth/http/field.rb
```diff
@@ -5,9 +5,8 @@ module HTTP
     # Represents an HTTP field.
     class Field
       # @param [String] name The name of the field.
-      # @param [Array|#to_s] value (nil) The values for the field. It can be any
-      #   object that responds to `#to_s` or an Array of objects that respond to
-      #  `#to_s`.
+      # @param [String|#to_s] value (nil) The value for the field. It can be any
```
mullermp, line 8: I think just #to_s here.

hearth/lib/hearth/http/field.rb
```diff
@@ -25,15 +24,10 @@ def initialize(name, value = nil, kind: :header)
...
       # @return [String]
       def value(encoding = nil)
-        value =
-          if @value.is_a?(Array)
-            @value.compact.map { |v| escape_value(v.to_s) }.join(', ')
-          else
-            @value.to_s
-          end
+        value = @value.to_s
         value = value.encode(encoding) if encoding
```
mullermp, line 31: Can simplify @value.to_s.encode

hearth/lib/hearth/http/field.rb
```diff
@@ -53,9 +47,9 @@ def to_h
         { @name => value }
       end
 
-      private
-
-      def escape_value(str)
+      # Escapes header field value
+      # @return [String]
+      def self.escape_value(str)
```
mullermp, line 52: I think this could be moved to a HeaderListBuilder

hearth/spec/hearth/http/field_spec.rb
```diff
@@ -42,22 +42,6 @@ module HTTP
           end
         end
 
-        context 'value is an Array' do
```
mullermp, line 45: `context 'value is a Scalar type'` can be demoted (since value will only ever be scalar type, no context necessary)

mullermp (commented):
Good start. We need to change the RBS for Field too.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/RestBuilderGeneratorBase.java
jterapin: I'm going to assume that you are referring to protocol-test cases? Yeah, there's Rest-Json versions of [these test cases](https://github.com/smithy-lang/smithy-ruby/blob/702f6c5012e6dc439d7b9dcc4b59d1e3df306515/codegen/smithy-ruby-rails-codegen-test/model/protocol-test/http-headers.smithy#L30-L137)

alextwoods: I think v4 also won't need an update for this change since restjson/restxml rely on this base class, so I think those test cases shoudl continue to pass.   (once we get v4 protocol tests passing, I'd like to add a "downstream" test to the CI so that we at least know when our smithy-ruby changes break v4 and require a change there).

mullermp: I was asking if the rest json builders will change. It should generate new code for that right?

hearth/lib/hearth/http/header_list_builder.rb
```diff
@@ -0,0 +1,33 @@
...
+  module Http
+    # @api private
+    module HeaderListBuilder
+      class << self
+        # builds a string from a list of possibly quoted values
+        # ensures that quoted values are escaped
+        # could name the below to be join_string_list?
+        def build_string_list(value)
+          unless value.respond_to?(:compact)
+            raise ArgumentError,
```
mullermp, line 13: I think you can remove validation here - it's a private class and will be controlled by codegen. This case will never arise.

hearth/lib/hearth/http/header_list_builder.rb
```diff
@@ -0,0 +1,33 @@
...
+                  "Expected #{value} to be an Array, got #{value.class} instead"
+          end
+
+          value.compact
+               .map { |s| escape_value(s) }
```
mullermp, line 18: nit - can be single lined

hearth/lib/hearth/http/header_list_builder.rb
```diff
@@ -0,0 +1,33 @@
...
+               .join(', ')
+        end
+
+        private
+
+        # Escapes header field value
+        # @return [String]
```
mullermp, line 25: No need to document private methods with yard if you don't want to. It won't show up in documentation. Plus this method is self documenting - `escape_value(str)` essentially describes what it does and what it takes. In general, we strive to write self documenting code, instead of writing comments to describe code 💯

codegen/projections/rails_json/lib/rails_json/builders.rb
```diff
@@ -515,15 +515,21 @@ def self.build(http_req, input:)
...
+          http_req.headers['X-StringList'] = Hearth::Http::HeaderListBuilder.build_string_list(input[:header_string_list])
+        end
+        unless input[:header_string_set].nil? || input[:header_string_set].empty?
+          http_req.headers['X-StringSet'] = Hearth::Http::HeaderListBuilder.build_string_list(input[:header_string_set])
+        end
+        http_req.headers['X-IntegerList'] = input[:header_integer_list].compact.join(', ') unless input[:header_integer_list].nil? || input[:header_integer_list].empty?
+        http_req.headers['X-BooleanList'] = input[:header_boolean_list].compact.join(', ') unless input[:header_boolean_list].nil? || input[:header_boolean_list].empty?
         unless input[:header_timestamp_list].nil? || input[:header_timestamp_list].empty?
-          http_req.headers['X-TimestampList'] = input[:header_timestamp_list].map { |t| Hearth::TimeHelper.to_http_date(t) }.join(', ')
+          http_req.headers['X-TimestampList'] = input[:header_timestamp_list].compact.map { |t| Hearth::TimeHelper.to_http_date(t) }.join(', ')
```
mullermp, line 527: I think the method for this in the header list builder would be "build_http_date_list"

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/RestBuilderGeneratorBase.java
```diff
@@ -433,16 +433,26 @@ protected HeaderListMemberSerializer(String inputGetter, String dataSetter, Memb
...
-            writer.write("$1L$2L unless $2L.nil? || $2L.empty?", dataSetter, inputGetter);
+            writer.write("$1L$2L.compact.join(', ') unless $2L.nil? || $2L.empty?",
```
mullermp, line 436: I think getDefault probably goes away now, if possible.

mullermp (commented):
Looks good so far!

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/RestBuilderGeneratorBase.java
jterapin: Yup, yes any changes we make here will update the rest-json builder

jterapin:
After some team discussion, we are moving toward moving the building/parsing logic for header lists to hearth.

hearth/spec/hearth/http/header_list_parser_spec.rb
```diff
@@ -38,21 +72,77 @@ module Http
...
+        end
+
+        it 'parses a single item' do
+          expect(HeaderListParser.parse_epoch_seconds_list('0.0'))
+            .to eq([Time.at(0.0)])
+        end
+
+        it 'parses multiple items' do
+          expect(
+            HeaderListParser.parse_epoch_seconds_list('0.0, 946845296')
```
alextwoods, line 139: The parser always converts with `.to_i`, I can't remember exactly what is specified for epoch seconds and whether it needs to be able to handle fractional seconds here or not.

alextwoods (approved):
Nice!

hearth/lib/hearth/http/header_list_builder.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+module Hearth
+  module Http
```
mullermp, line 4: I'm pretty sure everything in this module uses HTTP in all caps. I'm noticing both parser and builder do not do this! Please change the namespace of both of these.

hearth/spec/hearth/http/header_list_builder_spec.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+
+module Hearth
+  module Http
+    describe HeaderListBuilder do
+      describe '.builds_list' do
```
mullermp, line 6: Test names have wrong method names

hearth/spec/hearth/http/header_list_parser_spec.rb
```diff
@@ -3,6 +3,40 @@
...
+          expect(HeaderListParser.parse_integer_list('1'))
+            .to eq([1])
+        end
+
+        it 'parses multiple items' do
+          expect(HeaderListParser.parse_integer_list('1, 2, 3'))
+            .to eq([1, 2, 3])
+        end
+      end
+
```
mullermp, line 39: Ditto on test name. Please check all of these

hearth/spec/hearth/http/header_list_parser_spec.rb
mullermp: It probably doesn't hurt to do float here so we get maximum precision

mullermp (commented):
Ship it after you fix comments. Nice!

hearth/spec/hearth/http/header_list_builder_spec.rb
jterapin: Thank youuuuu for catching this! 😅

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/RestParserGeneratorBase.java
```diff
@@ -357,17 +349,13 @@ public Void stringShape(StringShape shape) {
...
+            writer.write("$1LHearth::Http::HeaderListParser.parse_boolean_list($2L)", dataSetter, valueGetter);
             return null;
         }
 
         @Override
         public Void integerShape(IntegerShape shape) {
-            splitByComma()
-                    .write(".map { |s| s.to_i }")
-                    .dedent();
+            writer.write("$1LHearth::Http::HeaderListParser.parse_integer_list($2L)", dataSetter, valueGetter);
```
alextwoods, line 358: So I think this was an issue with the previous code likely - but I *think* header lists can contain any number type (from the documentation the trait selector says  `list > member > :test(boolean, number, string, timestamp)))` - so I think the `number` means we need to cover all numeric types here.  We can cover that with two methods, a parse_integer_list and parse_float_list.  But I think we should probably add that.

hearth/lib/hearth/http/header_list_parser.rb
```diff
@@ -25,6 +33,14 @@ def parse_http_date_list(value)
           value.split(',').each_slice(2).map { |v| Time.parse(v[0] + v[1]) }
         end
 
+        def parse_date_time_list(value)
+          value.split(',').map { |v| Time.parse(v) }
+        end
+
+        def parse_epoch_seconds_list(value)
+          value.split(',').map { |v| Time.at(v.to_i) }
```
alextwoods, line 41: I think this should probably be a `to_f` instead of a `to_i` since these are defined as supporting millisecond precision everywhere I can find, and to_f should give the same values for any integer anyway so should be safe.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/RestParserGeneratorBase.java
jterapin: Updated to cover all numeric types!

hearth/spec/hearth/http/header_list_parser_spec.rb
jterapin: Leaving at `.to_i` per standup convo:
 * We retraced our steps to see how [TimestampFormat](https://github.com/smithy-lang/smithy-ruby/blob/8acb2a1d88719aab5f354170d180fcabbee6b494/codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/util/TimestampFormat.java) is handled everywhere else in the codegen. Note that `.to_i` is used for epoch seconds.
* Then, we looked at the smithy's protocol-tests to see if epoch seconds are covered in terms of fractional seconds: https://github.com/smithy-lang/smithy/blob/main/smithy-aws-protocol-tests/model/restJson1/fractional-seconds.smithy
* Note that only `DateTime` scenario are covered in the above test. In addition, the [documentation](https://smithy.io/2.0/spec/protocol-traits.html#timestamp-formats) is not explicitly clear on what the Deserializers should be doing for epoch-seconds (comparing to other time formats)

Based on the above findings, leaving this for now until either SmithyDocs updates with specifics or future Smithy protocol-tests starts covering this.

hearth/lib/hearth/http/header_list_parser.rb
jterapin: Leaving at `.to_i` per discussion outlined here: https://github.com/smithy-lang/smithy-ruby/pull/191#discussion_r1581408052

alextwoods (approved):
Nice - LGTM!

PR #192, opened 2024-04-25
Upgrade to IDL 2 and use enum shapes

mullermp:
Also fixes the 50+ warnings we have.

Upgrades to IDL 2


codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/EnumGenerator.java
```diff
@@ -15,26 +15,26 @@
...
 @SmithyInternalApi
 public final class EnumGenerator extends RubyGeneratorBase {
 
-    private final Shape shape;
+    private final Set<EnumShape> enumShapes;
 
     public EnumGenerator(GenerateEnumDirective<GenerationContext, RubySettings> directive) {
         super(directive);
-        this.shape = directive.shape();
+        this.enumShapes = directive.model().getEnumShapes();
```
alextwoods, line 37: This still seems wrong - This should be doing directive.shape() and should only process one enum from the directive I'm pretty sure

mullermp: AH that was it. Thanks

AndrewFossAWS (approved):

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/EnumGenerator.java
```diff
@@ -15,26 +15,24 @@
...
 @SmithyInternalApi
 public final class EnumGenerator extends RubyGeneratorBase {
 
-    private final Shape shape;
+    private final EnumShape shape;
 
     public EnumGenerator(GenerateEnumDirective<GenerationContext, RubySettings> directive) {
         super(directive);
-        this.shape = directive.shape();
+        this.shape = (EnumShape) directive.shape();
```
alextwoods, line 35: Use expectEnumShape instead here

alextwoods (approved):
Nice, looks good overall!

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/EnumGenerator.java
mullermp: I had that originally but I copied what int enum shape did. I can change both

PR #193, opened 2024-04-30
Chores + prepare release

mullermp:
Misc improvements


jterapin:
Looks good so far - prob want to update this Hearth [README](https://github.com/smithy-lang/smithy-ruby/blob/main/hearth/README.md) since it appears that the code example is out of date.

hearth/lib/hearth/middleware_stack.rb
```diff
@@ -15,7 +15,7 @@ def use(middleware, **middleware_kwargs)
     # @param context
     # @return [Output]
     def run(input, context)
-      stack.call(input, context)
+      stack&.call(input, context) || Output.new
```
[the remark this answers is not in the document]
mullermp: My IDE says it can be - but likely never unless something code generated is catastrophic. If you try to use this on it's own, yes it can be.

tasks/rubocop.rake
```diff
@@ -0,0 +1,15 @@
+namespace :rubocop do
+  desc 'Runs rubocop on hearth'
+  task 'hearth' do
+    Dir.chdir('hearth') do
+      sh('bundle exec rubocop -E -S')
```
alextwoods, line 5: If we want to keep bundle exec in here, I think we should probably do something like 
```
sh('BUNDLE_GEMFILE=../Gemfile bundle exec rubocop -E -S')
```

tasks/steep.rake
```diff
@@ -0,0 +1,22 @@
+namespace :steep do
+  desc 'Run steep check on hearth'
+  task 'hearth' do
+    Dir.chdir('hearth') do
+      sh('bundle exec steep check')
```
alextwoods, line 5: Ditto for all of these

alextwoods (approved):
LGTM!

PR #200, opened 2024-06-04
Add config to interceptors

mullermp:
This change adds operation `config` (unfrozen) to Hearth::Context and also InterceptorContext, so that interceptors may modify operation config.

### DIFF
--- a/hearth/sig/lib/hearth/interceptor_context.rbs
+++ b/hearth/sig/lib/hearth/interceptor_context.rbs
@@ -1,6 +1,6 @@
 module Hearth
   class InterceptorContext
-    def initialize: (input: Structure, request: Request, response: Response, output: Output[Structure], logger: Logger) -> void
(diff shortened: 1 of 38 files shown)


hearth/sig/lib/hearth/interceptor_context.rbs
```diff
@@ -1,6 +1,6 @@
 module Hearth
   class InterceptorContext
-    def initialize: (input: Structure, request: Request, response: Response, output: Output[Structure], logger: Logger) -> void
+    def initialize: (input: Structure, request: Request, response: Response, output: Output[Structure], config: Configuration[instance]) -> void
```
alextwoods, line 3: What is instance here?  Should this be based on the service and be a type parameter on InterceptorContext (And should Request, Response and Output use it?)

alextwoods (approved):

PR #201, opened 2024-06-07
RpcV2Cbor Implementation

alextwoods:
*Description of changes:*
Implements the Smithy [Rpcv2Cbor protocol](https://smithy.io/2.0/additional-specs/protocols/smithy-rpc-v2.htm).

See V3 implementation: https://github.com/aws/aws-sdk-ruby/pull/3006

All protocol tests pass.  Tested (and benchmarked) in V4 against SecretsManager.


Rakefile
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 $white_label_dir = 'codegen/smithy-ruby-codegen-test/build/smithyprojections/smithy-ruby-codegen-test/white-label/ruby-codegen/white_label'
+$rpcv2_cbor_dir = 'codegen/smithy-ruby-codegen-test/build/smithyprojections/smithy-ruby-codegen-test/protocoltests-rpcv2cbor/ruby-codegen/rpcv2_cbor'
```
mullermp, line 4: nit: Probably can just name this service rpcv2cbor similar to our rails json which is also protocol tests.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/Hearth.java
```diff
@@ -82,6 +82,11 @@ public final class Hearth {
             .name("JSON")
             .build();
 
+    public static final Symbol CBOR = Symbol.builder()
+            .namespace("Hearth", "::")
+            .name("Cbor")
```
mullermp, line 87: I think the class/namespace should be CBOR technically. We did the pascal case in v3 because that was convention there. But here we've done JSON, XML, and now CBOR.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/ProtocolGenerator.java
```diff
@@ -35,6 +34,38 @@
...
 
+    /**
+     * Returns an ordered list of protocol generators from all integrations.
+     * The order of this list is used when resolving a protocol for a service.
+     *
+     * @param integrations ordered list of integrations
+     * @return ordered protocol generators
+     */
+    static List<ProtocolGenerator> collectSupportedProtocolGenerators(
+            List<RubyIntegration> integrations
```
mullermp, line 45: nit: method arguments in both of these methods should be in-line.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/HttpProtocolTestGenerator.java
```diff
@@ -157,8 +157,18 @@ private void renderResponseTests(OperationShape operation,
                             renderStreamingParamReader(outputShape);
                         }
                     })
-                    .write("expect(output.data.to_h).to eq($L)",
-                            getRubyHashFromParams(outputShape, testCase.getParams()))
+                    .call(() -> {
+                        if (testCase.getBodyMediaType().isPresent()
```
mullermp, line 161: Should we wrap both of these in a helper?

hearth/lib/hearth/cbor.rb
```diff
@@ -0,0 +1,63 @@
...
+require_relative 'cbor/decoder'
+
+module Hearth
+  # Hearth::Cbor is a purpose-built set of utilities for working with
+  # Cbor. It does not support all features of generic Cbor
+  # parsing and serialization.
+  # @api private
+  module Cbor
+    # Generic Cbor error, super class for specific encode/decode related errors.
+    class CborError < StandardError; end
```
mullermp, line 15: Ditto on CBOR for nested classes.

hearth/lib/hearth/cbor.rb
```diff
@@ -0,0 +1,63 @@
...
+    # Malformed buffer, unexpected break code
+    class UnexpectedBreakCodeError < CborError; end
+
+    # malformed buffer, unexpected additional information
+    class UnexpectedAdditionalInformationError < CborError
+      def initialize(add_info)
+        super("Unexpected additional information: #{add_info}")
+      end
+    end
+    class << self
```
mullermp, line 49: Spacing

hearth/lib/hearth/cbor/tagged.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module Hearth
+  module Cbor
+    Tagged = Struct.new(:tag, :value)
```
mullermp, line 5: Do these need docs? (Is this public?)

hearth/lib/hearth/cbor/half.rb
```diff
@@ -0,0 +1,34 @@
...
+
+module Hearth
+  module Cbor
+    # 16 bit IEEE 754 half-precision floats
+    # Support decoding only
+    # format:
+    # sign - 1 bit
+    # exponent - 5 bits
+    # precision - 10 bits
+    module Half
```
mullermp, line 11: I really think this module is not necessary and can just be in the half method in the decoder.

hearth/lib/hearth/response.rb
```diff
@@ -17,6 +17,7 @@ def initialize(body: StringIO.new)
     # @param [Response] other
     # @return [Response]
     def replace(other)
+      other.body.rewind
```
mullermp, line 20: I don't think rewind should be done here for other. I think it should be the responsibility of the caller.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/protocol/rpcv2cbor/generators/ErrorsGenerator.java
```diff
@@ -0,0 +1,46 @@
...
+    public ErrorsGenerator(GenerationContext context) {
+        super(context);
+    }
+
+    @Override
+    public void renderErrorCodeBody() {
+        writer
+                .write("")
+                .openBlock("if !(200..299).cover?(resp.status)")
+                .write("data = $T.decode(resp.body.read.force_encoding(Encoding::BINARY))", Hearth.CBOR)
```
mullermp, line 34: I think all of the force_encoding(encoding::binary) should be moved to the decode method in hearth CBOR. That way it's not in all of the codegen parsers. I think it's reasonable for the decoder to try and coerce it into what it expects to decode if it hasn't been already.

mullermp (approved):
I'm generally ok with this. I think we can clean this up before merge.

Rakefile
alextwoods: I went with `rpcv2_cbor` to match `rails_json` (and the protocol shape is `Rpcv2Cbor` so `rpcv2_cbor` is I think the natural snake case of that).

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/Hearth.java
alextwoods: Ah yeah, I think that makes sense.  Will move it.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/HttpProtocolTestGenerator.java
alextwoods: yeah, good call.

hearth/lib/hearth/cbor.rb
alextwoods: Changed this to just `Error` since it will be `CBOR::Error`

hearth/lib/hearth/cbor/tagged.rb
alextwoods: I initially thought this should be considered private, but in theory the `CBOR.decode` can return it (if the data has a Tagged item that is not one of the types we support like Time) - I've removed this file and move the Tagged class and added documentation.

hearth/lib/hearth/cbor/half.rb
alextwoods: Part of my orig thought was to try and reduce the overall complexity of the decoder class and isolate logic a bit.  But I think the decoder is WAY past that point anyway.

hearth/lib/hearth/response.rb
alextwoods: Good catch, I didn't mean to check this change in.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/protocol/rpcv2cbor/generators/ErrorsGenerator.java
alextwoods: Yeah - I think that actually makes sense.  I was initially trying to avoid baking the cost of this encoding into the decode method but looking at the actual C implementation, if the encoding already matches, it exits immediately and doesn't copy memory - so I think its reasonable to push down to the decoder.

Rakefile
mullermp: I was talking about `protocoltests-rpcv2cbor`

alextwoods: oh, I see.  updated.

PR #207, opened 2024-06-29
Global config for Hearth clients

mullermp:
Config defaults need to remain inside the Config struct and Config must expose the original provided options. This is so global config plugin knows what was configured already (it may be nil).

Needs tests and some more thought on approach. Options are being modified by resolver, it would be preferable not to.


mullermp:
TODO: test for global being overwritten (hierarchy?) and value validation

codegen/smithy-ruby-codegen-test/integration-specs/client_spec.rb
```diff
@@ -91,6 +91,12 @@ module WhiteLabel
         client.kitchen_sink({}, interceptors: [interceptor])
       end
 
+      it 'allows for global configuration' do
+        logger = Logger.new(IO::NULL, level: :debug)
+        Hearth.config[:logger] = logger
+        expect(client.config.logger).to eq(logger)
+      end
```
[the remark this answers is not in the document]
jterapin: Disregard. we had an offline convo

hearth/lib/hearth.rb
```diff
@@ -53,6 +53,26 @@
...
+  # @api private
+  @config = {}
+
+  class << self
+    # @return [Hash] Returns a hash of default configuration options shared
+    #   by all constructed clients.
+    attr_reader :config
+
+    # @param [Hash] config
+    def config=(config)
```
alextwoods, line 70: While I do think global config is okay for AWS clients, I'm a little skeptical about supporting it for all generic Hearth clients - I think it could be a little bit of a foot gun if someone is using smithy-ruby clients for vastly different services.

hearth/lib/hearth/client.rb
```diff
@@ -38,7 +38,7 @@ def initialize_config(options, config_class)
       config.plugins.each { |p| p.call(config) }
       config.interceptors.concat(client_interceptors)
       config.validate!
-      config.freeze
+      config
```
alextwoods, line 41: Why not freeze it here?  Only the plugins should be able to modify this and once they've done that, it should be frozen I think?

alextwoods (commented):
I like the approach of adding the `options` to Config, but I'm not sure about supporting a global Hearth.config.

hearth/lib/hearth/client.rb
mullermp: I think maybe Config should be mutable once it's been assigned to the client, as an additional way to change config? We do this in v3. I'm on the fence about it. There are probably situations where we want to do this.

hearth/lib/hearth.rb
mullermp: Maybe. Let's chat offline.

hearth/lib/hearth/client.rb
alextwoods: Mutation of config on the client is not thread safe so should be avoided - I think we should not allow it.  Additionally, I think we want to make sure that interceptors are not able to modify it.

mullermp: It's true it's not thread safe. That doesn't necessarily mean we shouldn't do it. We will probably need to freeze config twice if that's the case?

hearth/lib/hearth.rb
alextwoods: If we do want to support generic global config - what about only supporting it in the generated SDK's namespace.  Eg:  `RailsJson.config` or `AWS::SDK::S3.config`?

hearth/lib/hearth/client.rb
alextwoods: By twice you mean in two places (client initialize and operation invocation) right?  That seems correct.

hearth/lib/hearth.rb
mullermp: I think we will also need per-service global config and AWS global config, but I think Hearth global config is still fine here. Yes there are risks, and I understand it may seem funky, but I think it's correct to have config at every level.

alextwoods (approved):

PR #208, opened 2024-07-11
Add Telemetry support

jterapin:
## Description
Observability is the extent to which a system's current state can be inferred from the data it emits. The data emitted is commonly referred to as telemetry. This PR will set up the groundwork for SDK instrumentation and current/future telemetry providers. 

In addition, we will support OpenTelemetry as a telemetry provider. 

**Related PR**: https://github.com/alextwoods/aws-sdk-ruby-v4/pull/54

## Changes
* Add `opentelemetry-sdk` gem for testing purposes
* Codegen
  * Update FakeProtocol's BuilderGenerator to render `http_method` (intended for span attribute testing)
  * Update `config_spec.rb` to include the new Config, `telemetry_provider`.
  * Add `telemetry_spec.rb` to test Config and Client behaviors (includes OTel-related testings)
  * Add  smithy file to test isolate and test telemetry components
  * Update ClientGenerator to include render telemetry code for each operation
  * Update MiddlewareBuilder to include new config, `telemetry_provider`
  * Update RubySettings to generate/access `SDK_ID` (if no SDK ID is present, we use Service Shape name)
* Hearth
  * Update Context to have access to a tracer
  * Update Send Middleware to emit telemetry (for both stub and send responses)
  * Add telemetry components, no-op classes and otel support (only tracing supported)
  * Add RBS signatures for telemetry components
  * Update `send_spec.rb` as adding telemetry support has changed how the tests are run
  * Add specs for telemetry components

~~* Add V4 support (separate PR in V4 repo)~~
~~* Create child span on a middleware level~~
~~* Update specs based on the child span + add more specs~~
~~* Add documentation~~

---


hearth/lib/hearth/middleware/send.rb
```diff
@@ -71,25 +71,69 @@ def call(input, context)
...
+      end
+
+      def response_attributes(span, context)
+        attributes =
+          {
+            'http.status_code' => context.response.status
+          }
+        if context.response.headers.key?('Content-Length')
+          attributes['http.response_content_length'] =
+            context.response.headers['Content-Length']
```
jterapin, line 128: Assumption here is that the response will contain this header that I could use to set this span attribute. I'm not sure if I should take this a step further to check the `body.size` (if possible) when there is none set.

codegen/smithy-ruby-codegen-test/model/component-test/telemetry.smithy
```diff
@@ -0,0 +1,14 @@
+$version: "2.0"
+namespace smithy.ruby.tests
+
+@auth([])
+@http(method: "POST", uri: "/telemetry_test")
+operation TelemetryTest {
```
jterapin, line 6: I _could_ potentially remove this and use the `kitchen_sink` instead but I had some auth related failures when I tried. To resolve this, I think I need to pass in some extra config but I figured maybe we want separate components for each (I find it easier to go thru component file vs. looking at kitchen sink).
```
Auth scheme smithy.api#httpApiKeyAuth did not have an identity resolver configured
...
```

hearth/lib/hearth/telemetry.rb
```diff
@@ -0,0 +1,47 @@
...
+  # a telemetry signal.
+  #
+  # A telemetry provider is used to emit telemetry data. By default, this
+  # functionality is disabled. The SDK currently supports OpenTelemetry (OTel)
+  # as a provider. To use the OTel provider, load the `opentelemetry-sdk` gem
+  # and then, pass in an instance of a `Hearth::Telemetry::OTelProvider`
+  # for telemetry provider.
+  #
+  # # Configure a telemetry provider
+  # Below example uses the OTel-based provider that we support:
```
alextwoods, line 24: May want to also put a comment about otel exporter?

hearth/lib/hearth/telemetry/context_manager.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Hearth
+  module Telemetry
+    # Base for all ContextManager classes
+    class ContextManager
+      def current; end
```
alextwoods, line 7: These should probably have some level of documentation since this is part of the public interface.

hearth/lib/hearth/telemetry/telemetry_provider.rb
```diff
@@ -0,0 +1,14 @@
...
+
+module Hearth
+  module Telemetry
+    # Base for all TelemetryProvider classes
+    class TelemetryProvider
+      def initialize(tracer_provider: nil, context_manager: nil)
+        @tracer_provider = tracer_provider
+        @context_manager = context_manager
+      end
+      attr_reader :tracer_provider, :context_manager
```
alextwoods, line 11: should probably do one attr_reader per attribute and document them

hearth/lib/hearth/telemetry/telemetry_provider.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module Hearth
+  module Telemetry
+    # Base for all TelemetryProvider classes
+    class TelemetryProvider
+      def initialize(tracer_provider: nil, context_manager: nil)
```
alextwoods, line 7: I think we may want to document this method since its effectively part of the public api.

alextwoods (commented):
Nice - looking good!

I'd love to see spans around other potentially expensive parts of the sdk request life: param->input, endpoint resolution, auth_resolution, identity_resolution, validation, serialization, deserialization, signing.

However, I think those can be a second PR.  Once we have finalized our interface, we can then go and add more complete/granular instrumentation.

hearth/spec/hearth/telemetry_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+        expect(Hearth::Telemetry.send(:otel_loaded?))
+          .to be_truthy
+      end
+
+      it 'returns false when opentelemetry-sdk is not available' do
+        expect(Hearth::Telemetry).to receive(:require)
+          .with('opentelemetry-sdk')
+          .and_raise(LoadError)
+        expect(Hearth::Telemetry.send(:otel_loaded?))
+          .to be_falsey
```
mullermp, line 23: For question mark methods, these should check true and false. not truthy and falsey

hearth/lib/hearth/middleware/send.rb
```diff
@@ -71,25 +71,69 @@ def call(input, context)
...
 
       def stub_response(input, context, output)
-        stub = @stubs.next(context.operation_name)
-        log_debug(context, "Stubbing response with stub: #{stub}")
-        apply_stub(stub, input, context, output)
-        log_debug(context, "Stubbed response: #{context.response.inspect}")
-        return unless context.response.body.respond_to?(:rewind)
-
-        context.response.body.rewind
+        context.tracer.in_span(
```
mullermp, line 74: Do we need these in all or most middleware?

hearth/lib/hearth/middleware/send.rb
```diff
@@ -71,25 +71,69 @@ def call(input, context)
...
+          'Middleware.Stub',
```
mullermp, line 75: Middleware.Stub isn't a middleware. This can also be send.

hearth/lib/hearth/middleware/send.rb
```diff
@@ -71,25 +71,69 @@ def call(input, context)
...
-        log_debug(context, "Sending request: #{context.request.inspect}")
-        @client.transmit(
-          request: context.request,
-          response: context.response,
-          logger: context.config.logger
-        )
-        log_debug(context, "Received response: #{context.response.inspect}")
-      rescue Hearth::NetworkingError => e
-        output.error = e
+        context.tracer.in_span(
```
mullermp, line 88: It makes sense to maybe move these blocks outside of send_request and stub_response for readability, or create a wrapper method.

hearth/lib/hearth/middleware/send.rb
```diff
@@ -71,25 +71,69 @@ def call(input, context)
...
+            logger: context.config.logger
+          )
+          response_attributes(span, context)
+          log_debug(context, "Received response: #{context.response.inspect}")
+        rescue Hearth::NetworkingError => e
+          output.error = e
+        end
+      end
+
+      def request_attributes(context)
```
mullermp, line 105: Should these live somewhere else? Maybe in some tracer marshaller and unmarshaller thing?

hearth/lib/hearth/telemetry/context_manager.rb
mullermp: I think we should actually just delete this class. We don't have base classes anymore that are no-ops like retry strategy or signers.

hearth/lib/hearth/telemetry/span_kind.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module Hearth
+  module Telemetry
+    # Type of span.
+    # Can be used to describe relationships between spans in a trace.
+    module SpanKind
```
mullermp, line 7: You can possibly get rid of this module and make these something like Hearth::Telemetry::INTERNAL_SPAN. This current style is like some sort of C enum design.

hearth/lib/hearth/telemetry/span_status.rb
```diff
@@ -0,0 +1,55 @@
...
+
+      def initialize(code, description: '')
+        @code = code
+        @description = description
+      end
+
+      attr_reader :code, :description
+
+      # The operation completed successfully.
+      OK = 0
```
mullermp, line 46: Are these constants ever checked anywhere? Should these be private (by not existing)?

hearth/lib/hearth/telemetry/span_status.rb
```diff
@@ -0,0 +1,55 @@
...
+          new(ERROR, description: description)
+        end
+      end
+
+      def initialize(code, description: '')
+        @code = code
+        @description = description
+      end
+
+      attr_reader :code, :description
```
mullermp, line 43: Needs documentation too

mullermp (commented):
Nice! I can do a more thorough review another time.

codegen/smithy-ruby-codegen-test/integration-specs/telemetry_spec.rb
```diff
@@ -0,0 +1,171 @@
...
+            telemetry_provider: otel_provider
+          )
+          client.telemetry_test
+          expect(finished_stub_span).not_to be_nil
+          expect(finished_stub_span.attributes)
+            .to eq(
+              'http.method' => 'POST',
+              'net.protocol.name' => 'http',
+              'net.protocol.version' => '1.1',
+              'net.peer.name' => 'localhost',
```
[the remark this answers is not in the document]
jterapin: Per our conversation, we are keeping this stub response span around (and renaming the span name to `Middleware.Send.Stub` or `Middleware.Send` but removing all the network-related attributes as they are not useful.

hearth/lib/hearth/telemetry/span_kind.rb
jterapin: [OpenTelemetry-Ruby implementation of SpanKind](https://github.com/open-telemetry/opentelemetry-ruby/blob/406cabe54f8fb4937d0f6b73fe830a51c04fdb92/api/lib/opentelemetry/trace/span_kind.rb#L4)

Since Observability is heavily based on OTel, I'll opt to leave these defined under SpanKind

hearth/lib/hearth/telemetry/span_status.rb
jterapin: Yes, these constants are intended for use. I use them [here](https://github.com/smithy-lang/smithy-ruby/blob/dc2072409e1d4e76e28aa44ba4eb3960bd063a4f/codegen/projections/rails_json/lib/rails_json/client.rb#L102) for each operation. 

Customers who intend to implement their own telemetry provider could utilize these as well.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/ClientGenerator.java
```diff
@@ -197,6 +198,8 @@ private void renderOperation(RubyCodeWriter writer, OperationShape operation) {
                     }
                 })
                 .write("config = operation_config(options)")
+                .write("tracer = config.telemetry_provider.tracer_provider.tracer('$L.$L')",
+                        settings.getModule().toLowerCase(), getModule().toLowerCase())
```
jterapin, line 202: Note to self: Add a transformer

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/ClientGenerator.java
```diff
@@ -210,7 +213,17 @@ private void renderOperation(RubyCodeWriter writer, OperationShape operation) {
                                 .render(context))
                 .write("config: config,")
                 .write("operation_name: :$L,", operationName)
+                .write("tracer: tracer")
                 .closeBlock(")")
+                .openBlock("attributes = {")
+                .write("'rpc.service' => '$L',", serviceName)
```
jterapin, line 219: Note to self: Cover AWS-specific cases

codegen/smithy-ruby-codegen-test/integration-specs/telemetry_spec.rb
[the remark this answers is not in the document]
jterapin: I think I'll keep to `Middleware.Send` here.

alextwoods (approved):
Nice, looks good!

hearth/lib/hearth/telemetry/span_status.rb
mullermp: That's different. I'm talking about for OK, UNSET, and ERROR. In your linked code I see `Hearth::Telemetry::SpanKind::CLIENT`

hearth/sig/lib/hearth/telemetry.rbs
```diff
@@ -0,0 +1,54 @@
...
+
+    class OTelProvider < TelemetryProvider
+      def initialize: () -> void
+    end
+
+    class OTelContextManager
+      include _ContextManager
+    end
+
+    interface _TracerProvider
```
mullermp, line 19: I think these can all be base classes later if we do that approach.

hearth/sig/lib/hearth/telemetry.rbs
```diff
@@ -0,0 +1,54 @@
+module Hearth
+  module Telemetry
+    class TelemetryProvider
```
mullermp, line 3: This one doesn't belong here (all the rbs files should match rb directory structure).

hearth/spec/hearth/telemetry/span_status_spec.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module Hearth
+  module Telemetry
+    describe SpanStatus do
+      describe '.unset' do
+        it 'returns the correct expected code' do
+          status = Hearth::Telemetry::SpanStatus.unset
```
mullermp, line 8: This is what my previous comment is referring to about constants. The API expects these methods, so constants shouldn't exist I think.

hearth/spec/hearth/telemetry/otel_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+              expect(finished_span.name).to eq('some_span')
+              expect(finished_span.attributes).to eq('apple' => 'pie')
+              expect(finished_span.events[0].name).to eq('pizza party')
+              expect(finished_span.status)
+                .to be_an_instance_of(Hearth::Telemetry::SpanStatus)
+            end
+          end
+
+          describe '#in_span' do
+            let(:error) do
```
mullermp, line 100: You can just use StandardError.new to get an instance of it.

hearth/spec/hearth/telemetry/otel_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+          expect { otel_provider }.to raise_error(ArgumentError)
+        end
+
+        it 'sets up tracer provider and context manager' do
+          expect(tracer_provider).to be_a(OpenTelemetry::Trace::TracerProvider)
+          expect(context_manager).to be_a(Hearth::Telemetry::OTelContextManager)
+        end
+      end
+
+      describe ':context_manager' do
```
mullermp, line 31: Should this be ContextManager (the class name) instead of a symbolized method name?

hearth/spec/hearth/telemetry/otel_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+          it 'detaches the previously set context' do
+            token = context_manager.attach(new_context)
+            expect(context_manager.current).to eq(new_context)
+            context_manager.detach(token)
+            expect(context_manager.current).to eq(root_context)
+          end
+        end
+      end
+
+      describe ':tracer_provider' do
```
mullermp, line 68: Ditto

mullermp (commented):
Looks good so far

hearth/lib/hearth/telemetry/span_status.rb
jterapin: Oops. I was looking at the wrong tracer component. 

Hmm. I was thinking of the following usage:
```
class CustomSpan
  def status=(status)
    return if status.code == Hearth::Telemetry::SpanStatus::UNSET
    ...
  end
end

```
I'm guessing OpenTelemetry-Ruby did this for readability instead of figuring what code does what.

hearth/spec/hearth/telemetry/span_status_spec.rb
jterapin: I think the actual expectation should be more like, `expect(status.code).to eq(Hearth::Telemetry::SpanStatus::UNSET)` when re-looking at this

mullermp: Will customers use Hearth::Telemetry::SpanStatus::UNSET or SpanStatus.unset? If the latter, then former constant should be removed and the test was fine as is.

jterapin: Potentially, both could be used. Like below:
Customer could set the span status code like so:
```ruby
span = tracer.start_span('foo')
span.status = Hearth::Telemetry::SpanStatus.ok(description: 'hello!')
```

Customer may want to verify its status by using constant.
```ruby
# maybe some work relating to collecting certain spans based on status
error_spans = []
spans.each do | span |
  if span.status.code == Hearth::Telemetry::SpanStatus::ERROR
  exceptions << span
end
```

hearth/sig/lib/hearth/telemetry.rbs
jterapin: I was trying to consolidate but with the introduction of base classes, this is no longer needed (`TelemetryProviderBase` is replaces this)

codegen/projections/rails_json/lib/rails_json/client.rb
```diff
@@ -83,22 +83,32 @@ def initialize(options = {})
...
-        raise output.error
+        tracer: tracer
+      )
+      attributes = {
+        'rpc.service' => 'RailsJson',
+        'rpc.method' => 'AllQueryStringTypes',
+        'code.function' => 'all_query_string_types',
+        'code.namespace' => 'RailsJson::Client'
+      }
+      tracer.in_span('RailsJson.AllQueryStringTypes', attributes: attributes, kind: Hearth::Telemetry::SpanKind::CLIENT) do
```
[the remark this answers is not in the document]
jterapin: Great idea, I made the change and everything is much cleaner. Yeah the logger option needs some tidying as well so I will opt to do that on a separate PR after this one wraps up.

hearth/lib/hearth/telemetry.rb
```diff
@@ -0,0 +1,67 @@
...
+module Hearth
+  # # Telemetry
+  # Observability is the extent to which a system's current state can be
+  # inferred from the data it emits. The data emitted is commonly referred
+  # as telemetry. The AWS SDK for Ruby currently supports traces as
+  # a telemetry signal.
+  #
+  # A telemetry provider is used to emit telemetry data. By default, the
+  # `NoOpTelemetryProvider` will not record or emit any telemetry data.
+  # The SDK currently supports OpenTelemetry (OTel) as a provider. To use
```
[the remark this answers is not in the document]
jterapin: Makes sense, I moved most of the OTel docs but did leave a trace of mention so customer can navigate to it

hearth/lib/hearth/telemetry/no_op.rb
```diff
@@ -0,0 +1,152 @@
...
+    # No-op implementation for Tracer
+    class NoOpTracer
+      # Used when a caller wants to manage the activation/deactivation and
+      # lifecycle of the Span and its parent manually.
+      #
+      # @param [String] name Span name
+      # @param [optional Context] with_parent Parent Context
+      # @param [optional Hash] attributes Attributes to attach to the span
+      # @param [optional Hearth::Telemetry::SpanKind] kind Type of Span
+      # @return [Span]
```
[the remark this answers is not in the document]
jterapin: All moved to base classes

hearth/lib/hearth/telemetry/no_op.rb
```diff
@@ -0,0 +1,152 @@
+# frozen_string_literal: true
+
+module Hearth
+  module Telemetry
+    # No-op implementation for TelemetryProvider
+    class NoOpTelemetryProvider < TelemetryProvider
```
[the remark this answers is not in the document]
jterapin: Doing an approach with base classes. it definitely cleans things up a bit better but lmk

codegen/projections/rails_json/lib/rails_json/client.rb
```diff
@@ -93,13 +93,7 @@ def all_query_string_types(params = {}, options = {})
...
         tracer: tracer
       )
-      attributes = {
-        'rpc.service' => 'RailsJson',
-        'rpc.method' => 'AllQueryStringTypes',
-        'code.function' => 'all_query_string_types',
-        'code.namespace' => 'RailsJson::Client'
-      }
-      tracer.in_span('RailsJson.AllQueryStringTypes', attributes: attributes, kind: Hearth::Telemetry::SpanKind::CLIENT) do
+      Telemetry::AllQueryStringTypes.span(context) do
```
alextwoods, line 96: Nice, I like this!

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/TelemetryGenerator.java
```diff
@@ -0,0 +1,91 @@
...
+
+    private void renderOperation(RubyCodeWriter writer, OperationShape operation) {
+        String classOperationName = symbolProvider.toSymbol(operation).getName();
+        String operationName = RubyFormatter.toSnakeCase(classOperationName);
+        String sdkId = settings.getSdkId();
+
+        writer
+                .write("")
+                .openBlock("class $L", classOperationName)
+                .openBlock("def self.span(context, &block)")
```
alextwoods, line 76: nit: I think `in_span` might read better in the client code.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/TelemetryGenerator.java
```diff
@@ -0,0 +1,91 @@
...
+                .openBlock("attributes = {")
+                .write("'rpc.service' => '$L',", sdkId)
+                .write("'rpc.method' => '$L',", classOperationName)
+                .write("'code.function' => '$L',", operationName)
+                .write("'code.namespace' => '$L'", nameSpace())
+                .closeBlock("}")
+                .openBlock("context.tracer.in_span('$L.$L', attributes: attributes, "
+                        + "kind: Hearth::Telemetry::SpanKind::CLIENT) do",
```
alextwoods, line 84: Nit: I'd recommend introducing a constant in the Hearth java class for SpanKind.

hearth/lib/hearth/telemetry.rb
```diff
@@ -1,56 +1,32 @@
...
-  #     otel_provider = Hearth::Telemetry::OTelProvider.new
-  #     client = Service::Client.new(telemetry_provider: otel_provider)
-  #
-  # If a provider isn't supported, you can implement your own support by
-  # following the telemetry interfaces defined in RBS.
+  # The SDK currently supports OpenTelemetry (OTel) as a provider. See
+  # {OTelProvider} for more information.
+  #
+  # If a provider isn't supported, you can implement your own provider by
+  # following by inheriting the following base classes and implementing
```
alextwoods, line 21: Nit on wording.  "by following by inheriting" -> "by inheriting".  

Also "implementing interface defined" -> "implementing the interfaces defined"

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/TelemetryGenerator.java
```diff
@@ -0,0 +1,91 @@
...
+                    .call(() -> renderOperations(writer))
+                    .write("")
+                    .closeAllModules();
+        });
+        LOGGER.fine("Wrote telemetry to " + rbFile());
+    }
+
+    private void renderOperations(RubyCodeWriter writer) {
+        operations.stream()
+                .filter((o) -> !Streaming.isEventStreaming(model, o))
```
alextwoods, line 63: Totally out of scope for this PR, but we'll need to figure out how telemetry should work for event stream operations....

codegen/projections/rails_json/lib/rails_json/telemetry.rb
```diff
@@ -0,0 +1,925 @@
...
+    class AllQueryStringTypes
+      def self.in_span(context, &block)
+        attributes = {
+          'rpc.service' => 'RailsJson',
+          'rpc.method' => 'AllQueryStringTypes',
+          'code.function' => 'all_query_string_types',
+          'code.namespace' => 'RailsJson::Telemetry'
+        }
+        context.tracer.in_span('RailsJson.AllQueryStringTypes', attributes: attributes, kind: Hearth::Telemetry::SpanKind::CLIENT) do
+          block.call
```
mullermp, line 23: I don't think these need to call a block, the method could just return ` context.tracer.in_span('RailsJson.AllQueryStringTypes', attributes: attributes, kind: Hearth::Telemetry::SpanKind::CLIENT)` couldn't it? I would also format in_span multiline if it's easier to do so.

codegen/projections/white_label/lib/white_label/builders.rb
```diff
@@ -13,11 +13,13 @@ module Builders
 
     class CustomAuth
```
mullermp, line 14: Will you need to merge from main? WhiteLabel should have protocol stuff now.

codegen/smithy-ruby-codegen-test-utils/src/main/java/software/amazon/smithy/ruby/codegen/protocol/generators/BuilderGenerator.java
```diff
@@ -42,6 +43,10 @@ public BuilderGenerator(GenerationContext context) {
     @Override
     protected void renderOperationBuildMethod(OperationShape operation, Shape inputShape) {
         writer.openBlock("def self.build(http_req, input:)");
+        if (operation.hasTrait(HttpTrait.class)) {
```
mullermp, line 46: This probably goes away with a merge from main.

hearth/lib/hearth/telemetry/base.rb
```diff
@@ -0,0 +1,169 @@
+# frozen_string_literal: true
+
+module Hearth
+  module Telemetry
+    # Base for TelemetryProvider classes.
```
mullermp, line 5: I'd add a few more lines to these descriptions. What does the provider do? What does the tracer provider do?

hearth/sig/lib/hearth/telemetry/otel.rbs
```diff
@@ -0,0 +1,10 @@
+module Hearth
+  module Telemetry
+    class OTelProvider < TelemetryProviderBase
```
mullermp, line 3: Should other bases be wrapped here too in case they may change? (i.e. use the whole api as intended)

hearth/sig/lib/hearth/telemetry/base.rbs
```diff
@@ -0,0 +1,46 @@
...
+    end
+
+    class TracerProviderBase
+      def tracer: (?String name) -> TracerBase
+    end
+
+    class TracerBase
+      def start_span: (String name, ?untyped with_parent, ?Hash[String, untyped] attributes, ?untyped kind) -> SpanBase
+
+      def in_span: (String name, ?Hash[String, untyped] attributes, ?untyped kind) -> SpanBase
```
mullermp, line 17: Are all of these types filled out? I would think there's a type to kind, status, exception, etc

mullermp (approved):
Nice! I generally approve, just needs a few clean ups.

hearth/sig/lib/hearth/telemetry/otel.rbs
alextwoods: I still don't see a point in wrapping those inner classes.  We control *this* interface here.  If any of the Otel implementations change under the hood, then we can wrap them through this class without breaking anything.

jterapin: Per team discussion last week, we are doing the wrapping approach.

hearth/lib/hearth/telemetry/otel.rb
```diff
@@ -0,0 +1,162 @@
...
+    end
+
+    # OpenTelemetry-based Span, represents a single operation
+    # within a trace.
+    class OTelSpan < SpanBase
+      def initialize(span)
+        @span = span
+      end
+
+      # @api private
```
alextwoods, line 95: Why is this api private (and why do we have have it?)

hearth/lib/hearth/telemetry/otel.rb
```diff
@@ -0,0 +1,162 @@
...
+
+      def finish(end_timestamp: nil)
+        @span.finish(end_timestamp: end_timestamp)
+      end
+
+      def record_exception(exception, attributes: nil)
+        @span.record_exception(exception, attributes: attributes)
+      end
+    end
+    # rubocop:enable Lint/MissingSuper
```
alextwoods, line 123: Should we just be calling super? I know it doesn't do anything on the base class, but I don't see a reason we cant call it

hearth/lib/hearth/telemetry/otel.rb
```diff
@@ -0,0 +1,162 @@
...
+    #     otel_provider = Hearth::Telemetry::OTelProvider.new
+    #     client = Service::Client.new(telemetry_provider: otel_provider)
+    class OTelProvider < TelemetryProviderBase
+      def initialize
+        unless Hearth::Telemetry.otel_loaded?
+          raise ArgumentError,
+                'Requires the `opentelemetry-sdk` gem to use OTel Provider.'
+        end
+        super(
+          # tracer_provider: OpenTelemetry.tracer_provider,
```
alextwoods, line 43: Nit: Probably remove this now

hearth/lib/hearth/telemetry/otel.rb
```diff
@@ -0,0 +1,162 @@
...
+          with_parent: with_parent,
+          attributes: attributes,
+          kind: kind
+        )
+        OTelSpan.new(span)
+      end
+
+      def in_span(name, attributes: nil, kind: nil, &block)
+        @tracer.in_span(name, attributes: attributes, kind: kind) do |span|
+          wrapped_span = OTelSpan.new(span)
```
alextwoods, line 82: Could probably one line this.

hearth/lib/hearth/telemetry/otel.rb
jterapin: Per our conversation, we are doing the `instance_variable_get` in tests in lieu of having a reader

hearth/lib/hearth/telemetry/base.rb
```diff
@@ -36,7 +43,7 @@ class TracerBase
       # lifecycle of the Span and its parent manually.
       #
       # @param [String] name Span name
-      # @param [optional Context] with_parent Parent Context
+      # @param [optional object] with_parent Parent Context
```
mullermp, line 46: You should remove "optional" from these - yardoc doesn't care about this, only about the types. Why is Object correct? Is it because it can be arbitrary? This should also be Object not object if so.

hearth/sig/lib/hearth/telemetry/otel.rbs
```diff
@@ -4,6 +4,18 @@ module Hearth
       def initialize: () -> void
     end
 
+    class OTelTracerProvider < TracerProviderBase
+      def initialize: () -> void
+    end
+
+    class OTelTracer < TracerBase
+      def initialize: (untyped tracer) -> void
```
mullermp, line 12: I think you can add types to these. Tracer and Span

mullermp (approved):

hearth/sig/lib/hearth/telemetry/otel.rbs
jterapin: Not needed per offline convo

PR #209, opened 2024-07-13
Remove Struct from Types

mullermp:
Remove Struct from Types and use PORO instead. RBS caught issues now that struct isn't being used, including fixing documents and string maps which had symbolized keys


mullermp:
TODO: changelog, update reserved words

hearth/sig/lib/hearth/interfaces.rbs
```diff
@@ -84,4 +84,10 @@ module Hearth
   interface _WritableIO
     def write: (*_ToS) -> Integer
   end
+
+  interface _ReadableIO
+    def read: (*_ToS) -> String
+    def readpartial: (Integer, ?String) -> String
+    def size: () -> Integer
```
alextwoods, line 91: I think this might not always have size?

codegen/projections/rails_json/lib/rails_json/types.rb
```diff
@@ -76,52 +76,59 @@ module Types
...
-      :query_enum_list,
-      :query_integer_enum,
-      :query_integer_enum_list,
-      :query_params_map_of_string_list,
-      keyword_init: true
-    ) do
+    class AllQueryStringTypesInput
       include Hearth::Structure
+
+      MEMBERS = %i[
```
alextwoods, line 82: What about having Hearth::Structure define a class method for `members` which would then setup the attr accessors (and set the MEMBERS).   Eg:

```
class MyType
  include Hearth::Structure

  members %i(member1 member2)
end

File: hearth/lib/hearth/structure.rb [hunk 3]
```diff
@@ -6,28 +6,34 @@ module Structure
...
+      if is_a?(Union)
+        super
+      else
+        if respond_to?(:_defaults, true)
+          _defaults.each do |k, v|
+            options[k] = v unless options.include?(k)
+          end
+        end
+        options.each do |k, v|
+          send("#{k}=", v)
```
alextwoods, line 18: Is there a reason to not add a `:[]` and `:[]=` method to continue supporting the way we can access values in V3?

codegen/projections/white_label/lib/white_label/params.rb
```diff
@@ -13,55 +13,55 @@ module Params
 
     class ClientError
       def self.build(params, context:)
-        Hearth::Validator.validate_types!(params, ::Hash, Types::ClientError, context: context)
+        Hearth::Validator.validate_types!(params, ::Hash, context: context)
```
alextwoods, line 16: Why are we removing the ability to pass types?  I think thats a great feature, esp when using parts of an output.

mullermp: Because of input output shapes having different types, this does not work as expected. You still have to manipulate new input. I think hash only simplifies paths too.

hearth/sig/lib/hearth/interfaces.rbs
mullermp: Is there a way to put this as optional?

codegen/projections/rails_json/lib/rails_json/types.rb
mullermp: I was avoiding adding new methods because then those have to be reserved. Class method may be ok  too.

hearth/lib/hearth/structure.rb
mullermp: Because union inherits struct and union shouldn't have those methods. I tried to keep it less hash like and more class like. I don't think we use the hash like interface anyway in v3. It should be a distinction between param and type.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/types/StructureGenerator.java
```diff
@@ -90,13 +96,26 @@ public void render() {
...
                     .call(() -> {
                         shape.members().forEach(memberShape -> {
-                            String name = symbolProvider.toMemberName(memberShape);
-                            Symbol target = symbolProvider.toSymbol(model.expectShape(memberShape.getTarget()));
-                            writer.write("attr_accessor $L (): $L", name, target.getProperty("rbsType").get());
+                            String memberName = symbolProvider.toMemberName(memberShape);
+                            Shape target = model.expectShape(memberShape.getTarget());
+                            Symbol symbol = symbolProvider.toSymbol(target);
+                            String rbsType;
+                            // TODO: not sure if any of this belongs here
```
alextwoods, line 107: I think we could move this logic to the RubySymbolProvider.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/types/StructureGenerator.java
```diff
@@ -90,13 +96,26 @@ public void render() {
...
+                            if (target.hasTrait(StreamingTrait.class)) {
+                                // TODO: determine if input or output and set accordingly
```
alextwoods, line 109: This is tricky to do I think and a shape can in theory be used by both input and output, but I think one of the knowledge indexes might let us do it?

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/types/StructureGenerator.java
```diff
@@ -90,13 +96,26 @@ public void render() {
...
+                                // String is a hack because of protocol tests comparing blob? FIXME
```
alextwoods, line 110: Where do we compare the blob in the protocol tests?  

I think on [output](https://github.com/smithy-lang/smithy-ruby/blob/main/codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/HttpProtocolTestGenerator.java#L157) we handle streaming correctly by reading.

But I guess on input when we create our [param inputs](https://github.com/smithy-lang/smithy-ruby/blob/main/codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/util/ParamsToHash.java#L83) we don't check for streaming... maybe we need to there?

codegen/projections/white_label/lib/white_label/params.rb
mullermp: NVM. I think it does make sense for unions and sub structures..

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/types/StructureGenerator.java
mullermp: Some of it does belong here.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/types/StructureGenerator.java
mullermp: I can't find a better way around this.

alextwoods (approved):
This will require downstream (V4) fixes as well right since its breaking changes?

I'm not quite sure why the downstream tests are passing - something may not be working correctly with them....

PR #210, opened 2024-07-22
Remove Struct from everything

mullermp:
Removes Struct from:
* Auth
* Endpoints
* Various Hearth internal structures
* Config

This is because RBS with Struct is not well supported.

### DIFF
--- a/hearth/lib/hearth/structure.rb
+++ b/hearth/lib/hearth/structure.rb
@@ -23,7 +23,7 @@ module Hearth
     # Deeply converts the Struct into a hash. Structure members that
     # are `nil` are omitted from the resultant hash.
     #
-    # @return [Hash, Structure, Object]
+    # @return [Hash, Structure]
     def to_h(obj = self)
       case obj
       when Union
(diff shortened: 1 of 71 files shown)


hearth/lib/hearth/structure.rb
```diff
@@ -23,7 +23,7 @@ def initialize(options = {})
     # Deeply converts the Struct into a hash. Structure members that
     # are `nil` are omitted from the resultant hash.
     #
-    # @return [Hash, Structure, Object]
+    # @return [Hash, Structure]
     def to_h(obj = self)
```
alextwoods, line 27: Should we also implement a to_s method?

alextwoods (approved):
Nice!  

Minor comments.  Also - add changelogs?

hearth/lib/hearth/structure.rb
mullermp: Yeah, but we need a SENSITIVE members array too I think

alextwoods: We currently generate a to_s method only when sensitive members exist.  

We can either:
1. Write a default to_s method on Hearth::Structure and make no changes to codegen (continue generating a custom to_s that overrides when senstive members exist).
2. Generate a to_s method for all generated Types.
3. Generate a SENSITIVE members list and then write a default to_s on Hearth::Strucutre that handles that.

My preference would be 1.  But I'm fine as long as we have both useful default printing of the types + consideration of sensitive members.

PR #212, opened 2024-07-27
Refactor benchmarking + fix related warnings and issues

mullermp:
Refactors benchmarking to use the shared tooling


codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/HttpProtocolTestGenerator.java
```diff
@@ -508,7 +508,8 @@ RubyImportContainer.CGI, getRubyArrayFromList(queryParams))
                     .write("actual_query = $T.parse(request.uri.query)",
                             RubyImportContainer.CGI)
                     .openBlock("expected_query.each do |k, v|")
-                    .write("expect(actual_query[k]).to eq(v)")
+                    .write("actual = actual_query[k].map { |s| s.force_encoding('utf-8') }")
```
mullermp, line 511: Not 100% on this one, but looks like query params are only ASCII but protocol tests are UTF-8, so I think no change to our actual builders but instead just compare what we would have sent as utf-8. Doing this in reverse has a frozen string error.

mullermp:
We did not have jruby on CI .. had to do some misc work to fix that build.

.github/workflows/downstream_ci.yml
```diff
@@ -32,7 +29,14 @@ jobs:
           ruby-version: ${{ env.ruby_version }}
           bundler-cache: true
 
-      - name: publish smithy-ruby to maven local
+      - uses: actions/checkout@v4
+
+      - name: Install gems
```
alextwoods, line 34: I don't think we need to do this here, should be handled by using checkout and then setup ruby after.

mullermp: It was needed to get CI to pass. Try removing it and see.

hearth/lib/hearth/http/fields.rb
```diff
@@ -107,9 +108,8 @@ def each(&block)
         end
         alias each_pair each
 
-        # @api private
         def inspect
-          to_h
+          to_h.to_s
```
alextwoods, line 112: should this be to_h.inspect instead?

mullermp: Maybe but it should be the same value I think. Will change.

.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "tasks/benchmark"]
+	path = tasks/benchmark
+	url = git@github.com:aws/aws-sdk-ruby-benchmark-tools.git
+	branch = initial-benchmarking
```
alextwoods, line 4: Should switch to main?

alextwoods (approved):

PR #213, opened 2024-07-29
Test protocol

alextwoods:
Refactor the FakeProtocol to TestProtocol and implement a very basic json RPC (based loosely on AWS Json).

### DIFF
--- a/codegen/projections/white_label/lib/white_label/builders.rb
+++ b/codegen/projections/white_label/lib/white_label/builders.rb
@@ -7,118 +7,297 @@
 #
 # WARNING ABOUT GENERATED CODE
 
+require 'stringio'
+
 module WhiteLabel
   # @api private
   module Builders
 
     class CustomAuth
       def self.build(http_req, input:)
+        http_req.http_method = 'POST'
+        http_req.append_path('/')
+        http_req.headers['Content-Type'] = 'application/json'
+        http_req.headers['X-Rpc-Target'] = 'WhiteLabel.CustomAuth'
+        data = {}
+        http_req.body = ::StringIO.new(Hearth::JSON.dump(data))
       end
     end
 
     class DataplaneEndpoint
       def self.build(http_req, input:)
+        http_req.http_method = 'POST'
+        http_req.append_path('/')
+        http_req.headers['Content-Type'] = 'application/json'
+        http_req.headers['X-Rpc-Target'] = 'WhiteLabel.DataplaneEndpoint'
+        data = {}
+        http_req.body = ::StringIO.new(Hearth::JSON.dump(data))
       end
     end
 
     class Defaults
+      def self.build(input)
+        data = {}
+        data['String'] = input.string unless input.string.nil?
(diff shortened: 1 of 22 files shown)


codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/BuilderGeneratorBase.java
```diff
@@ -288,7 +288,6 @@ protected void renderBuildersForOperation(OperationShape operation, Shape inputS
     /**
      * @param inputShape inputShape from a streaming operation to render for.
      */
-    // TODO: should probably be in Rest Base
```
mullermp, line 291: Should it?

codegen/projections/white_label/lib/white_label/builders.rb
```diff
@@ -7,118 +7,297 @@
...
 module WhiteLabel
   # @api private
   module Builders
 
     class CustomAuth
       def self.build(http_req, input:)
+        http_req.http_method = 'POST'
+        http_req.append_path('/')
+        http_req.headers['Content-Type'] = 'application/json'
+        http_req.headers['X-Rpc-Target'] = 'WhiteLabel.CustomAuth'
```
mullermp, line 21: Is this header necessary?

mullermp (approved):
I'm assuming this is just json 1.1 pretty much. You could go a step further and slim it down some, it doesn't need to be aware of every trait since those are tested with protocol tests.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/BuilderGeneratorBase.java
alextwoods: I believe it should be in the base - its used by all protocols at this point and is unrelated to REST.

codegen/projections/white_label/lib/white_label/builders.rb
alextwoods: For this protocol to actually work we need to send the method in some way - it could be through the header or on the url.  I opted for the header as it seemed easier.

alextwoods:
> I'm assuming this is just json 1.1 pretty much. You could go a step further and slim it down some, it doesn't need to be aware of every trait since those are tested with protocol tests.

Yeah - its pretty close to json 1.1 but actually needs to support MORE traits (httpPayload) for us to test all of the features we have in our integration tests.  This a minimal implementation of traits that are required for our existing white label integration tests to pass.  It does support more shapes than is technically necessary, but I think its reasonable to just support all of them as it makes writing other tests down the road easier (ie, you don't need to think about what shapes are supported or not when writing a test).

codegen/projections/white_label/lib/white_label/builders.rb
mullermp: I thought rpc uses url

PR #215, opened 2024-08-13
Event Stream Stubbing

alextwoods:
Implement Event Stream Stubbing.

Adds support for stubbing for event streams.  Interface:
```
client.stub_responses(
    :event_stream_operation,
    {
        events: [
            Types::ChatEvent.new(message: 'message'),
            {message: Message.new(headers: {}, payload: 'somedata')}
        ], 
        initial_response: {initial_structure: {message: 'data'}}}
    },
    {
        error: Hearth::NetworkingError # an error in initial request, no events are sent
    }
)
```


codegen/projections/rails_json/lib/rails_json/middleware.rb
```diff
@@ -53,8 +53,10 @@ def self.build(config, options = {})
         stack.use(Middleware::RequestId)
         stack.use(Hearth::Middleware::Send,
           client: config.http_client,
+          response_events: false,
           stub_data_class: Stubs::AllQueryStringTypes,
           stub_error_classes: [],
+          stub_message_encoder: Hearth::EventStream::Binary.const_get(:MessageEncoder).new,
```
mullermp, line 59: Why do you need const_get here?

codegen/projections/rails_json/lib/rails_json/middleware.rb
```diff
@@ -53,8 +53,10 @@ def self.build(config, options = {})
         stack.use(Middleware::RequestId)
         stack.use(Hearth::Middleware::Send,
           client: config.http_client,
+          response_events: false,
```
mullermp, line 56: Is there a true example here?

codegen/projections/white_label/spec/event_stream_stubs_spec.rb
```diff
@@ -0,0 +1,305 @@
...
+          Hearth::EventStream::HeaderValue.new(value: v, type: 'string')
+      end
+      Hearth::EventStream::Message.new(
+        headers: headers,
+        payload: StringIO.new('{"message":"event_message"}')
+      )
+    end
+
+    it 'signals the stubbed event' do
+      subject.stub_responses(:start_event_stream, {
```
mullermp, line 34: Wonky formatting on all of these.

hearth/lib/hearth/middleware/event_streams.rb
```diff
@@ -90,6 +92,7 @@ def setup_response_events(context)
           event_handler: @event_handler
         )
         context.response.body = decoder
+        context.metadata[:event_handler] = @event_handler
```
mullermp, line 95: Should there be a named key for this? Or is metadata fine?

mullermp (approved):
Nice! I probably missed a bunch, but I trust the test coverage.

codegen/projections/rails_json/lib/rails_json/middleware.rb
alextwoods: This is how the module is intended to work from codegen - similar to the way we do an error module.

The specific encoding of messages is protocol specific (all existing ones currently use this same logic, but its supposed to be configurable in codegen per protocol) - so the protocol itself defines just a module to use and that module is expected to porovide a `MessageEncoder` and a `MessageDecoder`.

codegen/projections/rails_json/lib/rails_json/middleware.rb
alextwoods: Not for rails json - I don't think it really makes sense to try to add. 

But theres good examples in white label with this.

codegen/projections/white_label/spec/event_stream_stubs_spec.rb
alextwoods: ugh, rubocop kept mucking these up.  I think I've fixed them all.

PR #216, opened 2024-08-20
Transport Agnostic Errors / EventStream errors - consolidate error handling

alextwoods:
*Description of changes:*
Previously the generated `Errors` module was specific to the HTTP Transport.  This PR makes the generated errors transport agnostic, allowing errors classes to be used as either operation response errors (with or without http transports) and event errors.  This required moving the creation of events into the parsers, which *do* have the transport specific knowledge and the the knowledge of which error class they map to. 

Consolidates the error handling for event stream errors to a single handler for consistency with the SDK specification.  Adds modeled event stream errors to new generated Errors::EventStream module. Note - these must be in a separate namespace as the error structures may also be the target of operation errors.


codegen/projections/white_label/lib/white_label/errors.rb
```diff
@@ -84,5 +84,29 @@ def initialize(http_resp:, **kwargs)
       attr_reader :data
     end
 
+    module EventStream
```
mullermp, line 87: Does it make sense to nest these? Shouldn't we make the service errors transport agnostic (and potentially event/response agnostic) instead?

codegen/projections/white_label/lib/white_label/errors.rb
```diff
@@ -84,5 +84,29 @@ def initialize(http_resp:, **kwargs)
...
+
+      # Base class for all event errors returned by this service
+      class Error < Hearth::ApiError; end
```
mullermp, line 90: If these must live like this, they should probably match ApiError and etc?

codegen/projections/white_label/sig/white_label/errors.rbs
```diff
@@ -47,6 +47,24 @@ module WhiteLabel
 
       attr_reader data: Types::ServerErrorEvent
     end
+    module EventStream
```
mullermp, line 50: newline

codegen/projections/rpcv2_cbor/lib/rpcv2_cbor/parsers.rb
```diff
@@ -29,14 +29,15 @@ def self.parse(list)
...
-      def self.parse(http_resp)
+      def self.parse(http_resp, **kwargs)
         data = Types::ComplexError.new
         body = http_resp.body.read
-        return data if body.empty?
-        map = Hearth::CBOR.decode(body)
-        data.top_level = map['TopLevel']
-        data.nested = (ComplexNestedErrorData.parse(map['Nested']) unless map['Nested'].nil?)
-        data
+        unless body.empty?
```
mullermp, line 35: Even if body is empty, wouldn't we parse an empty map easily here?

codegen/projections/white_label/lib/white_label/event_stream.rb
```diff
@@ -96,15 +82,38 @@ def on_unknown_event(&block)
...
         end
       end
+
+      def parse_exception_event(type, message)
+        case type
+        when 'ServerErrorEvent'
+          data = Parsers::EventStream::ServerErrorEvent.parse(message)
+          Errors::ServerErrorEvent.new(data: data, error_code: 'WhiteLabel::Types::Events::ServerErrorEvent')
+        else
+          data = Types::Events::Unknown.new(name: type || 'unknown', value: message)
```
mullermp, line 106: Maybe 'unknown' should be baked into the type as a default name? Maybe error code as well?

hearth/lib/hearth/http/api_error.rb
```diff
@@ -1,25 +0,0 @@
-# frozen_string_literal: true
```
mullermp, line 1: I think there's rbs for this file? Check to see if any other rbs needs updates.

hearth/spec/hearth/http/error_parser_spec.rb
```diff
@@ -23,12 +23,20 @@ class ApiClientError < ApiError; end
       class ApiServerError < ApiError; end
 
       class ModeledError < ApiClientError; end
+
+      module Parsers
```
mullermp, line 27: I think this should be under TestParsers, next to TestErrors? (Errors and Parsers are siblings right?)

mullermp (approved):
Nice!

codegen/projections/rpcv2_cbor/lib/rpcv2_cbor/parsers.rb
alextwoods: I think in at least some cases we used to fail on this (xml?) - it also means you can skip any parsing + checking for each field in the map when you already know you don't need to do any more parsing.  We used to just return the data here immediately, so I think this new logic is equivalent to that + the error class wrapper.

codegen/projections/white_label/lib/white_label/event_stream.rb
alextwoods: Yeah good call, I think that makes sense.

hearth/lib/hearth/http/api_error.rb
alextwoods: Good call removed and checked all other rbs.

hearth/spec/hearth/http/error_parser_spec.rb
alextwoods: Was trying to avoid extra namespaces.  The module arrangement doesn't really matter here, though I guess technically it doesn't reflect what we generate.  To be fully accurate we would need:

```
module TestService
  module Errors
     class ModeledError < ApiClientError; end # and all the other ones
   end

  module Parsers
      class ModeledError; end
  end
end
```

But it didn't really seem worth it to create more modules/nest more deeply here.

mullermp: In general I didn't like that we pollute the test namespace. There must be a fix.. maybe we should make a unit test service somehow? I guess like an API helper.

PR #218, opened 2024-09-11
Add customizations hook

mullermp:
Generate a blank customizations file that persists through copy task


codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/DirectedRubyCodegen.java
```diff
@@ -182,6 +183,9 @@ public void generateService(GenerateServiceDirective<GenerationContext, RubySett
         GlobalConfigPluginGenerator globalConfigPluginGenerator = new GlobalConfigPluginGenerator(directive);
         globalConfigPluginGenerator.render();
 
+        CustomizationsGenerator customizationsGenerator = new CustomizationsGenerator(directive);
+        customizationsGenerator.render();
```
alextwoods, line 187: As an alternative to always writing out a blank customization.rb what about having the servce.rb check for the existence of customization.rb before requiring it, eg:
```ruby
# lib/white_label.rb
# regular requires
require_relative 'white_label/validators'
require_relative 'white_label/waiters'
require_relative 'white_label/event_stream'
if file.exists?(File.join(__dir__, 'white_label/customizations.rb'))
  require_relative 'white_label/customizations'
end

```

alextwoods (commented):
I think we should be able to move all of the specs into the projection dirs instead of having codegen tasks copy them.

codegen/smithy-ruby-codegen-test/build.gradle.kts
```diff
@@ -146,22 +146,29 @@ tasks.register("generate-smithy-build") {
     }
 }
 
-
 tasks.register<Copy>("copyWhiteLabelGem") {
+    val src = "$buildDir/smithyprojections/smithy-ruby-codegen-test/white-label/ruby-codegen"
```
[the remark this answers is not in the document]
mullermp: Chose a different approach so not much is repeated.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/CustomizationsGenerator.java
```diff
@@ -0,0 +1,60 @@
...
+package software.amazon.smithy.ruby.codegen.generators;
+
+import java.util.logging.Logger;
+import software.amazon.smithy.codegen.core.directed.ContextualDirective;
+import software.amazon.smithy.ruby.codegen.GenerationContext;
+import software.amazon.smithy.ruby.codegen.RubySettings;
+import software.amazon.smithy.utils.SmithyInternalApi;
+
+@SmithyInternalApi
+public class CustomizationsGenerator extends RubyGeneratorBase {
```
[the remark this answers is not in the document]
mullermp: Deleted this file

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/DirectedRubyCodegen.java
mullermp: Good idea, I do a rescue here if the file exists or not.

codegen/smithy-ruby-codegen-test/build.gradle.kts
```diff
@@ -146,44 +146,29 @@ tasks.register("generate-smithy-build") {
...
     from("$buildDir/smithyprojections/smithy-ruby-codegen-test/rpcv2cbor/ruby-codegen")
     into("$buildDir/../../projections/")
 }
 
-tasks.register<Delete>("cleanProjections") {
-    delete("$buildDir/../../projections/white_label/")
-    delete("$buildDir/../../projections/rpcv2_cbor/")
-}
-
-tasks.register<Copy>("copyIntegrationSpecs") {
```
alextwoods, line 167: nice

codegen/smithy-ruby-codegen-test/integration-specs/auth_spec.rb
```diff
@@ -1,418 +0,0 @@
-# frozen_string_literal: true
-
-require_relative 'spec_helper'
```
alextwoods, line 3: I don't see these files added into the projections dir?

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/ModuleGenerator.java
```diff
@@ -32,7 +32,7 @@ public class ModuleGenerator {
             Logger.getLogger(ModuleGenerator.class.getName());
 
     private static final String[] DEFAULT_REQUIRES = {
-        "auth", "builders", "client", "config", "errors", "endpoint", "middleware",
+        "auth", "builders", "client", "config", "customizations", "errors", "endpoint", "middleware",
```
alextwoods, line 35: Since this is a special case in how we handle requiring it, I'd lean towards not including it in this list and handling it separately after all others. I think that order will be important as well and we want it last - files required from the customizations.rb may want to reference namespaces/classes,ect from the other generated files.

alextwoods (approved):

PR #219, opened 2024-10-13
Update CBOR to always have an accept header

mullermp:
Needs to update Smithy and test against protocol tests.


alextwoods (approved):

alextwoods (approved):
Looks good.

Worth adding a changelog entry for?

codegen/smithy-ruby-codegen/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 Unreleased Changes
 ------------------
 
+* Issue - Support services migrating from AWS Query to AWS JSON or RPC v2.
```
alextwoods, line 4: This AWS specific bit is kind of weird to have in our generic changelog.

codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/generators/ParserGeneratorBase.java
```diff
@@ -301,9 +300,9 @@ public void render(FileManifest fileManifest) {
     }
 
     protected void renderParsers() {
-        TreeSet<Shape> shapesToRender = CodegenUtils.getAlphabeticalOrderedShapesSet();
-        TreeSet<Shape> eventStreamEventsToRender = CodegenUtils.getAlphabeticalOrderedShapesSet();
-
+        Comparator<Shape> comparator = Comparator.comparing(o -> o.getId().getName() + " " + o.getId());
```
alextwoods, line 303: Why this change? The codegenutils helper method for this seems cleaner than implementing a comparator for each one?

alextwoods (approved):

codegen/smithy-ruby-codegen/CHANGELOG.md
mullermp: Yeah. I can change this "from AWS Query to RPC v2" - the query trait (AWS) is used generically now so that bit must stay.

PR #221, opened 2024-12-12
Decaf API 

jterapin:
*Description of changes:*
Implementation of shapes. While this is V3-inspired, there are some notable differences:
* API is now referenced as Schema.
* Supports various shape classes and prelude shapes live in smithy-client
* Added a mechanism to omit traits during codegeneration


gems/smithy/lib/smithy/anvil/client/views/shapes.rb
```diff
@@ -0,0 +1,238 @@
...
+              @name = name
+              @shape = shape
+              @traits = traits
+            end
+
+            attr_reader :name, :shape, :traits
+          end
+
+          # TODO: Not a complete list
+          OMITTED_TRAITS = %w[
```
jterapin, line 201: Thinking about moving these somewhere else.  Maybe a ShapesHelper, not sure yet.

I also think that there will be selective traits that we want to omit for service, operation and shapes (all three separate categories - MAYBE).

gems/smithy/lib/smithy/anvil/client/views/shapes.rb
```diff
@@ -0,0 +1,238 @@
...
+          end
+
+          private
+
+          def assemble_shapes
+            serializable_shapes = []
+            operation_shapes = []
+            service_shape = nil
+
+            @model.shapes.each_value do |v|
```
jterapin: I haven't looked into how the new files (OperationParser) in Vise works yet but I _kind_ of like  doing this single pass and creating class instances that carries relevant data that we only care about for `shapes.rb`

demo/client.rb
```diff
@@ -0,0 +1,57 @@
...
+
+
+    # TODO
+    def initialize(*args)
+      super(*args)
+    end
+
+    # TODO!
+    def get_city(params = {}, options = {})
+      input = build_input(:get_city, "example.weather#GetCity", params)
```
jterapin: Could just keep the symbolized ruby name here to pass to `build_input`  - I didn't want to muddle the existing usage for the `operation_name` for handlers.

demo/client.rb
```diff
@@ -0,0 +1,57 @@
...
+      input.send_request(options)
+    end
+
+    private
+
+    def build_input(operation_name, operation_id, params)
+      handlers = @handlers.for(operation_name)
+      context = Smithy::Client::HandlerContext.new(
+        operation_name: operation_name,
+        operation: config.service_shape.operation(operation_id),
```
jterapin: Sets the `OperationShape` instance to the context. I actually might rename `operation` here to `operation_shape` to indicate that it just holds shape data.

demo/shapes.rb
```diff
@@ -0,0 +1,172 @@
...
+      input: GetForecastInput,
+      output: GetForecastOutput,
+    )
+    ListCities = OperationShape.new(
+      shape_id: "example.weather#ListCities",
+      input: ListCitiesInput,
+      output: ListCitiesOutput,
+    )
+
+    WEATHER = ServiceShape.new(
```
jterapin: Could I expand on weather model to include all different type of shapes or wait until we bring in WhiteLabel? Not sure what we exactly want to do.

gems/smithy-client/lib/smithy-client/base.rb
```diff
@@ -166,24 +166,24 @@ def plugins
           Array(@plugins).freeze
         end
 
-        # @return [API]
-        def api
-          @api ||= API.new
+        # @return [Shapes::ServiceShape]
+        def service_shape
+          @service_shape ||= Shapes::ServiceShape.new
```
jterapin: I felt that `service_shape` is more appropriate since it is indeed a shape.

jterapin (commented):
I wanted to demonstrate what I imagine the new "`client_api.rb`" will look so writing out tests and some optimization are skipped under smithy.

Currently passing spec (rubocop is failing due to the demo files 😅)

demo/client.rb
mullermp: If we know operation names are globally unique across namespaces, I think we should only use symbolized name here, especially if handlers and plugins have ability to check against operation names. Model purity is not necessary, we build our own runtime based on the model.

demo/shapes.rb
```diff
@@ -0,0 +1,172 @@
+# frozen_string_literal: true
+
+# This is generated code!
+
+module Weather
+  module Shapes
+    include Smithy::Client::Shapes
+
+    CityCoordinates = StructureShape.new(
```
mullermp: I think it's fine to single line these. This file will get exceptionally large otherwise.

demo/shapes.rb
```diff
@@ -0,0 +1,172 @@
+# frozen_string_literal: true
```
mullermp: What is this file? If we want to check in the build to source control, we should do that instead.

demo/shapes.rb
```diff
@@ -0,0 +1,172 @@
...
+    ListCities = OperationShape.new(
+      shape_id: "example.weather#ListCities",
+      input: ListCitiesInput,
+      output: ListCitiesOutput,
+    )
+
+    WEATHER = ServiceShape.new(
+      shape_id: "example.weather#Weather",
+      version: "2006-03-01",
+    ).tap do | service |
```
mullermp: You shouldn't need tap, you already do `yield self` on new.

demo/shapes.rb
```diff
@@ -0,0 +1,172 @@
...
+      service.add_operation("example.weather#GetCity", GetCity)
+      service.add_operation("example.weather#GetCurrentTime", GetCurrentTime)
+      service.add_operation("example.weather#GetForecast", GetForecast)
```
mullermp: Symbolized names are still preferred otherwise all plugins/handlers will need to know the model's namespace and full operation name, even after a transform.

demo/shapes.rb
mullermp: We will want a fixture that kitchen sinks all the different types that we care about. I think white label from the old branch is too bloated but some parts of it would be applicable to this test. We can come up with something offline.

demo/client.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+# This is generated code!
+
+module Weather
+  class Client < Smithy::Client::Base
+    self.service_shape = Shapes::WEATHER
```
mullermp: I thought we were calling this API? I think from the perspective of the client base, it shouldn't care if it gets a "service shape" - my preference would be `self.api = Shapes::SERVICE` or API::SERVICE where we just use SERVICE constant always. Making this unique doesn't really matter.

gems/smithy-client/lib/smithy-client/shapes.rb
```diff
@@ -0,0 +1,213 @@
...
+      class Shape
+        def initialize(options = {})
+          @shape_id = options[:shape_id]
+          @traits = options[:traits] || {}
+        end
+
+        # @return [String, nil]
+        attr_reader :shape_id
+
+        # @return [Hash<String, [Boolean, Hash, Integer, String]>]
```
mullermp: https://yardoc.org/types.html

I believe you need Hash{String => Boolean, Hash, Integer, String} if you want multi values. Otherwise, I'ver been using Hash<String, Object> and that is sufficient.

gems/smithy-client/lib/smithy-client/shapes/prelude_shapes.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    module Shapes
+      module PreludeShapes
+        String = StringShape.new(shape_id: 'smithy.api#String')
```
mullermp: Alphabetize. You can also single line these and remove rubocop checks.

gems/smithy-client/lib/smithy-client/shapes/prelude_shapes.rb
```diff
@@ -0,0 +1,55 @@
...
+        Timestamp = TimestampShape.new(shape_id: 'smithy.api#Timestamp')
+        Document = DocumentShape.new(shape_id: 'smithy.api#Document')
+        Short = IntegerShape.new(shape_id: 'smithy.api#Short')
+        Integer = IntegerShape.new(shape_id: 'smithy.api#Integer')
+        Long = IntegerShape.new(shape_id: 'smithy.api#Long')
+        Float = FloatShape.new(shape_id: 'smithy.api#Float')
+        Double = FloatShape.new(shape_id: 'smithy.api#Double')
+        PrimitiveBoolean = BooleanShape.new(
+          shape_id: 'smithy.api#PrimitiveBoolean',
+          traits: { 'smithy.api#default' => false }
```
mullermp: I don't think we should codify defaults of any kind. These can get stale and cause conflicts in versioning when they need to change. For example, there was significant work to handle new defaults/nullability. We should let the model inform whether these traits exist.

gems/smithy/lib/smithy/anvil/client/views/module.rb
```diff
@@ -38,7 +38,7 @@ def requires
             if @plan.type == :types
               [:types]
             else
-              %i[client errors types]
+              %i[types shapes client errors]
```
mullermp: Alphabetize

gems/smithy/lib/smithy/anvil/client/views/shapes.rb
mullermp: Yes, that is the pattern, to have "view classes" that are used to render your view "shapes". Vise is there to help iterate the model and give convenience.

gems/smithy/lib/smithy/anvil/client/views/shapes.rb
```diff
@@ -0,0 +1,238 @@
...
+            Tools::Namespace.namespace_from_gem_name(@plan.options[:gem_name])
+          end
+
+          def shapes_with_members
+            @shapes.reject { |s| s.members.empty? }
+          end
+
+          private
+
+          def assemble_shapes
```
mullermp: I would consider a private ShapesParser in anvil for hiding away most of this logic. You can define the "view classes" here though and populate them with data the shapes parser gives you. Also look at what vise gives you now.

gems/smithy/lib/smithy/anvil/client/views/shapes.rb
```diff
@@ -0,0 +1,238 @@
...
+                filter_member_traits(v['traits'])
+              )
+            end
+          end
+
+          def assemble_shape_name(shape_id)
+            if PRELUDE_SHAPES_MAP.include?(shape_id)
+              PRELUDE_SHAPES_MAP[shape_id]
+            else
+              shape_id.split('#').last
```
mullermp: This is the kind of stuff Vise is supposed to help with. This is the relative_id method right? Perhaps prelude shapes should be loaded into the model somehow.

gems/smithy/lib/smithy/anvil/client/views/shapes.rb
mullermp: I don't think we should be omitting traits. I suppose documentation can get lengthy so maybe we want to snip the value. I think we may want to check any trait exists on any shape in the runtime.

gems/smithy-client/lib/smithy-client/shapes/prelude_shapes.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    module Shapes
+      module PreludeShapes
```
mullermp: I think whether these are prelude shapes or not doesn't matter, they are shapes. I think we should include them where all the other shapes are defined rather than including this module always. Possibly these should be added to code generated output such that if we need defaults listed in prelude, they will change with service update and not runtime update. Can team discuss.

mullermp (commented):
Nice so far

demo/client.rb
jterapin: Updating the constant to just `SERVICE` makes sense. I'll update that.

I keep walking back and forward with "API" and "Service Shape"...  I was also thinking of `self.service` or `self.shape` 

To me, I don't think "API" is a good name to describe what the contents of the `Shapes::Service` are. API feels very vague and broad. I also think `shapes` is a bloated term as well.


Another thought ....  what about schemas a try? 
I believe that's how other SDKs described our shapes instead. 

So in `client.rb`, it would be `self.schema = Schemas::SERVICE` 

I would have to change up shapes -> schemas in other areas. 
Instead of PreludeShapes, it would be PreludeSchemas and such.

We could take this discussion offline but I would like to see Alex weigh in as well.

demo/shapes.rb
jterapin: I just wanted to show what it looks like to get thoughts. I plan to revert this commit (but yeah I could've just commited the contents of the build/ stuff - probably better to get the full contet )

demo/shapes.rb
jterapin: Cool, I will make that change.

gems/smithy-client/lib/smithy-client/shapes/prelude_shapes.rb
jterapin: In V3, these were included in the codegenerated output. Since PreludeShapes are reused across different services, I figured it made sense to have these in smithy-client. Yup, we could do a team discussion.

gems/smithy-client/lib/smithy-client/shapes/prelude_shapes.rb
jterapin: From what I understand, these PreludeShapes are public and should not be changed: https://smithy.io/2.0/spec/model.html#prelude

Other SDKs who have implemented PreludeShapes (PreludeSchemas) have these shapes defined and it includes the default trait as well.

If we want the model to drive this, I think I will be fine with that - just need to ensure that model includes prelude shapes defined. Thinking a bit forward when we implement V4 - how are aws service model files distributed? Will they include prelude shapes?

gems/smithy/lib/smithy/anvil/client/views/module.rb
jterapin: I don't think we can alphabetize these - these files need to be required in a certain order or we get an ` uninitialized constant` error of some kind.

Above order outputs the following in the module file:
```
require_relative 'weather/types'
require_relative 'weather/shapes'
require_relative 'weather/client'
require_relative 'weather/errors'
```

gems/smithy/lib/smithy/anvil/client/views/shapes.rb
jterapin: Yeah I want to refactor this section once the new Vise PR gets merged in.

gems/smithy/lib/smithy/anvil/client/views/shapes.rb
jterapin: We probably should have a team discussion on prelude shapes (and few other aspects). I can start few threads in our channel with the context for the entire team to input.

gems/smithy/lib/smithy/anvil/client/views/shapes.rb
jterapin: Far as I know, the documentation trait are used to render documentation during codegen - so I don't think we care to bring them over.

I was under the impression that the intention of these shapes is to drive the (de)serialization process. Are there any other use cases for a reason to have all traits defined?

gems/smithy-client/lib/smithy-client/shapes.rb
jterapin: I had these in that format before: https://github.com/smithy-lang/smithy-ruby/commit/0ba548a71821fcc0001c7589a1bef6b5e5e147f6

When I read your offline comment on the formatting - I was thinking that I was on wrong on both yard formatting (single value type hash vs multiple value types hash). I should have clarified.

So ... we do the following:
Single Value type Hash: `Hash<KeyType, ValueType>`
Multiple Value types Hash: `Hash{KeyTypes=>ValueTypes}`

Can we do the following for the single value type hash? `Hash{String => Symbol}` to match the format consistency with multiple value types hash?

When I use the YARD Type Parser, it seems ok... but i noticed that parser is experimental and not "official"
```
a Hash with keys made of (Strings) and values of (Symbols)
```

Found the following documentation on [yardoc tags](https://rubydoc.info/gems/yard/file/docs/Tags.md):
> Hashes can be specified either via the parametrized type discussed above, in the form Hash<KeyType, ValueType>, or using the hash specific syntax: Hash{KeyTypes=>ValueTypes}. In the latter case, KeyTypes or ValueTypes can also be a list of types separated by commas.

demo/client.rb
jterapin: I was looking how build inputs param for `operation_name` are used in plugins and handlers. it seems like we always access these through the context object. Since the context object will have operation name and operation shape (operation.api) so I think the check will be fine?   I might be lookin at the wrong places so let me know.

I also think OperationShape should have an instance method that provides rubyified method name. But the structure of operations for service shape should follow shape id.

demo/client.rb
mullermp: Yeah, let's discuss. API includes a service shape, as well as other shapes.

gems/smithy-client/lib/smithy-client/shapes.rb
mullermp: Whatever we choose, let's just be consistent. I think Hash<String, Object> is fine, since the value can be any number of things. I think RBS can be more specific about boolean/hash/integer/string.

gems/smithy/lib/smithy/anvil/client/views/shapes.rb
alextwoods: Yeah - I think its correct to remove some traits.  At minimum documentation we really should remove.   And maybe we just start with that.  I'm sure smithy-java will come up with a good list we can steal later, but I do think its important to exclude traits that are not needed at runtime, so I think this mechanism is good.

demo/client.rb
alextwoods: Yeah, this is a good general question.  I could see us using: `operation` for the name and `operation_shape` for the shape.  OR `operation` for the shape and `operation_name` for the name...

But - we don't really need both name and shape right?  We really only need the operation shape since that can tell us the name.  So I actually lean towards just having a single member `operation` (that is the shape)

demo/shapes.rb
alextwoods: We created the white label model slowly over time to cover all the features we need, so I wouldn't call it bloated.  But Its fair to say we don't need/want all of it right now.  

In general for working on endpoints I've been just generating test SDKs from the fixtures (either in rspec to iinspect behavior or, using a simple command if I want to inspect the files).  Heres what I've been using:

```
bundle exec smithy-ruby smith client --gem-name fixture --gem-version 1.0.0 <<< $(cat gems/smithy/spec/fixtures/endpoints/default-values/model.json)

```

gems/smithy/lib/smithy/anvil/client/views/shapes.rb
jterapin: Per our team discussion, we agreed to have the mechanism to omit traits but start out with just omitting the documentation trait.

demo/client.rb
jterapin: Per team discussion, we are going to go with symbolized ruby names as the key here since the operation shape will have the shape id as one of their shape attributes

demo/client.rb
jterapin: I think I'm just going to leave this as it is for now.

gems/smithy-client/lib/smithy-client/base.rb
jterapin: Update: changing to `schema` per team discussion

demo/shapes.rb
jterapin: Per team discussion, going with symbolized ruby names: https://github.com/smithy-lang/smithy-ruby/pull/221#discussion_r1896009136

demo/client.rb
jterapin: per team discussion - going with Schema naming

gems/smithy-client/lib/smithy-client/shapes.rb
jterapin: Agreed, going with `Hash<String, Object>`

gems/smithy-client/lib/smithy-client/shapes/prelude_shapes.rb
jterapin: Per team discussion, we need to keep the default as part of the prelude primitive shapes but move all prelude shapes under Shapes module to flatten it

gems/smithy-client/lib/smithy-client/shapes/prelude_shapes.rb
jterapin: see: https://github.com/smithy-lang/smithy-ruby/pull/221#discussion_r1896011649

for the team discussion result

build/smithy/weather/smithy-ruby/lib/weather/client.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+# This is generated code!
+
+module Weather
```
jterapin: Self-reminder to revert these commit

.rubocop.yml
```diff
@@ -10,20 +10,25 @@ AllCops:
 Gemspec/RequireMFA:
   Enabled: false
 
+Metrics/AbcSize:
+  Exclude:
+    - 'gems/smithy/lib/smithy/forge/client.rb'
+    - 'gems/smithy/lib/smithy/anvil/client/views/shapes.rb'
```
mullermp: Fine for now, but please fix.

gems/smithy-client/lib/smithy-client/shapes.rb
```diff
@@ -0,0 +1,225 @@
...
+        attr_accessor :name
+
+        # @return [Shape]
+        attr_accessor :shape
+
+        # @return [Hash<String, Object>]
+        attr_accessor :traits
+      end
+
+      PreludeBigDecimal = IntegerShape.new(shape_id: 'smithy.api#BigDecimal')
```
mullermp: I think the `Prelude` prefix is overly verbose and not necessary. I think make these 1:1 with the API, i.e. BigDecimal.

gems/smithy/lib/smithy/anvil/client/templates/client_class.erb
```diff
@@ -29,14 +29,14 @@ module <%= namespace %>
...
+        operation: config.schema.operation(operation_name),
         client: self,
         params: params,
-        config: config
+        config: config,
       )
       context[:gem_name] = '<%= gem_name %>'
       context[:gem_version] = '<%= gem_version %>'
-      Smithy::Client::Input.new(handlers, context)
+      Smithy::Client::Input.new(handlers: handlers, context: context)
```
mullermp, line 42: Since Input is API private, these are fine to be positional params. Input class I think may have options hash on accident. This is mostly a note for me later (or if you want to quickly fix that).

gems/smithy/lib/smithy/anvil/client/templates/shapes.erb
```diff
@@ -0,0 +1,57 @@
...
+       <%- unless shape.traits.empty? -%>
+        operation.traits = <%= shape.traits -%>
+       <% end -%>
+       <%- unless shape.errors.empty? -%>
+       <% shape.errors.each do |err| %>
+        operation.errors << <%= err -%>
+       <% end -%>
+      <% end %>
+      end)
+    <% end %>
```
mullermp: Please check all usage of <%- and -%> for correct trim formatting.

gems/smithy/lib/smithy/anvil/client/templates/shapes.erb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+# This is generated code!
+
+module <%= namespace %>
+  module Shapes
+    include Smithy::Client::Shapes
+
+<%# Shapes definition -%>
```
mullermp: The comments in this file are not really necessary here.. it's clear these are the shapes.

gems/smithy/lib/smithy/anvil/client/templates/shapes.erb
```diff
@@ -0,0 +1,57 @@
...
+  <%- @shapes.each do |shape| -%>
+    <%- params = ["id: \"#{shape.id}\""] -%>
```
mullermp: This entire block is really wonky rendering. If you need to embed multiple lines, you can do so with

```
<%-
  some
  ruby
  code
-%>
```
but also this mostly looks static anyway. For example this line should probably instead be

```
params = ["id: "<%= shape.id %>"]
```

gems/smithy/lib/smithy/anvil/client/templates/shapes.erb
```diff
@@ -0,0 +1,57 @@
...
+    <%- params << "type: Types::#{shape.name}" if shape.typed -%>
+    <%- params << "traits: #{shape.traits}" unless shape.traits.empty? -%>
+    <%= "#{shape.name} = #{shape.shape_type}.new(" %><%= params.join(', ') %>)
```
mullermp: Is this trying to render keyword arguments? This is not a great way to do this. Perhaps you should prepare the params in a view and then iterate over them here.

gems/smithy/lib/smithy/anvil/client/templates/shapes.erb
```diff
@@ -0,0 +1,57 @@
...
+      )
+    <%- @operation_shapes.each do |shape| %>
+      schema.add_operation(:<%= shape.name %>, OperationShape.new do |operation|
+        operation.id = "<%= shape.id %>"
+        operation.input = <%= shape.input %>
+        operation.output = <%= shape.output %>
+       <%- unless shape.traits.empty? -%>
+        operation.traits = <%= shape.traits -%>
+       <% end -%>
+       <%- unless shape.errors.empty? -%>
```
mullermp: Enumerable empty check is not necessary here - the iterator below will just render nothing.

gems/smithy/lib/smithy/anvil/client/templates/shapes.erb
```diff
@@ -0,0 +1,57 @@
...
+       <% shape.errors.each do |err| %>
+        operation.errors << <%= err -%>
```
mullermp: Should not end in -%> for an <%=. Please inspect rendered code.

gems/smithy/lib/smithy/anvil/client/templates/shapes.erb
```diff
@@ -0,0 +1,57 @@
...
+        id: "<%= @service_shape.id %>",
+        version: "<%= @service_shape.version %>",
+        traits: <%= @service_shape.traits %>
+      )
+    <%- @operation_shapes.each do |shape| %>
+      schema.add_operation(:<%= shape.name %>, OperationShape.new do |operation|
+        operation.id = "<%= shape.id %>"
+        operation.input = <%= shape.input %>
+        operation.output = <%= shape.output %>
+       <%- unless shape.traits.empty? -%>
```
mullermp: I don't think we care to check this. Have it set a hash or nil to this method.

gems/smithy/lib/smithy/anvil/client/views/shapes.rb
```diff
@@ -0,0 +1,241 @@
...
+            Tools::Namespace.namespace_from_gem_name(@plan.options[:gem_name])
+          end
+
+          def shapes_with_members
+            @shapes.reject { |s| s.members.empty? }
+          end
+
+          private
+
+          def assemble_service_shape
```
mullermp: Naming nit, build_ is a fine prefix. But also, this could simply just be a method called service_shape? There is no need to make an instance variable IMO. Possibly the same for other cases.

gems/smithy/lib/smithy/anvil/client/templates/shapes.erb
```diff
@@ -0,0 +1,57 @@
...
+    <%= "#{shape.name}.set_member_#{member.name}(" %><%= params.join(', ') %>)
+    <%- else -%>
+    <%= "#{shape.name}.add_member(" %><%= params.join(', ') %>)
+    <%- end -%>
+    <%- end -%>
+  <%- end -%>
+<%# Schema definition %>
+    SCHEMA = Smithy::Client::Schema.new do |schema|
+      schema.service = ServiceShape.new(
+        id: "<%= @service_shape.id %>",
```
mullermp: Did you mean to reference the method? Let's not reference instance variables

gems/smithy/spec/fixtures/shapes/model.smithy
```diff
@@ -0,0 +1,115 @@
...
+        SomeResource
+    ]
+    operations: [
+        SomeOperation
+    ]
+}
+
+operation SomeOperation {
+    input: SomeOperationInput
+    output: SomeOperationOutput
```
mullermp: I think you'll want something more kitchen sink for both input and output. SomeOperationInputOutput probably where it has some structure that contains all of the shapes. It should also reference itself, which would make it nested/recursive. We will eventually need it on both input and output for building and parsing.

gems/smithy/spec/fixtures/shapes/model.smithy
```diff
@@ -0,0 +1,115 @@
+$version: "2"
+
+namespace smithy.ruby.tests
+
+@paginated(inputToken: "nextToken", outputToken: "nextToken", pageSize: "pageSize")
+service ShapeService {
+    version: "2018-10-31"
+    resources: [
+        SomeResource
```
mullermp: Probably not necessary to have a resource anyway. A single operation is fine.

gems/smithy/spec/fixtures/shapes/model.smithy
```diff
@@ -0,0 +1,115 @@
...
+
+    @required
+    string: String
+
+    timestamp: Timestamp
+}
+
+@output
+structure SomeOperationOutput {
+    list: SomeList
```
mullermp: Naming nit, can probably drop the Some prefixing. It's fine for operation or structure name.

gems/smithy/spec/interfaces/client/shapes_spec.rb
```diff
@@ -0,0 +1,163 @@
+# frozen_string_literal: true
+
+describe 'Component: Shapes' do
+  shape_tests = JSON.load_file(
+    File.expand_path('../../fixtures/shapes/shape_tests.json', __dir__)
```
mullermp: Instead of having this file, which would require updates when the model updates, and also is not quite a test file but rather a reference, why not reference the model.json fixture to assert your tests? This seems very unmaintainable.

gems/smithy-client/lib/smithy-client/base.rb
```diff
@@ -163,24 +163,24 @@ def plugins
           Array(@plugins).freeze
         end
 
-        # @return [API]
-        def api
-          @api ||= API.new
+        # @return [Schema]
+        def schema
+          @schema ||= Schema.new
```
alextwoods, line 167: I'm not sure we should return an empty schema if none exists from this

gems/smithy-client/lib/smithy-client/base.rb
```diff
@@ -163,24 +163,24 @@ def plugins
...
         end
 
-        # @param [API] api
-        def api=(api)
-          @api = api
+        # @param [Schema] schema
+        def schema=(schema)
+          @schema = schema
           define_operation_methods
```
alextwoods, line 173: Why are we defining the operation methods here? I think this might cause the methods to effectively be double defined since we codegen them as well.

I'm a little uncertain about having this method at all - I know we're using it in our code generated clients to do `self.schema = ... ` but I find it a little weird to allow mutability of a classes schema.

gems/smithy-client/lib/smithy-client/shapes.rb
```diff
@@ -0,0 +1,225 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    # Represents shape types from Smithy Model
+    module Shapes
+      # A base shape that all shapes inherits from
+      class Shape
+        def initialize(options = {})
+          @id = options[:id]
```
alextwoods, line 10: I'm wondering if we should have any validation - I think a shape is invalid without an ID

gems/smithy-client/lib/smithy-client/shapes.rb
```diff
@@ -0,0 +1,225 @@
...
+      class EnumShape < Shape
+        def initialize(options = {})
+          super
+          @members = {}
+        end
+
+        # @return [Hash<String, MemberShape>]
+        attr_accessor :members
+
+        # @return [MemberShape, nil]
```
alextwoods: when would this return nil?

gems/smithy-client/lib/smithy-client/shapes.rb
```diff
@@ -0,0 +1,225 @@
...
+        def set_member(shape, traits: {})
+          @member = MemberShape.new('member', shape, traits: traits)
+        end
+      end
+
+      # Represents a Map shape
+      class MapShape < Shape
+        def initialize(options = {})
+          super
+          @member_key = nil
```
alextwoods: we could potentially just call these "key" and "value" and drop the member prefix.

gems/smithy/lib/smithy/anvil/client/templates/shapes.erb
jterapin: I just think its help for readability at glance but I can remove them.

gems/smithy-client/lib/smithy-client/shapes.rb
jterapin: Yeah it's verbose and I could change to just `BigDecimal` - I'm was concerned about how BigDecimal could be taken in few different ways - like `BigDecimal` module vs the IntegerShape, `BigDecimal` - I thought adding `Prelude` prefix helps to clarify its intent at glance

gems/smithy/spec/fixtures/shapes/model.smithy
jterapin: I think we should have a resource here to ensure that the operation defined within the resource is created

gems/smithy/spec/interfaces/client/shapes_spec.rb
jterapin: Yeah I can take a look at that possibility

gems/smithy/lib/smithy/anvil/client/templates/shapes.erb
jterapin: Yeah was planning to refine on the `shapes.erb` file this morning - thanks for the comments.

gems/smithy/lib/smithy/anvil/client/templates/client_class.erb
jterapin: Sure I can include this fix

gems/smithy-client/lib/smithy-client/shapes.rb
alextwoods: Maybe these belong in a module named Prelude?  

I do agree the name should probably include Prelude to keep them from conflicting with the shape classes.

gems/smithy/lib/smithy/anvil/client/views/shapes.rb
alextwoods: I'd lean towards a build_ prefix for these.

gems/smithy/lib/smithy/anvil/client/views/shapes.rb
```diff
@@ -0,0 +1,241 @@
...
+            serializable_shapes = []
+            operation_shapes = []
+
+            @model['shapes'].each do |id, shape|
+              case shape['type']
+              when 'service', 'resource' then next
+              when 'operation'
+                operation_shapes << assemble_operation_shape(id, shape)
+              else
+                serializable_shapes << assemble_shape(id, shape)
```
alextwoods: How do we know these are serializable?  I assume this is more about rendering the runtime model and not having checking traits to see if a shape is serializable or not.

gems/smithy/lib/smithy/anvil/client/templates/shapes.erb
jterapin: I was under the impression that `<%= EXPRESSION %>` could end with `-%> ` to trim the following line break.

Rendered code based on `<%= err -%>`:
```
schema.add_operation(:get_city, OperationShape.new do |operation|
  operation.id = "example.weather#GetCity"
  operation.input = GetCityInput
  operation.output = GetCityOutput
  operation.traits = {"smithy.api#readonly"=>{}}              
  operation.errors << NoSuchResource             
end)
```
Where as,  `<%= err %>` will produce:
```
schema.add_operation(:get_city, OperationShape.new do |operation|
  operation.id = "example.weather#GetCity"
  operation.input = GetCityInput
  operation.output = GetCityOutput
  operation.traits = {"smithy.api#readonly"=>{}}              
  operation.errors << NoSuchResource
             
end)
```

gems/smithy/lib/smithy/anvil/client/views/shapes.rb
```diff
@@ -0,0 +1,241 @@
...
+              traits: filter_traits(shape_data['traits']),
+              version: shape_data['version']
+            )
+          end
+
+          def assemble_shapes
+            serializable_shapes = []
+            operation_shapes = []
+
+            @model['shapes'].each do |id, shape|
```
alextwoods: This is where we need a service walker or some sort of "index" - the model may have shapes that are not part of the service's closure (ie, the graph of shapes connected to the service through its operations).

gems/smithy/lib/smithy/anvil/client/templates/shapes.erb
jterapin: Nice, good catch.

gems/smithy/lib/smithy/anvil/client/templates/shapes.erb
jterapin: Yeah I was torn whether we want to render empty hashes or not.

alextwoods (commented):
Nice - looking good overall

gems/smithy-client/lib/smithy-client/shapes.rb
jterapin: I think there would be always a `id` attached to a shape when `smithy build` runs.

gems/smithy-client/lib/smithy-client/shapes.rb
jterapin: Good call, I'll fix.

gems/smithy-client/lib/smithy-client/base.rb
jterapin: I'm guessing this section was brought over from V3. I could see if anything breaks when I set this to nil. Maybe Matt could pitch in for his thoughts?

gems/smithy-client/lib/smithy-client/shapes.rb
mullermp: My earlier comment was to remove the prelude module. What BigDecimal module are you referring to? That's on the root namespace while this is under Smithy Client. I don't see any issue here.

mullermp: Shapes are shapes, I don't think we care if they are prelude or not.

gems/smithy-client/lib/smithy-client/base.rb
jterapin: I'll take a look about this intent. I want to say that this is brought over from V3: https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-core/lib/seahorse/client/base.rb#L177-L181

gems/smithy/lib/smithy/anvil/client/templates/shapes.erb
mullermp: I believe that is because your outer loop does not have dash trimming.

gems/smithy/lib/smithy/anvil/client/templates/shapes.erb
mullermp: I don't understand. Either an empty hash or nil will be assigned here. We wouldn't want an empty hash.

gems/smithy/spec/fixtures/shapes/model.smithy
mullermp: We already test that operations are provided in the operation index if they are defined in resources.

gems/smithy-client/lib/smithy-client/base.rb
mullermp: It's useful for testing and I think an empty API is fair.

gems/smithy/lib/smithy/anvil/client/templates/shapes.erb
jterapin: Yeah, I was going to fine-tune this

gems/smithy-client/lib/smithy-client/shapes.rb
mullermp: I don't think we care. This is validated by the smithy CLI when we get the model. Code generation would also break on lack of ID. I think we assume everything comes in and use options hash here is fine.

gems/smithy/lib/smithy/anvil/client/templates/shapes.erb
jterapin: I was trying to think of ways to make this easier to read since we want all the params rendered in a single line. Yeah, I'll revisit.

gems/smithy/lib/smithy/anvil/client/views/shapes.rb
mullermp: Yes, serializable is not good naming. They are just shapes to be rendered.

gems/smithy/lib/smithy/anvil/client/views/shapes.rb
mullermp: Yes. I will work on adding this soon.

gems/smithy/lib/smithy/anvil/client/views/shapes.rb
jterapin: Yeah I think my naming could be better here, will advise

gems/smithy/lib/smithy/anvil/client/templates/shapes.erb
jterapin: My intention was that if `OperationShape` is created without a traits param, it will default to an empty hash. Therefore, here I want to check to see if the traits exists to worth rendering

gems/smithy/lib/smithy/anvil/client/templates/shapes.erb
jterapin: Yeah that fixes the above operation and the following operations will have a weird formatting so I had to keep that dash.

gems/smithy/lib/smithy/anvil/client/templates/shapes.erb
mullermp: I think you can simply just assign shape.traits, and shape.traits will either be a hash with stuff or nil.

gems/smithy-client/lib/smithy-client/base.rb
mullermp: Everything is mutable anyway? How would you not make this mutable? I think it's fine to inform the base class of the schema.

gems/smithy-client/lib/smithy-client/shapes.rb
jterapin: Ahh, I think was thinking about how I tried naming all the Shape classes without suffix so `BigDecimal` class was having issues. Anyway, will remove prefix to see how that goes.

gems/smithy/spec/fixtures/shapes/model.smithy
jterapin: That's fair, I will remove.

projections/weather/lib/weather/shapes.rb
```diff
@@ -0,0 +1,92 @@
...
+    GetForecastInput = StructureShape.new(id: "example.weather#GetForecastInput", traits: {"smithy.api#input"=>{}})
+    GetForecastOutput = StructureShape.new(id: "example.weather#GetForecastOutput", traits: {"smithy.api#output"=>{}})
+    ListCitiesInput = StructureShape.new(id: "example.weather#ListCitiesInput", traits: {"smithy.api#input"=>{}})
+    ListCitiesOutput = StructureShape.new(id: "example.weather#ListCitiesOutput", traits: {"smithy.api#output"=>{}})
+    NoSuchResource = StructureShape.new(id: "example.weather#NoSuchResource", traits: {"smithy.api#error"=>"client"})
+
+    CityCoordinates.add_member("latitude", Float, {"smithy.api#required"=>{}})
+    CityCoordinates.add_member("longitude", Float, {"smithy.api#required"=>{}})
+    CityCoordinates.type = Types::CityCoordinates
+
```
jterapin, line 26: I could omit these whitelines tbh

gems/smithy-client/lib/smithy-client/base.rb
jterapin: Leaving as it is per our team discussion

gems/smithy-client/lib/smithy-client/base.rb
jterapin: Leaving as it is per our team discussion

gems/smithy/lib/smithy/anvil/client/templates/client_class.erb
jterapin: Punting this to another PR

gems/smithy/spec/fixtures/shapes/model.smithy
```diff
@@ -0,0 +1,86 @@
...
+    short: Short
+    string: String
+    timestamp: Timestamp
+
+    // member with trait
+    @required
+    id: String
+
+    // complex members
+    structure: OperationInputOutput
```
alextwoods, line 45: This is meant to be recursive right? Its a little weird to use the OperationInputOutput inside the OperationInputOutput and indeed the `@output` trait forbids it: https://smithy.io/2.0/spec/type-refinement-traits.html#output-trait

Its great to include a recursive shape in here, but I think we may want to not use the Input/Output?

projections/weather/lib/weather/shapes.rb
alextwoods: Its only one line in between these right?  I think the seperation is useful since its effectively a different section

projections/weather/lib/weather/shapes.rb
```diff
@@ -0,0 +1,92 @@
+# frozen_string_literal: true
+
+# This is generated code!
+
+module Weather
+  module Shapes
+    include Smithy::Client::Shapes
+
+    CityCoordinates = StructureShape.new(id: "example.weather#CityCoordinates", traits: {})
+    CityId = StringShape.new(id: "example.weather#CityId", traits: {"smithy.api#pattern"=>"^[A-Za-z0-9 ]+$"})
```
alextwoods: Just a general comment/discussion on how we render hashes for things like this.  I ran into the same issue in the endpoint specs.  

This format is easy to write (since it relies on the default to_s method of a Hash), but produces code that is a (potentially very long) single line and violates rubocop rules (double quotes, missing spaces around brackets).  

We could create a utility for rendering these as multiline... or just leave it like this.  Thoughts?

gems/smithy/spec/fixtures/shapes/model.smithy
alextwoods: Edit: NVM - I see Smithy's kitchen sink does the same thing: https://github.com/smithy-lang/smithy/blob/main/smithy-aws-protocol-tests/model/awsJson1_1/kitchen-sink.smithy#L1035

alextwoods (approved):
Nice! This is looking good

gems/smithy-client/lib/smithy-client/shapes.rb
```diff
@@ -0,0 +1,225 @@
...
+        def initialize(options = {})
+          super
+          @members = {}
+          @type = nil
+        end
+
+        # @return [Hash<String, MemberShape>]
+        attr_accessor :members
+
+        # @return [Type]
```
mullermp: Struct? Class? Type isn't a real class.

gems/smithy/lib/smithy/anvil/client/templates/shapes.erb
```diff
@@ -0,0 +1,47 @@
...
+  module Shapes
+    include Smithy::Client::Shapes
+
+<% shapes.each do |shape| -%>
+    <%= shape.name %> = <%= shape.type %>.new(id: "<%= shape.id %>", traits: <%= shape.traits %>)
+<% end -%>
+
+<% shapes_with_members.each do |shape| -%>
+<% shape.members.each do |member| -%>
+<% case shape.type -%>
```
mullermp: Why not use the type that's already in shapes such as `list` or `map`?

gems/smithy/lib/smithy/anvil/client/templates/shapes.erb
```diff
@@ -0,0 +1,47 @@
...
+module <%= namespace %>
+  module Shapes
+    include Smithy::Client::Shapes
+
+<% shapes.each do |shape| -%>
+    <%= shape.name %> = <%= shape.type %>.new(id: "<%= shape.id %>", traits: <%= shape.traits %>)
+<% end -%>
+
+<% shapes_with_members.each do |shape| -%>
+<% shape.members.each do |member| -%>
```
mullermp, line 15: One improvement you can make here is push this logic to the view class. These are very easy strings you can prepare.

gems/smithy/lib/smithy/anvil/client/templates/shapes.erb
```diff
@@ -0,0 +1,47 @@
...
+    <%= shape.name %>.set_member(<%= member.shape %>, <%= member.traits %>)
+<% when 'MapShape' -%>
+    <%= shape.name %>.set_<%= member.name %>(<%= member.shape %>, <%= member.traits %>)
+<% else -%>
+    <%= shape.name %>.add_member("<%= member.name %>", <%= member.shape %>, <%= member.traits %>)
+<% end -%>
+<% end -%>
+<% if shape.typed -%>
+    <%= shape.name %>.type = Types::<%= shape.name %>
+<% end %>
```
mullermp: Missing trim line (and then add blank space here)

gems/smithy/lib/smithy/anvil/client/views/shapes.rb
```diff
@@ -0,0 +1,254 @@
...
+          def build_member_shape(name, shape, traits)
+            MemberShape.new(
+              name: name.underscore,
+              shape: build_shape_name(shape),
+              traits: filter_traits(traits)
+            )
+          end
+
+          def build_shape_name(id)
+            if PRELUDE_SHAPES_MAP.include?(id)
```
mullermp: Doesn't Vise::Shape.relative_id take care of prelude shapes in the same way? It should just remove the namespace i.e. `smithy.api#BigInteger` should be BigInteger.

gems/smithy/spec/fixtures/shapes/model.smithy
mullermp: Instead of recursion here, we could use the recursive test model that I added and test it in a separate context? The ensures we always cover the case because if removed from the model, this still passes tests. I'm not sure.

projections/weather/lib/weather/shapes.rb
mullermp: Whiteline is fine.

projections/weather/lib/weather/shapes.rb
mullermp: I think single line is fine for generated nonsense especially if this file will be very large.

mullermp (approved):
Nice. Mostly fine for now.

gems/smithy-client/lib/smithy-client/shapes.rb
jterapin: Ahh, good catch. I'll update!

gems/smithy/lib/smithy/anvil/client/views/shapes.rb
jterapin: Ahh yes, that change will work fine. I think previously - I was named all the reference values with Prelude* so I needed the mapping.

projections/weather/lib/weather/shapes.rb
jterapin: I _think_ we should go with single lines - considering we want to reduce the amount of lines in cases like EC2. 

Also - I don't think we look at this file often unless its for debugging purposes (in the context of making this file readable)

gems/smithy/lib/smithy/anvil/client/templates/shapes.erb
jterapin: Yeah I think I figured out a way to push the logic there so the .erb looks cleaner.

gems/smithy/lib/smithy/anvil/client/templates/shapes.erb
jterapin: I'm not entirely sure what you mean by above but assuming that you are referring to the type defined on the shape hash itself. 

I figured it would be easier to just reference the shape class representation here via `.type` since it is set on the Shape class.

gems/smithy/lib/smithy/anvil/client/views/shapes.rb
```diff
@@ -192,6 +184,17 @@ def initialize(options = {})
             end
 
             attr_reader :name, :shape, :traits
+
+            def add_member(shape)
```
mullermp: I think this is a method_call?

mullermp (approved):

alextwoods (approved):

gems/smithy/lib/smithy/anvil/client/views/shapes.rb
jterapin: Yeah, it is a method call - depending on the parent shape that this member shape belongs to - the adding of the member will be different. Open to making this better (and what makes sense).

jterapin: Per our convo, I need a better name here so updated! thanks

PR #222, opened 2024-12-14
Networking plugin

mullermp:
Adds the networking plugin and connection pool. Both are tied to Net::HTTP. Plugin can be replaced by another implementation.


gems/smithy-client/lib/smithy-client/base.rb
```diff
@@ -9,8 +9,8 @@ class Base
       # @api private
       @plugins = PluginList.new(
         [
-          Plugins::Endpoint
-          # Plugins::NetHttp,
+          Plugins::Endpoint,
+          Plugins::NetHTTP
```
alextwoods: Nit: this would also need to be determined by the api.

Also - how are we planning on mixing http1 and http2.  These plugins are at the client level, not at the operation level?

gems/smithy-client/lib/smithy-client/net_http/connection_pool.rb
```diff
@@ -0,0 +1,279 @@
...
+            e.query = nil
+          end.to_s
+        end
+
+        class << self
+          # Returns a connection pool constructed from the given options.
+          # Calling this method twice with the same options will return
+          # the same pool.
+          #
+          # @option options [Numeric] :http_continue_timeout (nil)
```
alextwoods: we probably don't really need all of this documentation here since this is api private? should probably just live in the http.... handler? We don't really have a client class right?

Edit: NVM, I guess maybe this is the best place for this documentation since this class is the one actually using them.  I'm not sure, I find it a little weird to have documented here.

mullermp: It's mostly for us.

gems/smithy-client/lib/smithy-client/base.rb
mullermp: Ah yeah. I think for now we can keep it default for the sake of making it work and remove that bit later.  We need to think about how clients with both http and http 2 operations will work. I think we can register both plugins but the handlers will only register for given operations.

gems/smithy-client/lib/smithy-client/net_http/connection_pool.rb
```diff
@@ -0,0 +1,279 @@
...
+
+module Smithy
+  module Client
+    module NetHTTP
+      # @api private
+      class ConnectionPool
+        @pools_mutex = Mutex.new
+        @pools = {}
+
+        OPTIONS = {
```
jterapin, line 83: I noticed that the `Hearth::HTTP::Client` had some values set. Does any of them apply here? Linked reference here: https://github.com/smithy-lang/smithy-ruby/blob/main/hearth/lib/hearth/http/client.rb#L13

mullermp: I think using the net http defaults is probably preferred. We mostly did that before. I think those values are also the same defaults anyway.

PR #223, opened 2024-12-16
Decaf endpoints

alextwoods:
*Description of changes:*

Changes in `smithy-client`:
* Remove `Endpoints` plugin - this is instead code generated per service.
* Add rules engine stdlib functions. (copied from hearth).
* Add jmespath as a dependency (used by endpoint parameter bindings).

Changes in `smithy`:
* Codegeneration for:
  * endpoint parameters
  * endpoint provider/resolver
  * endpoint plugin
  * endpoint provider specs (translate endpoint test cases).
* extension points for register endpoint built in bindings and stdlib functions.
* default weld that adds default endpoint rules/tests.
* All Endpoint2.0 test models.
* Task to generate sdk per test model + run specs (added to CI).

Not part of this PR (future work):
* OperationInput tests (ie, E2E endpoint tests of the client.  Requires being able to make client operation calls)
* Endpoints + Auth.  (requires us to implement Auth schemes/resolver, ect).


gems/smithy/lib/smithy/anvil/endpoint_parameter.rb
```diff
@@ -0,0 +1,134 @@
+# frozen_string_literal: true
+module Smithy
+  module Anvil
+    # @api private
+    class EndpointParameter
```
mullermp: Probably need to revise the namespace a little bit, but I think this needs to be up one more level. It's not a view but a shared class, only for client? (Or does this apply to servers too?) Possibly our namespace needs to be smithy/anvil/client/views instead of smithy/anvil/views/client.

gems/smithy/lib/smithy/anvil/templates/client/plugins/endpoint.erb
```diff
@@ -0,0 +1,33 @@
...
+
+module <%= namespace %>
+  module Plugins
+    class Endpoint < Smithy::Client::Plugin
+      option(
+        :endpoint_provider,
+        doc_type: '<%= namespace %>::EndpointProvider',
+        docstring: <<~DOCS) do |_cfg|
+The endpoint provider used to resolve endpoints. Any object that responds to
+`#resolve_endpoint(parameters)` where `parameters` is a Struct similar to
```
mullermp: You can remove the similar to bit

gems/smithy/lib/smithy/anvil/templates/client/plugins/endpoint.erb
```diff
@@ -0,0 +1,33 @@
...
+`Aws::ACM::EndpointParameters`.
+        DOCS
+        EndpointProvider.new
+      end
+
+      <%- parameters.select(&:built_in?).each do |param| -%>
+      <%= param.built_in_binding.render_config(plan) %>
```
mullermp: Wouldn't param already know about plan? You can probably simplify this to just have a method that returns configs to render. In general we can have logic in these but I think it makes sense to minimize that

gems/smithy/lib/smithy/anvil/views/client/endpoint_parameters.rb
```diff
@@ -0,0 +1,34 @@
...
+  module Anvil
+    module Views
+      module Client
+        # @api private
+        class EndpointParameters < View
+          def initialize(plan)
+            @plan = plan
+            @model = plan.model
+            # TODO: I reference the endpoint trait ids in many places.
+            # Should we have a constant somewhere for all of them or is it fine to just use the ids as is?
```
mullermp: I think just the string is fine. Constant doesn't give us anything useful and it's private API.

gems/smithy/lib/smithy/anvil/views/client/module.rb
```diff
@@ -38,7 +38,7 @@ def requires
             if @plan.type == :types
               [:types]
             else
-              [:client, :errors, :types]
+              ['plugins/endpoint', :client, :errors, :types, :endpoint_parameters, :endpoint_provider]
```
mullermp: May as well make this all strings. Can you alphabetize?

gems/smithy/lib/smithy/plan.rb
```diff
@@ -10,6 +10,11 @@ def initialize(model, type, options = {})
       @model = Vise::Model.new(model)
       @type = type
       @options = options
+      # TODO: Where should this happen.  In old V4, these were part of generation context
+      # build maps of id to binding for all available endpoint bindings
+      # TODO: We need to validate somewhere that all of the builtin's used in a ruleset have valid bindings
```
mullermp: What do you mean? Does CLI not do this?

gems/smithy/lib/smithy/plan.rb
```diff
@@ -10,6 +10,11 @@ def initialize(model, type, options = {})
       @model = Vise::Model.new(model)
       @type = type
       @options = options
+      # TODO: Where should this happen.  In old V4, these were part of generation context
```
mullermp: Why not in the endpoints view?

gems/smithy/lib/smithy/tools/underscore.rb
```diff
@@ -20,6 +20,14 @@ def underscore(string)
           new_string.downcase!
           new_string
         end
+
+        # @param [String<JMESPath>]
+        # @return [String]
+        def underscore_jmespath(expression)
```
mullermp, line 10: Is this taken from somewhere?

gems/smithy/lib/smithy/vise/endpoints.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+require_relative 'endpoints/built_in_binding'
+require_relative 'endpoints/function_binding'
+
+module Smithy
+  module Vise
+    # Endpoint Rules related metadata
+    # TODO: Where does this actually belong.
+    # it could be in vise because its related to understanding the model
```
mullermp: I think vise makes sense if you're just using it to parse endpoint rules into objects/hash/lists to make it easier for views to render.

gems/smithy/lib/smithy/vise/endpoints/built_in_binding.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Vise
+    module Endpoints
+
+      # Endpoint BuiltInBinding
+      class BuiltInBinding
+        def initialize(id:, render_config:, render_build:, render_test_set:)
```
mullermp: I've just been using options = {} everywhere because of how easy it is to just pass options down the line without extracting keyword args... Not sure how you feel about that. I think I lean towards keyword args for required public interfaces.

gems/smithy/lib/smithy/welds/endpoints.rb
```diff
@@ -0,0 +1,106 @@
...
+module Smithy
+  module Welds
+
+    # Provides default endpoint builtin/function bindings.
+    # TODO : There are many ways we could "register" the base (ie, non-AWS) endpoint builtins/functions
+    # I like this since it uses the same extension point that AWS will and makes it easy to remove/replace.
+    class Endpoints < Weld
+
+      def self.preprocess(model)
+        # TODO: There are quite a few places we could do this
```
mullermp: Makes sense

gems/smithy/lib/smithy/welds/endpoints.rb
```diff
@@ -0,0 +1,106 @@
...
+          unless shape['traits']['smithy.rules#endpointRuleSet']
+            add_default_endpoints(shape['traits'])
+          end
+        end
+      end
+      def self.built_in_bindings
+        [Vise::Endpoints::BuiltInBinding.new(
+          id: 'SDK::Endpoint',
+          render_config: proc do |_plan|
+            <<-ADD_OPTION
```
mullermp: I've been using ~ with indentation. I think you can make this a method that takes option values. Custom options come from the endpoint params but static values for endpoint.

gems/smithy/lib/smithy/welds/endpoints.rb
```diff
@@ -0,0 +1,106 @@
...
+          render_test_set: proc do |_plan, _operation, _node|
+
+          end
+        )]
+      end
+
+      def self.function_bindings
+        # TODO: Add all stdlib from:
+        # https://github.com/smithy-lang/smithy-ruby/blob/main/codegen/smithy-ruby-codegen/src/main/java/software/amazon/smithy/ruby/codegen/rulesengine/FunctionBinding.java#L49
+        [Vise::Endpoints::FunctionBinding.new(
```
mullermp: I think these classes are basically views in anvil are they not? Vise would be for parsing the model

gems/smithy/lib/smithy/welds/endpoints.rb
```diff
@@ -0,0 +1,106 @@
...
+          ruby_method: 'Smithy::Client::EndpointRules::valid_host_label?'
+        )]
+      end
+
+      def self.add_default_endpoints(service_traits)
+        service_traits['smithy.rules#endpointRuleSet'] = default_endpoint_rules
+        service_traits['smithy.rules#smithy.rules#endpointTests'] = default_endpoint_tests
+      end
+
+      def self.default_endpoint_rules
```
mullermp: Should we just make these json files that we read into code or copy over?

gems/smithy-client/lib/smithy-client/endpoint_rules.rb
```diff
@@ -0,0 +1,152 @@
...
+
+          @authority = _authority(url, uri)
+          @path = uri.path
+          @normalized_path = uri.path + (uri.path[-1] == '/' ? '' : '/')
+          @is_ip = _is_ip(uri.host)
+        end
+
+        attr_reader :scheme, :authority, :path, :normalized_path, :is_ip
+
+        def as_json(_options = {})
```
jterapin: Curious to know why `_options = {}` is necessary than omitting it. I'm guessing it is due to the way this class is used as a "wrapper" for URI?

gems/smithy/lib/smithy/anvil/templates/client/plugins/endpoint.erb
alextwoods: Param knows about plan but the built in binding does not and needs to be called with it.  I could make a convience method on the `EndpointParameter` view helper for `render_builtin_binding_config` which wouldn't need to pass plan (since the parameter knows it), but that doesn't seem much better.

The reason I went with a generic block was to keep things simple - I didn't want to model a codegen config the way we do in smithy-ruby java.  The configs can be fairly complex (eg, they need some arbitrary default or resolution logic).

gems/smithy-client/lib/smithy-client/endpoint_rules.rb
```diff
@@ -0,0 +1,152 @@
...
+            'authority' => authority,
+            'path' => path,
+            'normalizedPath' => normalized_path,
+            'isIp' => is_ip
+          }
+        end
+
+        private
+
+        def _authority(url, uri)
```
jterapin: From what I understand, the starting underscore within a method indicates a private method and this helps with readability. Should we follow this format across the codebase?

gems/smithy-client/lib/smithy-client/endpoint_rules.rb
```diff
@@ -0,0 +1,152 @@
...
+          # don't include port if it's default and not parsed originally
+          if uri.default_port == uri.port && !url.include?(":#{uri.port}")
+            uri.host
+          else
+            "#{uri.host}:#{uri.port}"
+          end
+        end
+
+        def _is_ip(authority)
+          IPAddr.new(authority)
```
jterapin, line 165: Ahh, this is how we are validating IP addresses?  Otherwise, this will raise an `IPAddr::InvalidAddressError` and returns false. Neat!

gems/smithy/lib/smithy/anvil/endpoint_parameter.rb
jterapin: Would it be weird to have a directory that is dedicated to helpers that is shared between several views?

I would say some views are not quite `client` specific since we forges some `client` views for types.
Is it necessary to have namespaces specific to whether something is client or server? 

Thinking of ways to flatten things here.

gems/smithy/lib/smithy/anvil/endpoint_parameter.rb
```diff
@@ -0,0 +1,134 @@
...
+
+      private
+
+      # Highest to lowest priority:
+      # staticContextParams
+      # contextParam
+      # operationContextParams
+      # clientContextParams (always sourced from config)
+      # Built-In Bindings (sourced from config in most cases, context in some cases to allow operation level overrides)
+      # Built-in binding default values
```
jterapin: Bullet points or numbered will be good for readability.

gems/smithy/lib/smithy/anvil/templates/client/plugins/endpoint.erb
```diff
@@ -0,0 +1,33 @@
...
+module <%= namespace %>
+  module Plugins
+    class Endpoint < Smithy::Client::Plugin
+      option(
+        :endpoint_provider,
+        doc_type: '<%= namespace %>::EndpointProvider',
+        docstring: <<~DOCS) do |_cfg|
+The endpoint provider used to resolve endpoints. Any object that responds to
+`#resolve_endpoint(parameters)` where `parameters` is a Struct similar to
+`Aws::ACM::EndpointParameters`.
```
jterapin: This plugin lives in generic side of smithy - should we have this specific reference to an AWS service?

gems/smithy/lib/smithy/anvil/views/client/endpoint_parameters.rb
jterapin: If there is a file that contains shared logic between endpoints views, maybe constants there is preferred to keep all references in a single place.

gems/smithy/lib/smithy/anvil/views/client/plugins/endpoint.rb
```diff
@@ -0,0 +1,38 @@
...
+            def namespace
+              Tools::Namespace.namespace_from_gem_name(@plan.options[:gem_name])
+            end
+
+            def documentation
+              "# TODO: Documentation"
+            end
+
+            def parameters
+              @paramters ||= @endpoint_rules.data['parameters']
```
jterapin: Typo here: parameters.

gems/smithy/lib/smithy/tools/underscore.rb
jterapin: Thinking from V3 codegen: https://github.com/aws/aws-sdk-ruby/blob/version-3/build_tools/aws-sdk-code-generator/lib/aws-sdk-code-generator/underscore.rb#L23-L29

jterapin (commented):
May want to run rubocop to fix some indentations.

gems/smithy/lib/smithy/vise/endpoints/built_in_binding.rb
alextwoods: Yeah - I think for required public interfaces I'd generally prefer keyword args.  Since this is an extension point/public interface, I think keeping them as keywords make sense.

gems/smithy/lib/smithy/anvil/views/client/module.rb
alextwoods: Changed to all strings, but the order matters here so I can't alphabetize.  specially, the plugin must come before client.  I chose to do it this way to keep it simple.  The alternative is to come up with a codegen time dependency modeling.

gems/smithy/lib/smithy/welds/endpoints.rb
alextwoods: I can switch to `<<~` for indentation - I don't have a style preference on those.

What do you mean about having this method take option values.  I'm not quite sure I understand.

In general, this needs to be fairly flexible to support all of our existing/future built ins (eg: account id).

gems/smithy/lib/smithy/welds/endpoints.rb
alextwoods: What do you mean?  I'm not sure they really belong as views.   They're used on plan and I think thats probably where they belong (Since they relate to what is the broader generation context).

gems/smithy/lib/smithy/welds/endpoints.rb
alextwoods: Yeah, we can make them json files, but then those files need to be included in the gem.  They're fairly short, so I leaned towards keeping it simple here and keeping them in code.

gems/smithy-client/lib/smithy-client/endpoint_rules.rb
alextwoods: good question, I guess we could use `*` or `...` instead to indicate it can take arguments, but doesn't use them.

Edit, I've updated to use just `*`.

gems/smithy/lib/smithy/anvil/templates/client/plugins/endpoint.erb
alextwoods: Yeah good call, I removed that.  I

gems/smithy/lib/smithy/anvil/views/client/plugins/endpoint.rb
alextwoods: Good call, fixed in all the files.

gems/smithy-client/lib/smithy-client/endpoint_rules.rb
alextwoods: Yeah - I had just copied this over.  I think in our newer style we shouldn't really need to prefix with "_", so I'll remove that here. 

EDIT:  nvm, I see why this was done.  We have `attr_reader` for `authority` and `is_ip` already, so this avoids the conflict without requiring us to write more verbose method names like `extract_authority`.  I think reasonable to leave this, but I think our general style guide should be to not use "_" to just denote private.

gems/smithy/lib/smithy/anvil/endpoint_parameter.rb
alextwoods: Yeah - I think in theory this helper is probably specific to client.

Maybe inverting the namespace to `smithy/anvil/client/views` makes sense.  I won't make that change as part of this PR though (to avoid conflicts for everyone since this isn't ready to merge)

.github/workflows/ci.yml
```diff
@@ -43,6 +43,21 @@ jobs:
       - name: Test
         run: bundle exec rake smithy-client:spec
 
+  smithy-test-endpoints:
```
mullermp: It would be preferable to run these tests as part of smithy spec instead of making an entirely new task and CI.

Rakefile
```diff
@@ -11,6 +11,29 @@ namespace :smithy do
...
+    require 'tmpdir'
+
+    Dir.glob('gems/smithy/spec/fixtures/endpoints/*/model.json') do |model_path|
+      Dir.mktmpdir do |tmp_dir|
+        test_name = model_path.split('/')[-2]
+        gem_name = test_name.gsub('-', '')
+        puts "Building SDK for #{test_name} at: #{model_path}"
+        model = JSON.load_file(model_path)
+
+        plan = Smithy::Plan.new(model, :client, {
```
mullermp: Formatting

gems/smithy-client/lib/smithy-client/endpoint_rules.rb
```diff
@@ -0,0 +1,188 @@
...
+        # @return [Hash]
+        attr_accessor :headers
+      end
+
+      # Evaluates whether the input string is a compliant RFC 1123 host segment.
+      # When allowSubDomains is true, evaluates whether the input string is
+      # composed of values that are each compliant RFC 1123 host segments
+      # joined by dot (.) characters.
+      # @api private
+      # rubocop:disable Style/OptionalBooleanParameter
```
mullermp: I don't remember if we were able to fix this, but we should just solve this one if possible.

gems/smithy-client/lib/smithy-client/endpoint_rules.rb
mullermp: We should just change the method names.

gems/smithy-client/spec/smithy-client/endpoint_rules_spec.rb
```diff
@@ -0,0 +1,202 @@
...
+
+        it 'returns true for a single valid character' do
+          expect(EndpointRules.valid_host_label?('h')).to be true
+        end
+
+        it 'returns false for 64 characters' do
+          expect(EndpointRules.valid_host_label?('h' * 64)).to be false
+        end
+
+        context 'allow_sub_domains=false' do
```
mullermp, line 39: Wonky formatting on all of these tests.

gems/smithy-client/spec/smithy-client/endpoint_rules_spec.rb
```diff
@@ -0,0 +1,202 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    describe EndpointRules do
+      describe '.valid_host_label?' do
```
mullermp, line 6: Do we not cover these tests from standard endpoint test rules? I thought most of these are defined in a json somewhere.

gems/smithy/lib/smithy/anvil/client/templates/endpoint_parameters.erb
```diff
@@ -0,0 +1,64 @@
...
+    def initialize(options = {})
+    <%- parameters.each do |param| -%>
+      self[:<%= param.name %>] = options.fetch(:<%= param.name %>, <%= param.default_value %>)
+    <%- end -%>
+    <%- parameters.select(&:validate_required?).each do |param| -%>
+      raise ArgumentError, "Missing required EndpointParameter: :<%= param.name %>" if <%= param.name %>.nil?
+    <%- end -%>
+    end
+
+    def self.create(config, operation_name=nil, params={})
```
mullermp: Formatting in method signature and case statement below.

gems/smithy/lib/smithy/anvil/client/templates/endpoint_parameters.erb
```diff
@@ -0,0 +1,64 @@
...
+      <%- if operation_params.empty? -%>
+      new({
+      <%- parameters.select(&:create_from_config?).each do |param| -%>
+        <%= param.name %>: <%= param.value %>,
+      <%- end -%>
+      }.compact)
```
mullermp, line 31: Is compact necessary? You are selecting values that should not be nil.

gems/smithy/lib/smithy/anvil/client/templates/plugins/endpoint.erb
```diff
@@ -0,0 +1,61 @@
...
+# This is generated code!
+
+module <%= namespace %>
+  module Plugins
+    class Endpoint < Smithy::Client::Plugin
+      option(
+        :endpoint_provider,
+        doc_type: '<%= namespace %>::EndpointProvider',
+        docstring: <<~DOCS) do |_cfg|
+The endpoint provider used to resolve endpoints. Any object that responds to
```
mullermp: Should be indented with tilde heredoc.

gems/smithy/lib/smithy/anvil/client/templates/plugins/endpoint.erb
```diff
@@ -0,0 +1,61 @@
...
+`#resolve_endpoint(parameters)`.
+        DOCS
+        EndpointProvider.new
```
mullermp, line 17: This could just be passed as `default:` instead of as the block?

gems/smithy/lib/smithy/anvil/client/templates/specs/endpoint_provider_spec.erb
```diff
@@ -0,0 +1,33 @@
...
+
+# This is generated code!
+
+require_relative 'spec_helper'
+
+module <%= namespace %>
+  describe EndpointProvider do
+    subject { EndpointProvider.new }
+
+    <% test_cases.each do |test_case| %>
```
mullermp: Not apparent unless you inspect the generated code, but all ERB tags should have dashes <%- or -%> because of the trim mode that is selected in the base view class.

gems/smithy/lib/smithy/anvil/client/views/endpoint_provider.rb
```diff
@@ -0,0 +1,241 @@
...
+module Smithy
+  module Anvil
+    module Client
+      module Views
+        # @api private
+        class EndpointProvider < View
+          def initialize(plan)
+            @plan = plan
+            @model = plan.model
+            service = Vise::ServiceIndex.new(@model).service.values.first
```
mullermp: Outside scope, I am thinking we can put a resolved service id on the plan, so fetching the service shape would be to index into shapes with it. There wouldn't be an index I don't think. That would also force validation instead of first index usage.

gems/smithy/lib/smithy/anvil/client/views/module.rb
```diff
@@ -38,7 +38,8 @@ def requires
             if @plan.type == :types
               [:types]
             else
-              %i[client errors types]
+              # Order matters here - plugins must come before client
```
mullermp, line 41: This is fine. Perhaps we should break this into two methods, plugins and components. This list will get long.

gems/smithy/lib/smithy/anvil/client/views/plugins/endpoint.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Anvil
+    module Client
+      module Views
```
mullermp, line 6: We should think about using a compact namespace Anvil::Client::Views or just dropping Views (and removing views folder) as Anvil is already the designated namespace for code writing.

gems/smithy/lib/smithy/plan.rb
```diff
@@ -31,5 +33,19 @@ def initialize(model, type, options = {})
 
     # @return [Array<Polish>] The polishes that apply to this plan.
     attr_reader :polishes
+
+    # @return [Hash[String, BuiltInBinding]] Array of all registered builtins
```
mullermp: Hash<string, val>

gems/smithy/lib/smithy/plan.rb
```diff
@@ -31,5 +33,19 @@ def initialize(model, type, options = {})
...
+    attr_reader :built_in_bindings
+
+    # @return [Hash[String, FunctionBinding]] Array of all registered functions
+    attr_reader :function_bindings
```
mullermp: These are not clear that they are endpoint related. I also don't think these should be on plan. Could a view not call the plans welds (that's why weld is an instance variable on plan?) and initialize all of the bindings for the view to use?

gems/smithy/lib/smithy/vise/endpoints/built_in_binding.rb
mullermp: I don't think this is public at all. Please just use options for vise usage. But I think these are better used as view classes now - since vise does not have any shape wrapping.

gems/smithy/lib/smithy/vise/endpoints.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+require_relative 'endpoints/built_in_binding'
+require_relative 'endpoints/function_binding'
+
+module Smithy
+  module Vise
```
mullermp: I don't think vise changes are necessary anymore. I think these binding classes probably are view classes.

gems/smithy/lib/smithy/weld.rb
```diff
@@ -20,5 +20,17 @@ def for?(_model)
     def preprocess(_model)
       # no-op
     end
+
+    # Called when constructing a plan
+    # @return [BuiltInBinding] list of built in bindings for use in endpoint rules.
+    def built_in_bindings
```
mullermp: Theses should probably be "endpoint_built_ins" and "endpoint_functions" with or without binding suffix.

gems/smithy/lib/smithy/welds/endpoints.rb
mullermp: This still needs to be indented. What I was saying before was to have a method that is responsible for rending the config, but this is fine to hardcode endpoint for now. Eventually this will need to be a method.

gems/smithy/lib/smithy/welds/endpoints.rb
```diff
@@ -0,0 +1,134 @@
...
+          end,
+          render_test_set: proc do |_plan, _operation, _node|
+          end
+        )]
+      end
+
+      # rubocop:disable Metrics/MethodLength
+      def function_bindings
+        [
+          Vise::Endpoints::FunctionBinding.new(
```
mullermp: Similar to the vise comments before, I think these should really just be a collection of hashes/lists from the model (it could really just be the same format as the model, too), and then passed to the view, which may wrap those to help with rendering.

gems/smithy/lib/smithy/welds/endpoints.rb
```diff
@@ -0,0 +1,134 @@
...
+      private
+
+      def add_default_endpoints(service_traits)
+        service_traits['smithy.rules#endpointRuleSet'] = default_endpoint_rules
+        service_traits['smithy.rules#endpointTests'] = default_endpoint_tests
+      end
+
+      # rubocop:disable Metrics/MethodLength:
+      def default_endpoint_rules
+        JSON.parse(<<~JSON)
```
mullermp: I still think it makes sense to make this a JSON file (it gives linting and what not) and have the smithy gem include that in the lib/ path. Default test cases will probably get larger too.

gems/smithy/spec/fixtures/endpoints/endpoints-string-array/model.json
```diff
@@ -0,0 +1,319 @@
+{
```
mullermp, line 1: Are there .smithy file equivalents we should copy in too?

gems/smithy/spec/interfaces/client/endpoint_provider_spec.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+describe 'Component: EndpointProvider' do
+  before(:all) do
+    @tmpdir = SpecHelper.generate(['Defaults'], :client, fixture: 'endpoints/default-values')
```
mullermp: Naming nit, EndpointDefaults (as opposed to model defaults which we may have to test later).

alextwoods: The name here shouldn't matter right? It gets cleaned up so shouldn't be a conflict?

gems/smithy/spec/fixtures/endpoints/endpoints-string-array/model.json
alextwoods: I guess - they're in AwsDrSeps.  I'm not sure I like this mixed model.json/model.smithy we have going for our test fixtures, though I'm not quite sure what we should do.

.github/workflows/ci.yml
alextwoods: Discussed offline - I'd prefer they are separate and its simpler to keep them separate for now.  We can update later.

gems/smithy-client/lib/smithy-client/endpoint_rules.rb
alextwoods: I removed the default value for this.  It makes testing a little easier but all usage of it in endpoint rules requires the parameter, so it isn't needed.

gems/smithy-client/lib/smithy-client/endpoint_rules.rb
alextwoods: What name would you prefer - "build_authority" or "extract_authority"?

gems/smithy-client/spec/smithy-client/endpoint_rules_spec.rb
alextwoods: These do end up getting tested through endpoint rules - but to have coverage in this library, we need the specs (and we had them in hearth, I just copied them over).

gems/smithy-client/spec/smithy-client/endpoint_rules_spec.rb
alextwoods: yea, rubocop -a did something weird to these :-(

gems/smithy/lib/smithy/anvil/client/templates/endpoint_parameters.erb
alextwoods: They can be nil here and if so, they should get default values (if applicable) so we do need the compact.

gems/smithy/lib/smithy/anvil/client/templates/plugins/endpoint.erb
alextwoods: I think that will create one object and not a new one each time?

gems/smithy/lib/smithy/anvil/client/views/module.rb
alextwoods: yeah that makes sense.  I'll leave to a future refactoring outside of this PR.

gems/smithy/lib/smithy/anvil/client/views/plugins/endpoint.rb
alextwoods: Yeah, I agree this namespace is getting long.  I can do that refactor outside of this PR.

gems/smithy/lib/smithy/plan.rb
alextwoods: Yeah - I guess we could push that to views to do when needed.  The issue I ran into is that each instance of EndpointParameter needs the built in bindings, which means I either need to call the weld many times (once per parameter x operations) or build it in the view(s) and pass it around.  It seemed like something that should only get done once and stored.

gems/smithy/lib/smithy/vise/endpoints.rb
alextwoods: They don't really feel like view classes to me - at least in a traditional MVC sort of breakdown, I'd consider these more "model" related, which is why I had put them into vice.  

Additionally these are public extension points, so don't feel like they belong in views. 

Maybe they belong in the weld namespace?

gems/smithy/lib/smithy/vise/endpoints/built_in_binding.rb
alextwoods: These must be public - these are the extension point for users to add endpoint related bindings.

gems/smithy/spec/fixtures/endpoints/endpoints-string-array/model.json
mullermp: I find it easier to edit the model files in smithy IDL then generate the json with it, that's why both are there. We could require smithy cli and shell out to convert to AST and not have the json, but also having the json inspectable (in the folder) helps with writing generation.

gems/smithy/spec/interfaces/client/endpoint_provider_spec.rb
mullermp: It does get cleaned up. It was mostly just a naming of what the service is - it's endpoints related. I don't feel strongly.

gems/smithy/lib/smithy/welds/endpoints.rb
alextwoods: This is the correct indenting for it to render correctly in the generated code.

Why should this be a method?  I think I'm still a little confused about that. Using a proc/callable is what lets a user define arbitrary configuration(s) that may apply for the builtin.  I think thats perferable to coming up with a codegen time modeling of config.

gems/smithy/lib/smithy/welds/endpoints.rb
alextwoods: I'd disagree since these are the public extension point - I think having a class, even if its a simple one, help define that interface.

gems/smithy-client/lib/smithy-client/endpoint_rules.rb
mullermp: Yeah, build or parse authority, or just authority(thing) is fine, as that's different than standard getters/setters.

.github/workflows/ci.yml
mullermp: Pushed a simplification, feel free to take or not.

alextwoods: Thanks - simplification looks good.

gems/smithy/lib/smithy/welds/endpoints.rb
alextwoods: Moved to json files.

gems/smithy/lib/smithy/welds/endpoints.rb
mullermp: Maybe these should be in Smithy::Endpoints and not vise. Vise doesn't seem correct. But I also don't think a class is necessary.

gems/smithy-client/spec/smithy-client/endpoint_rules_spec.rb
mullermp: That's strange - are we sure they were even tested through endpoint rules?

gems/smithy/lib/smithy/anvil/client/templates/endpoint_parameters.erb
mullermp: Could we just ignore nil values in our select during generation rather than at runtime?

gems/smithy/lib/smithy/anvil/client/templates/plugins/endpoint.erb
mullermp: What do you mean? The block or default in config options is run once in client initialize

gems/smithy/lib/smithy/plan.rb
mullermp: I think it's the view's responsibility to get what it needs (bindings) from welds or whatever sources, and pass it to any sub views if it has to.

gems/smithy-client/spec/smithy-client/endpoint_rules_spec.rb
alextwoods: Some of them are better tested than others through test rule files.  Its one of my long term goals (when I get extra time...) to really improve those shared endpoint test cases.

gems/smithy/lib/smithy/vise/endpoints.rb
mullermp: What would it look like just as an array of hash data? It doesn't look like we need objects at all. Can we just start simple?

gems/smithy/lib/smithy/anvil/client/templates/endpoint_parameters.erb
alextwoods: I dont think so - we don't know at codegen time what runtime values users are going to use?

gems/smithy/lib/smithy/anvil/client/templates/plugins/endpoint.erb
alextwoods: The default (if its not a block) is only ever run once (ie, at the time the file is required), which means multiple instances of a class will share the same object here, so it must be a block.

gems/smithy/lib/smithy/vise/endpoints.rb
alextwoods: I think that makes this much messier as an extension point - having a well defined class makes it easy for users to extend.

mullermp: In practicality we will be the only ones adding endpoint bindings let's be real..

gems/smithy/lib/smithy/anvil/client/templates/plugins/endpoint.erb
mullermp: Ah ok

gems/smithy/lib/smithy/anvil/client/templates/endpoint_parameters.erb
mullermp: Makes sense.

gems/smithy/lib/smithy/vise/endpoints.rb
alextwoods: I do generally prefer the classes, but I've removed them in the latest commit and removed the bindings from the plan.

gems/smithy/lib/smithy/vise/endpoints/built_in_binding.rb
alextwoods: I've removed the classes and just using arrays/hashes.

Rakefile
```diff
@@ -5,13 +5,42 @@ require 'rspec/core/rake_task'
...
   end
+
+  task 'spec:endpoints' do
+    require_relative 'gems/smithy/spec/spec_helper'
+
+    spec_paths = []
+    include_paths = []
+    tmp_dirs = []
+    Dir.glob('gems/smithy/spec/fixtures/endpoints/*/model.json') do |model_path|
+      test_name = model_path.split('/')[-2]
```
mullermp, line 23: Is there a better way to express this one?

gems/smithy-client/lib/smithy-client.rb
```diff
@@ -1,18 +1,20 @@
 # frozen_string_literal: true
 
+require 'jmespath'
+
 # client
 
 require_relative 'smithy-client/block_io'
 require_relative 'smithy-client/configuration'
 require_relative 'smithy-client/handler'
+require_relative 'smithy-client/handler_context'
```
mullermp: Alphabetize under handler builder

gems/smithy-client/lib/smithy-client/base.rb
```diff
@@ -9,7 +9,7 @@ class Base
       # @api private
       @plugins = PluginList.new(
         [
-          Plugins::Endpoint,
+          # Plugins::Endpoint,
```
mullermp: You can delete this (and in the test file)

gems/smithy-client/spec/smithy-client/endpoint_rules_spec.rb
```diff
@@ -0,0 +1,192 @@
...
+          it 'returns false for sub domains starting with a dash' do
+            expect(EndpointRules.valid_host_label?('part1.-part2', true)).to be false
+          end
+        end
+      end
+
+      describe '.parse_url' do
+        it 'parses a valid url' do
+          expect(EndpointRules.parse_url('https://example.com'))
+            .to eq({
```
mullermp: Formatting is still wonky on these. I think if you remove the curly braces, rubocop does a better job. Or, do an expected = { hash with formatting } and eq(hash)

gems/smithy/lib/smithy/anvil/client/endpoint_parameter.rb
```diff
@@ -0,0 +1,144 @@
...
+        def context_param_value(operation)
+          return nil unless operation['input']
+
+          input_shape = @model['shapes'].fetch(operation['input']['target'], {})
+          members = input_shape.fetch('members', {})
+          context_param_member(members)
+        end
+
+        def context_param_member(members)
+          members.detect do |(member_name, member_def)|
```
mullermp: Use find instead of detect (they are aliases anyway)

gems/smithy/lib/smithy/anvil/client/templates/endpoint_parameters.erb
```diff
@@ -0,0 +1,65 @@
...
+    <%- parameters.each do |param| -%>
+    :<%= param.name %>,
+    <%- end -%>
+    keyword_init: true
+  ) do
+    include Smithy::Client::Structure
+
+    # @api private
+    class << self
+      PARAM_MAP = {
```
mullermp: Is PARAM_MAP still necessary for this to function or is there an alternative?

gems/smithy/lib/smithy/anvil/client/views/endpoint_provider.rb
```diff
@@ -0,0 +1,244 @@
...
+module Smithy
+  module Anvil
+    module Client
+      module Views
+        # @api private
+        class EndpointProvider < View
+          def initialize(plan)
+            @plan = plan
+            @model = plan.model
+            service = @plan.service.values.first
```
mullermp, line 14: I know this is a sharp edge. One other way is `_, service = @plan.service.first`

gems/smithy/lib/smithy/forge/client.rb
```diff
@@ -24,6 +24,7 @@ def forge
 
       private
 
+      # rubocop:disable Metrics/AbcSize
```
mullermp, line 27: I added the methods here to solve this but I guess it was just deferred. Is there a way to solve this one? Maybe if enumerator is passed to a method and that method does the yielding? Otherwise I guess it's fine to just ignore this one as it's a giant list.

gems/smithy/lib/smithy/forge/client.rb
```diff
@@ -32,8 +33,15 @@ def source_files
           e.yield "lib/#{@gem_name}/customizations.rb", render_customizations
           e.yield "lib/#{@gem_name}/types.rb", render_types
           e.yield "lib/#{@gem_name}/errors.rb", render_errors
+          e.yield "lib/#{@gem_name}/endpoint_parameters.rb", render_endpoint_parameters
+          e.yield "lib/#{@gem_name}/endpoint_provider.rb", render_endpoint_provider
+          e.yield "lib/#{@gem_name}/plugins/endpoint.rb", render_endpoint_plugin
+
+          e.yield 'spec/spec_helper.rb', render_spec_helper
+          e.yield 'spec/endpoint_provider_spec.rb', render_endpoint_provider_spec
```
mullermp: I think by convention this should still be spec/gemname/specs

gems/smithy/lib/smithy/welds/endpoints.rb
```diff
@@ -0,0 +1,65 @@
...
+      def preprocess(model)
+        model['shapes'].select { |_k, s| s['type'] == 'service' }.each_value do |shape|
+          add_default_endpoints(shape['traits']) unless shape['traits']['smithy.rules#endpointRuleSet']
+        end
+      end
+
+      def endpoint_built_in_bindings
+        {
+          'SDK::Endpoint' => {
+            # Text indenting is used in generated view.
```
mullermp, line 16: It would be better if indenting was relative instead of absolute here. So, using `<<~ADD_OPTION` with one level of indenting, then in our view, apply the current indenting level plus this one level.

gems/smithy/spec/fixtures/endpoints/endpoint-bindings.smithy
```diff
@@ -0,0 +1,346 @@
+$version: "2.0"
```
mullermp: I think this file is added in by mistake? You already have model.smithy under endpoint-bindings

smithy-build.json
```diff
@@ -11,6 +11,9 @@
                 }
             ],
             "plugins": {
+                "model::withPrelude": {
```
mullermp: Mistake?

mullermp (approved):
Nice!

gems/smithy-client/lib/smithy-client.rb
alextwoods: ah whoops, I think this came from bad merges.  So many merge conflicts over time in this pr, haha.

Rakefile
alextwoods: Do you mean the `model_path.split('/')[-2]` part? I could do `model_path.split('/').last(2).first` instead, but that felt worse.  Alternatively we coudl use regex to extract it, but I think splitting on / is simpler here.

gems/smithy/lib/smithy/forge/client.rb
alextwoods: Yeah, its just a list so I don't think rubucop is useful for that...  We could solve it by breaking it into even more methods, but I dont think that really helps readability.

gems/smithy/lib/smithy/anvil/client/templates/endpoint_parameters.erb
alextwoods: Its not currently used so I've removed it. I think it was legacy from V3.  I don't actually see any usage anymore.

smithy-build.json
alextwoods: Yup, from testing.  Removed.

gems/smithy/lib/smithy/forge/client.rb
alextwoods: Was following what we had in V3, but I can update.

gems/smithy/lib/smithy/welds/endpoints.rb
alextwoods: I'm struggling with this because of the multiline text.  Is there a way to have erb views indent multiline text?

Rakefile
mullermp: Was thinking in terms of readability. What exactly is extracted? It's otherwise fine.

gems/smithy/lib/smithy/forge/client.rb
mullermp: Yeah. In that case the multimethods are kinda unnecessary.

PR #224, opened 2024-12-18
Refactor polish and weld

mullermp:
Pulls model out of plan, so that polish and welding can accept plan, and at various steps, would not have incomplete models or mismatched types.

### DIFF
--- a/gems/smithy/lib/smithy.rb
+++ b/gems/smithy/lib/smithy.rb
@@ -10,7 +10,9 @@ require_relative 'smithy/forge'
 require_relative 'smithy/vise'
 require_relative 'smithy/plan'
 require_relative 'smithy/polish'
+require_relative 'smithy/polishes'
 require_relative 'smithy/weld'
+require_relative 'smithy/welds'
 
 # Smithy is a modeling language created by AWS. This gem provides a Ruby
 # implementation of the Smithy language and generates clients, servers, and types.
@@ -19,10 +21,14 @@ module Smithy
 
   class << self
     # Generates a Ruby artifact from a Smithy model.
-    # @param plan [Smithy::Plan] the plan to generate the artifact from
-    def smith(plan)
-      artifact = Smithy::Forge.forge(plan)
-      Smithy::Polish.descendants.each { |p| p.new(plan).polish(artifact) }
+    # @param [Hash] model The Smithy model to generate the artifact from.
+    # @param [Plan] plan The plan to generate the artifact from.
+    def smith(model, plan)
+      Welds.load!(plan)
+      Polishes.load!(plan)
+
+      artifact = Smithy::Forge.forge(model, plan)
+      Polishes.for(model).each { |polish| polish.polish(artifact) }
       artifact
     end
   end
--- a/gems/smithy/lib/smithy.rb
+++ b/gems/smithy/lib/smithy.rb
@@ -21,14 +21,10 @@ module Smithy
 
   class << self
     # Generates a Ruby artifact from a Smithy model.
-    # @param [Hash] model The Smithy model to generate the artifact from.
     # @param [Plan] plan The plan to generate the artifact from.
-    def smith(model, plan)
-      Welds.load!(plan)
-      Polishes.load!(plan)
-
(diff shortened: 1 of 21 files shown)


gems/smithy/lib/smithy/plan.rb
```diff
@@ -3,18 +3,13 @@
 module Smithy
   # The Plan class is a simple data structure that holds the model, type, and options for a generator.
   class Plan
-    # @param model [Hash] The API model as a JSON hash.
-    # @param type [Symbol] The type of code to generate, either :client, :server, or :types.
-    # @param options [Hash] The options passed to the generator.
-    def initialize(model, type, options = {})
-      @model = Vise::Model.new(model)
```
alextwoods, line 10: I still feel like model is central to plan and should be part of it.  It seems clunky to pass model and plan to every view for example.

mullermp: Yeah, I tried to make it work with model inside of plan, but the major issue I had was that model inside of plan could differ based on the welding step. For example in preprocess it was a hash and any other step it would have been a vise model.

mullermp: I also removed welding hooks out of initializer methods which I think is code smell.

gems/smithy/lib/smithy/plan.rb
```diff
@@ -3,22 +3,35 @@
...
       @type = type
       @options = options
+
+      Welds.load!(self)
+      @welds = Welds.for(model)
+      Polishes.load!(self)
+      @polishes = Polishes.for(model)
+
+      @welds.each { |weld| weld.preprocess(model) }
+      @model = Vise::Model.new(model)
```
alextwoods, line 19: Nice! I like this.

gems/smithy/lib/smithy/plan.rb
```diff
@@ -3,22 +3,35 @@
...
+    # @return [Vise::Model] The API model wrapped by Vise.
     attr_reader :model
 
     # @return [Symbol] The type of code to generate.
     attr_reader :type
 
     # @return [Hash] The options passed to the generator.
     attr_reader :options
+
+    # @return [Array<Weld>] The welds that apply to this model.
```
alextwoods: nit: apply to this plan.  (since in theory they may apply only for client or server, ect).

gems/smithy/lib/smithy/polishes.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module Smithy
+  # @api private
+  module Polishes
```
alextwoods, line 5: Naming discussion.  `Polishes` reads a little weirdly to me - it almost feels like this should be `Polishers`.  Ie, a `Polisher` can take the `polish` action, eg: `@polishers.each { |p| p.polish }`.

gems/smithy/lib/smithy.rb
```diff
@@ -19,10 +21,10 @@ module Smithy
 
   class << self
     # Generates a Ruby artifact from a Smithy model.
-    # @param plan [Smithy::Plan] the plan to generate the artifact from
+    # @param [Plan] plan The plan to generate the artifact from.
     def smith(plan)
       artifact = Smithy::Forge.forge(plan)
-      Smithy::Polish.descendants.each { |p| p.new(plan).polish(artifact) }
+      plan.polishes.each { |polish| polish.polish(artifact) }
```
alextwoods, line 27: I'm struggling a little bit with why we have separate Welds and Polishes.  Why are these separate extension points - couldn't "polish" just be a hook on  a weld?

alextwoods (approved):
Nice - I like the approach!

gems/smithy/lib/smithy.rb
mullermp: Polish required instances because it uses Thor actions. Weld now has instances but doesn't include thor. I'm not sure if we should keep the distinction anymore or not.

gems/smithy/lib/smithy/polishes.rb
mullermp: Polishes is the plural of polish. We can think about changing this.

PR #226, opened 2024-12-19
Add top level shapes to vise

mullermp:
Adds top level objects to vise to make parsing much smoother.


gems/smithy/lib/smithy/vise/resource_shape.rb
```diff
@@ -0,0 +1,74 @@
...
+      #   @lifecycle_operations['update']
+      # end
+      #
+      # # @return [Hash<String, String>, nil]
+      # def delete
+      #   @lifecycle_operations['delete']
+      # end
+      #
+      # # @return [Hash<String, String>, nil]
+      # def list
```
mullermp: I can add these back or just delete them

gems/smithy/lib/smithy/anvil/client/views/types.rb
```diff
@@ -19,7 +19,7 @@ def namespace
           def types
             @model
               .shapes
-              .select { |_key, shape| shape.type == 'structure' }
+              .select { |_key, shape| shape.is_a?(Vise::StructureShape) }
```
jterapin, line 22: `Types.rb` is not just limited to holding structure shapes, right? It should also include enums and unions based on the approach we took in smithy-ruby.

gems/smithy/lib/smithy/vise/resource_shape.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Vise
+    # Represents a resource shape from the model.
+    class ResourceShape < Shape
```
jterapin, line 6: I like that we have dedicated classes for `ResourceShape` and such. It makes things easier to read (and properly maintain down the road).

gems/smithy/lib/smithy/vise/service_shape.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Vise
+    # Represents a service shape from the model.
+    class ServiceShape < Shape
+      # (see Shape#initialize)
+      def initialize(id, shape)
+        super
```
jterapin, line 9: I didn't know if there was a standard to have `super` first or after attributes. 
I think I'll follow this format (and update what I have going on.

gems/smithy/lib/smithy/vise/structure_shape.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Vise
+    # Represents a structure shape from the model.
+    class StructureShape < Shape
+      # (see Shape#initialize)
+      def initialize(id, shape)
+        super
+        @members = shape['members']
```
jterapin: Members could have their own traits. Could we do use `Vise::Trait.new` for those as well? I think it would be great to have an easy way to access members within a `Vise::Structure` too.

gems/smithy/lib/smithy/vise/model.rb
```diff
@@ -30,7 +30,19 @@ def operations
...
+            when 'service'
+              ServiceShape.new(id, shape)
+            when 'resource'
+              ResourceShape.new(id, shape)
+            when 'operation'
+              OperationShape.new(id, shape)
+            when 'structure'
+              StructureShape.new(id, shape)
+            else
+              Shape.new(id, shape)
```
jterapin: So this `Vise::Shape` will cover all the remaining types but we need to make sure we cover members for non-structure shapes (i.e. enum, intEnum, list, map and union)

gems/smithy/lib/smithy/vise/operations_parser.rb
```diff
@@ -4,10 +4,6 @@ module Smithy
   module Vise
     # @api private
     class OperationsParser
```
jterapin: What about having these methods public to access? (unknown if we need them later)
* a method to retrieve just operations defined on the service shape's operations (I guess technically you can retrieve this from the service shape)
* a method to retrieve just operations defined on the resource shapes

jterapin (commented):
Cool, I would like to use them in codegenerating the shapes file.

gems/smithy/lib/smithy/anvil/client/views/types.rb
mullermp: I don't think we should support enum constants and yes we will need unions but that's not a concern right now.

gems/smithy/lib/smithy/vise/service_shape.rb
mullermp: It depends on what the superclass is doing or in most cases it does not matter. I prefer it first if functionality is added and last if values set by the subclass need to influence the super class.

gems/smithy/lib/smithy/vise/structure_shape.rb
mullermp: Yes. The example model did not have those, but we can add them here.

gems/smithy/lib/smithy/vise/model.rb
mullermp: Yes, we can add some of those, as we build it out.

gems/smithy/lib/smithy/vise/operations_parser.rb
mullermp: I don't think we will need those, but if we do, we can refactor. Code for now.

gems/smithy/lib/smithy/anvil/client/views/errors.rb
```diff
@@ -19,7 +19,7 @@ def namespace
           def errors
             @model
               .shapes
-              .select { |_key, shape| shape.traits.any? { |_id, trait| trait.id == 'smithy.api#error' } }
+              .select { |_key, shape| shape.traits&.any? { |_id, trait| trait.id == 'smithy.api#error' } }
```
alextwoods, line 22: Should traits always return an empty hash?

gems/smithy/lib/smithy/vise/enum_shape.rb
```diff
@@ -0,0 +1,25 @@
...
+        super
+        @members = parse_members(shape['members'])
+      end
+
+      # @return [Hash<String, MemberShape>]
+      attr_reader :members
+
+      private
+
+      def parse_members(members)
```
alextwoods: A general naming discussion - I'm not sure I would call this "parsing".  I'd actually call this "building" - we're taking generic ruby data (Hashs) and using it to build a structured model.

gems/smithy/lib/smithy/vise/enum_shape.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Vise
+    # Represents an enum shape from the model.
+    class EnumShape < Shape
```
alextwoods, line 6: General thought / style question - is it worth having tons of these small shape files or, as we do in V3, should it just be a single file?

gems/smithy/spec/fixtures/vise/model.smithy
alextwoods, line 1: I assume that the model.smithy is used to create the model.json in this directory?  I don't love having both, but I'm not sure theres a good, simple way to auto translate this (without requiring that the smithy cli is installed) during tests.  

I wonder if we could at least add a comment to the json to say its generated, do not hand edit.

alextwoods (commented):
Nice - looking good

gems/smithy/lib/smithy/anvil/client/views/errors.rb
mullermp: The model may or may not have the traits hash. I think it makes sense to have nil instead. Slightly less usable but clear about whether traits were even present or not.

gems/smithy/spec/fixtures/vise/model.smithy
mullermp: I have a readme in the directory that basically says this already.

gems/smithy/lib/smithy/vise/enum_shape.rb
mullermp: I don't think it matters too much. I can put them in one file if that helps.

mullermp: I think the file would be too big otherwise. These shape classes do more than the runtime ones.

gems/smithy/lib/smithy/anvil/client/views/errors.rb
alextwoods: I'd disagree - There isn't a functional difference between nil traits and empty hash traits.  So I think we can remove a sharp edge from this interface by ensuring non-nil.

alextwoods (approved):

PR #228, opened 2024-12-20
Slim down vise and add new tests

mullermp:
It actually doesn't provide much value other than the indexes.


gems/smithy/lib/smithy/vise/service_index.rb
```diff
@@ -4,19 +4,21 @@ module Smithy
   module Vise
     # Finds a service shape in a set of shapes.
```
jterapin, line 5: Should update the documentation here since we are no longer having specific shapes.

gems/smithy/lib/smithy/vise/service_index.rb
```diff
@@ -4,19 +4,21 @@ module Smithy
...
     class ServiceIndex
-      def initialize(shapes)
-        @shapes = shapes
+      def initialize(model)
+        @service = find_service(model['shapes'])
       end
 
       # @return [ServiceShape] The service shape for the shapes.
```
jterapin: Documentation update here as well.

gems/smithy/lib/smithy/vise/service_index.rb
```diff
@@ -4,19 +4,21 @@ module Smithy
...
 
-        service = services.first
-        raise 'No service shape found' if service.nil?
+      def find_service(shapes)
+        services = shapes.select { |_, shape| shape['type'] == 'service' }
+        raise 'Multiple service shapes found' if services.size > 1
+        raise 'No service shape found' if services.empty?
 
-        service
+        services
```
jterapin: I think it makes more sense if this is singular than plural - since the method indicates that we are just returning a single service.

gems/smithy/lib/smithy/vise/service_index.rb
```diff
@@ -4,19 +4,21 @@ module Smithy
...
-        raise 'Multiple service shapes found' if services.size > 1
+      attr_reader :service
+
+      private
 
-        service = services.first
-        raise 'No service shape found' if service.nil?
+      def find_service(shapes)
+        services = shapes.select { |_, shape| shape['type'] == 'service' }
+        raise 'Multiple service shapes found' if services.size > 1
```
jterapin: I wonder if its better if we just build out the service based on the first one found than raising an error. What are the pros and cons of doing that? Also - how do other smithy-based SDKs handles this?

gems/smithy/lib/smithy/vise/operation_index.rb
```diff
@@ -4,12 +4,15 @@ module Smithy
...
     class OperationIndex
-      def initialize(shapes)
-        @shapes = shapes
+      RESOURCE_LIFECYCLE_KEYS = %w[create put read update delete list].freeze
+
+      def initialize(model)
+        @shapes = model['shapes']
       end
 
       def for(service)
```
jterapin, line 16: Add documentation on what this method takes and returns

gems/smithy/lib/smithy/anvil/client/views/types.rb
```diff
@@ -35,11 +34,11 @@ def documentation
             end
 
             def name
-              @structure.name
+              Vise::Shape.name(@id)
```
jterapin, line 37: So `Vise::Shape` will contain convenient class methods... What other ideas that you have in mind that we can build upon this? Maybe an easy way to access members if there's any? Or just keep it to minimum of altering given shape ID to whatever we need?

gems/smithy/lib/smithy/vise/shape.rb
```diff
@@ -13,55 +13,36 @@ module Vise
...
-        return nil unless @id.include?('$')
-
-        @id.split('$').last
-      end
-
-      private
-
-      def build_traits(traits)
-        return nil unless traits
+      def self.member_name(id)
```
jterapin, line 42: I was thinking about how this method will be used since while I was looking at generated `model.json` of our weather model - i couldn't find any shape id that includes member portion. 

(I also checked a random aws smithy `model.json` to see if there's any usage but didn't see there either)

gems/smithy/spec/smithy/vise/operation_index_spec.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Vise
+    describe OperationIndex do
+      let(:fixture) do
+        path = File.expand_path('../../fixtures/vise/model.json', __dir__)
+        JSON.parse(File.read(path))
```
jterapin: Could just do JSON.load_file ([documentation](https://ruby-doc.org/stdlib-3.0.0/libdoc/json/rdoc/JSON.html#method-i-load_file)), methinks.

gems/smithy/spec/smithy/vise/service_index_spec.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Vise
+    describe ServiceIndex do
+      let(:fixture) { File.expand_path('../../fixtures/vise/model.json', __dir__) }
+      let(:model) { JSON.parse(File.read(fixture)) }
```
jterapin: Ditto here from previous JSON-related comment.

gems/smithy/spec/smithy/vise/operation_index_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+
+      describe '#for' do
+        it 'returns a complete set of operations for the service' do
+          service =
+            fixture['shapes']
+            .select { |_, shape| shape['type'] == 'service' }
+          expected =
+            fixture['shapes']
+            .select { |_, shape| shape['type'] == 'operation' }
+            .reject { |id, _| id.include?('OrphanedOperation') }
```
jterapin, line 20: `OrphanedOperation` 🥺

jterapin (approved):
I love the simplification. Excited to update my pending PR to use its functionalities.

gems/smithy/lib/smithy/vise/service_index.rb
mullermp: They are still shapes, as a hash.

gems/smithy/lib/smithy/vise/service_index.rb
mullermp: I already had that, but we had a team discussion around validation.

gems/smithy/lib/smithy/anvil/client/views/types.rb
mullermp: I think that's all that this would have.

gems/smithy/lib/smithy/vise/shape.rb
mullermp: I don't think it's used right now, but I think paths in endpoint rules or jmespath expressions will use it.

gems/smithy/lib/smithy/vise/service_index.rb
jterapin: There was a lot of discussion we had yesterday and I forget where we landed on.

PR #229, opened 2024-12-23
Add raise response errors plugin

mullermp:
Adds the raise response errors plugin.

### DIFF
--- a/gems/smithy-client/lib/smithy-client/base.rb
+++ b/gems/smithy-client/lib/smithy-client/base.rb
@@ -9,9 +9,9 @@ module Smithy
       # @api private
       @plugins = PluginList.new(
         [
-          Plugins::Endpoint
+          Plugins::Endpoint,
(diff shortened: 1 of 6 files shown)


gems/smithy-client/lib/smithy-client/base.rb
```diff
@@ -9,9 +9,9 @@ class Base
       # @api private
       @plugins = PluginList.new(
         [
-          Plugins::Endpoint
+          Plugins::Endpoint,
```
alextwoods, line 12: I assume you need this temporarily for testing until I actually merge the endpoints?

alextwoods (approved):

gems/smithy-client/lib/smithy-client/base.rb
mullermp: It's part of your change is it not? I'm assuming it's still valid until you merge yours.

PR #230, opened 2024-12-23
HTTP response listeners and duck typing

mullermp:
This adds response listeners for http responses only. Error knowledge is pushed to output.

This also removes the base request/response/fields classes as they are not necessary as we should be coding for today's requirements, but may be added in the future. Only headers are necessary with our current sending pattern. Request/response is duck typed. Context will have default http request/response for now but may be refactored later.


gems/smithy-client/lib/smithy-client/net_http/handler.rb
```diff
@@ -18,66 +18,70 @@ def initialize(bytes_expected, bytes_received)
...
           bytes_expected = resp.headers['content-length'].to_i
-          return if bytes_expected == bytes_received
-
-          error = TruncatedBodyError.new(bytes_expected, bytes_received)
-          output.error = NetworkingError.new(error)
+          if bytes_expected == bytes_received
+            resp.signal_done
+          else
+            error = TruncatedBodyError.new(bytes_expected, bytes_received)
+            resp.signal_error(NetworkingError.new(error))
```
alextwoods, line 83: Why are we wrapping a truncated body error in another networking error?

gems/smithy-client/lib/smithy-client/output.rb
```diff
@@ -13,6 +13,9 @@ def initialize(options = {})
         @context = options[:context] || HandlerContext.new
         @data = options[:data]
         @error = options[:error]
+        @request = @context.request
+        @response = @context.response
+        @response.on_error { |error| @error = error }
```
alextwoods, line 18: This feels a bit round about to use the event mechanism to set the error on the output instead of having the http handler set it.

gems/smithy-client/lib/smithy-client/http/response.rb
```diff
@@ -1,39 +1,168 @@
...
+
+        def on_headers(status_code_range = nil, &block)
+          @listeners[:headers] << listener(status_code_range, block)
+        end
+
+        def on_data(&callback)
+          @listeners[:data] << callback
+        end
+
+        def on_done(status_code_range = nil, &callback)
```
alextwoods: I know this is what we do in V3 - but do we ever use the status code range in events?  To me this seems to add a lot of extra complexity that I'd rather avoid unless we think we really need it.  I already would prefer to not have events since they add yet another path for injecting logic into the lifecycle, but if we are adding them, we should at least try and use the simplest interface.

gems/smithy-client/lib/smithy-client/net_http/handler.rb
mullermp: It's a type of networking error, which wraps networking errors from net http.

alextwoods: I guess maybe more what I'm getting at is, why isnt our TruncatedBodyError a NetworkingError?

gems/smithy-client/lib/smithy-client/output.rb
mullermp: This is so we can signal error to populate the error. Otherwise I think the output object has to be constructed ahead of time and passed around the plugin.

gems/smithy-client/lib/smithy-client/http/response.rb
mullermp: Yes, we do use most of them. On headers we look at different ranges for when we do streaming retry. S3 responds with errors in 200 range so sometimes that range is necessary. We currently have listeners for all three signals (headers, data, done) and wither success (data) or errors (error) with ranges to handle weirdness. I think it makes sense to have them at these levels.

gems/smithy-client/lib/smithy-client/output.rb
mullermp: And if we're not signaling errors, listeners won't fire off. So we have to signal anyway.

alextwoods (approved):

gems/smithy-client/lib/smithy-client/net_http/handler.rb
mullermp: It is a networking error. Networking errors wrap other standard errors, so we just made our own, with a helpful message. I could make it inherit from NetworkingError but then I would have to change the initializer anyway.

PR #231, opened 2024-12-23
Fix weld spec

alextwoods:
*Description of changes:*

The previous spec was defining weld classes outside of the scope of :all and causing pollution into other tests.


mullermp (approved):
Polish could probably use this too.

PR #232, opened 2024-12-24
Add the response target plugin

mullermp:
Adds the response target plugin.

### DIFF
--- a/gems/smithy-client/lib/smithy-client/block_io.rb
+++ b/gems/smithy-client/lib/smithy-client/block_io.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    # IO object for response targets.
+    class BlockIO
+      # @param [Hash] headers (nil)
+      # @param [Proc] block
+      def initialize(headers = nil, &block)
+        @headers = headers
+        @block = block
+        @size = 0
+      end
+
+      # @return [Integer]
+      attr_reader :size
+
+      # @param [String] chunk
+      # @return [Integer]
+      def write(chunk)
+        @block.call(chunk, @headers)
+      ensure
+        chunk.bytesize.tap { |chunk_size| @size += chunk_size }
+      end
+
+      # @param [Integer] bytes (nil)
+      # @param [String] output_buffer (nil)
+      # @return [String, nil]
+      def read(bytes = nil, output_buffer = nil)
+        data = bytes ? nil : ''
+        output_buffer ? output_buffer.replace(data || '') : data
+      end
+    end
+  end
+end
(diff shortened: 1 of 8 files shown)


gems/smithy-client/lib/smithy-client/block_io.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    # IO object for response targets.
+    class BlockIO
+      # @param [Hash] headers (nil)
+      # @param [Proc] block
+      def initialize(headers = nil, &block)
+        @headers = headers
```
alextwoods, line 10: Why do we have headers on the blockIO?  Isn't it just intended to replace the body?

gems/smithy-client/lib/smithy-client/input.rb
```diff
@@ -65,7 +65,7 @@ def initialize(options = {})
       #   a request that may return a large payload that you don't want to
       #   load into memory.
       #
-      # @return [Output]
+      # @return [Output, nil]
```
alextwoods, line 68: What cases does this return nil? Shouldn't there always be an output (except ofc when an error is raised, but thats a different control flow)?

gems/smithy-client/lib/smithy-client/managed_file.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    # This utility class is used to track files opened by Smithy::Client.
+    # @api private
+    class ManagedFile < File
```
alextwoods, line 7: I'm not sure I really understand the purpose of this class.  In the response target handler we could just check for !closed? instead of open?

alextwoods (approved):

gems/smithy-client/lib/smithy-client/input.rb
mullermp: If there are no handlers, it can be nil. In practice this doesn't happen but satisfies linting.

gems/smithy-client/lib/smithy-client/managed_file.rb
mullermp: The method really doesn't matter. We wrap string targets (path names) with a managed file so that we know we can unlink files if it's a managed file type. If we are otherwise given a file, we don't want to unlink.

gems/smithy-client/lib/smithy-client/block_io.rb
mullermp: It's given as a block argument. I suppose you would want to know this for content type or length?

PR #234, opened 2024-12-27
Move service shape to plan and add service index traversal

mullermp:
Moves the service shape parsing to plan.


gems/smithy/lib/smithy/plan.rb
```diff
@@ -31,5 +32,18 @@ def initialize(model, type, options = {})
 
     # @return [Array<Polish>] The polishes that apply to this plan.
     attr_reader :polishes
+
+    # @return [Hash<String, Hash>] The service shape for the shapes.
+    attr_reader :service
```
alextwoods, line 37: Can this just be the service instead of a hash with a single entry?

mullermp: I think we also want the id for any lookup. So either the id or this hash.. I chose the hash because we would often have to look up based on that id anyway.

mullermp: The model parsing uses "shapes" which are the hashes. I think this makes sense to stay consistent for all shapes. Some times you need to re-reference via the ID. Service could work differently but then it's more confusing I think. So if we just expect a hash everywhere, it's easy to follow.

alextwoods: I guess I was confused - in the current ServiceIndex the `service` returned is a Hash with a single key of the id and value of the service shape (a Hash with values of Hash) so code that used it needed to do something like `service.values.first`. Maybe RBS typing would help here, but I can't actually tell what this method returns (ie, is it a map to shape or is it the shape?)

gems/smithy/lib/smithy/vise/operation_index.rb
```diff
@@ -2,73 +2,18 @@
...
 
-      def parse_resource_collection_operations(resource, operations)
-        resource['collectionOperations']&.collect do |shape|
-          id = shape['target']
-          operations[id] = @shapes[id]
-        end
+      # @param [Hash] operation Operation shape
+      # @return [Hash<String, Hash>] The shapes for the operation.
+      def shapes_for(operation)
+        @memo[operation] ||= @operation_parser.shapes_for(operation)
```
alextwoods: Is this class just a cache on top of the operation parser? It feels like it doesn't really need to exist and this memoization/caching could just be done in the OperationParser.

gems/smithy/lib/smithy/vise/structure_parser.rb
```diff
@@ -0,0 +1,54 @@
...
+        _id, structure = structure.first
+        structure['members']&.each_value { |member| parse_member(member, shapes) }
+        shapes
+      end
+
+      private
+
+      def parse_member(member_ref, shapes)
+        target = member_ref['target']
+        return if PRELUDE_SHAPES.include?(target)
```
alextwoods, line 46: I'm not quite sure this is how we should be handeling prelude shapes.  This leads to code that uses the model also needing to know about prelude shapes.

I'd potentially lean towards adding a default weld with a pre-process that checks the model for prelude shapes, and if missing, adds them in.  That ensures model integrity and simplifies view code.

mullermp: Do prelude shapes get included in c2j conversions of a smithy model?

gems/smithy/lib/smithy/vise/operation_index.rb
mullermp: Yeah, this one isn't quite useful. I had a thought last night to actually rewrite this index to get input and output shapes, and potentially a shape walker that does a depth first traversal of an operation. That would be useful for generating request and response documentation.

gems/smithy/lib/smithy/plan.rb
mullermp: My understanding of "shape" was a hash with id to shape data, and multiple shapes are the same but flat (one hash with id to shape data). I tried to make all of the interfaces accept a hash even if it's one key (one shape). I did this because we often need the id. I'll think more on how to change this however I'm keeping it consistent for now.

gems/smithy/lib/smithy/vise/structure_parser.rb
alextwoods: Without doing a deep dive into the code - the answer is yes, but they get translated (possibly a synthetic shape is created?). Here's an example:
* Smithy model: https://github.com/aws/aws-models/blob/master/notifications/smithy/model.json#L2145
* C2J: https://github.com/aws/aws-models/blob/master/notifications/2018-05-10/api-2.json#L1104

mullermp: Hm ok. It's slightly different since c2j has type fields. We probably want to add prelude shapes to the model in a preprocess only if they are referenced. Quite possibly we should do it before preprocess hooks too (bake it in). The prelude option in smithy run I don't believe adds these (can you check?)

alextwoods: Yeah - I think ideally we would add them only if they are referenced, but its simpler to just add them all and there aren't many (not including traits) and we have code that gets the service closure anyway (ie, this pr).  

Just tried with smithy run with the model plugin with `includePreludeShapes: true` - it does not add the "synthetic" like in c2j, just adds definitions for the shapes like:
```
"smithy.api#String": {
            "type": "string"
        },
```

It does also add all of the trait definitions too which we don't need.

mullermp: I think the code still needs to check for prelude shapes by checking for a types key. We can possibly go without changing the model. Let's see how this goes as we add other code generation pieces.

alextwoods: If the prelude shapes were included I don't think we'd need to special case anything here, but I guess can we go without and deal with target shapes not being in the model.

gems/smithy/lib/smithy/plan.rb
alextwoods: Ah I understand.  I find that a little hard to work - maybe RBS typing would help, but this is fine for now.

alextwoods (approved):

gems/smithy/lib/smithy/vise/structure_parser.rb
mullermp: How would it be handled for documentation? Some logic would need to check the existence of the type key and have a map of type to ruby type. But we may as well have a map of prelude shapes id to Ruby type.

alextwoods: For documentation its much simpler if you have the prelude types in the model.  I think it reduces special casing.  Consider the following model:

```json
{
  "example#S": {
      "type": "structure",
      "members": {
          "member1": {
              "target": "example#NonPrelude"
          },
          "member2": {
              "target": "smithy.api#String"
          }
      }
  },
  "example#NonPrelude": {
            "type": "string"
  }
}
```

Without having the prelude String defined, you need a check on target before getting it to get the smithy type. But if its included in the model, then you just get the member's target and get the type from it.  Then map the Smithy type to Ruby Type.  No special case check of prelude shape id first.

mullermp: Possibly - I think in either case we need a mapping of something to ruby type, and we can map prelude shape target to ruby class and avoid adding to the model and indexing into members an extra step. We can add this in another PR.

gems/smithy/lib/smithy/vise/operation_index.rb
mullermp: Deleted

gems/smithy/lib/smithy/plan.rb
mullermp: Discussed offline - I think we NEED service id here for reconciling module names eventually. I can also make the documentation better.

PR #235, opened 2024-12-31
Fix `shapes_for` for recursive shapes

alextwoods:
*Description of changes:*
the `shapes_for` method breaks for recursive shapes.  This PR adds a visited to track.


gems/smithy/spec/fixtures/service_index/model.smithy
```diff
@@ -91,3 +91,22 @@ structure OrphanedStructure {}
 
 @error("client")
 structure OrphanedError {}
+
+operation RecursiveShapesOperation {
```
mullermp: Probably can add a recursive service with just a simple operation and structure that references itself, then other indexes can use this same fixture. I can do that later if you want.

mullermp (approved):

PR #236, opened 2025-01-01
Request example

mullermp:
Only does the request example right now.

### DIFF
--- a/gems/smithy/lib/smithy/anvil/client/request_syntax_example.rb
+++ b/gems/smithy/lib/smithy/anvil/client/request_syntax_example.rb
@@ -20,56 +20,74 @@ module Smithy
             ''
           ]
           example = "  output = client.#{Vise::Shape.name(@id).underscore}"
-          input = @shapes[@operation['input']['target']]
+          target = @operation['input']['target']
+          return docstrings << example if target == 'smithy.api#Unit'
+
+          input = @shapes[target]
           return docstrings << example unless input['members'].any?
 
-          docstrings + ["#{example}("] + struct(input, '    ', Set.new) + ['  )']
+          docstrings << "#{example}("
+          docstrings << struct(input, '    ', Set.new)
+          docstrings << '  )'
+          docstrings.flatten
         end
 
         private
 
         def struct(shape, indent, visited)
-          lines = ["#{indent}{"]
+          docstrings = ["#{indent}{"]
           shape['members'].each do |member_name, member_shape|
-            lines << struct_member(member_name, member_shape, indent, visited)
+            docstrings << struct_member(member_name, member_shape, indent, visited)
           end
-          lines << "#{indent}}"
+          docstrings << "#{indent}}"
         end
 
         def struct_member(member_name, member_shape, indent, visited)
-          entry = "#{indent}  #{member_name.underscore}: #{target_value(member_shape, "#{indent}  ", visited)},"
-          required = member_shape['traits']&.any? { |k, _| k == 'smithy.api#required' }
-          apply_comments(entry, { required: required })
+          docstrings = []
+          target = member_shape['target']
+          shape = @shapes[target]
+
+          value = target_value(member_shape, "#{indent}  ", visited)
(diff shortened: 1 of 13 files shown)


gems/smithy/lib/smithy/anvil/client/request_syntax_example.rb
```diff
@@ -0,0 +1,162 @@
...
+        end
+
+        def visited_shape?(target, visited)
+          # Prelude shapes are used more than once usually
+          !Vise::PRELUDE_SHAPES.include?(target) && visited.include?(target)
+        end
+
+        def struct(struct_shape, indent, visited, top_level: false)
+          lines = []
+          lines << '{' unless top_level
```
alextwoods: we're adding {} at the top level as well - could we remove the special case for top_level here?

gems/smithy/lib/smithy/anvil/client/views/endpoint_parameters.rb
```diff
@@ -9,8 +9,8 @@ class EndpointParameters < View
           def initialize(plan)
             @plan = plan
             @model = plan.model
-            service = @plan.service
-            @endpoint_rules = service.values.first['traits']['smithy.rules#endpointRuleSet']
+            _, service = @plan.service.first
```
alextwoods: I still find this confusing, but if we're keeping the shape as a hash of id => data, I prefer the orig code to using a `_, service =` personally.

gems/smithy/spec/interfaces/client/client_spec.rb
```diff
@@ -27,4 +27,97 @@
...
+            },
+            structure: {
+              member: "String"
+            },
+            union: TODO: union
+          )
+      EXAMPLE
+      client_file = File.join(@tmpdir, 'lib', 'all_shapes', 'client.rb')
+      top_level = rdoc.parse_files([client_file]).first
+      actual = top_level.find_class_or_module('AllShapes::Client').find_method_named('operation')
```
alextwoods, line 93: nice! these tests are cool

projections/weather/lib/weather/client.rb
```diff
@@ -15,25 +15,40 @@ def initialize(*args)
       super
     end
 
-    # TODO!
+    # @example Request syntax with placeholder values
+    # 
+    #   output = client.get_city(
+    #     city_id: "CityId", # required
+    #   )
```
alextwoods: Do these need an empty comment line after for rdoc to render correclty?

alextwoods (approved):

gems/smithy/lib/smithy/anvil/client/request_syntax_example.rb
mullermp: My latest commit removes top level brackets. Which do you prefer? Outer hash causes another level of indent but outer hash is necessary for options (not shown here)

projections/weather/lib/weather/client.rb
mullermp: I'm not sure but that can be added later. I need to do a response example
 I am thinking of doing a to_h representation instead of method calls. I could reuse this code or do both examples in this class. Let me know what you think.

gems/smithy/lib/smithy/anvil/client/request_syntax_example.rb
alextwoods: I'd lean towards not using the outer hashes I think, so ignore this :-)

mullermp: I may change this to multiline:
```
params = {
 ...
}
options = { ... }
operation(params, options)
```

gems/smithy/lib/smithy/anvil/client/views/endpoint_parameters.rb
mullermp: I'll change it back

alextwoods (approved):
Do you need to rebuild the projections?

PR #237, opened 2025-01-03
[DRAFT] Demo different options for rbs creation

alextwoods:
*Description of changes:*
This PR demos three different methods for generating rbs files:
1. types.rbs - update the hammer method in View to accept a template file. (one view, multiple templates)
2. client.rbs - Duplicate view in the sigs folder (separate view and separate template with matchign names)
3. errors.rbs - Inheritance, create a view in the sigs folder that inherits the regular view. (separate view and separate template with matching names)


gems/smithy/lib/smithy/anvil/client/views/sig/errors.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Anvil
+    module Client
+      module Views
+        module Sig
+          # Still its own class, but inherit from the regular view to reduce duplication
+          class Errors < Errors
```
mullermp, line 9: I prefer own class, whether it is duplicate or inherited, is up to the class.

PR #238, opened 2025-01-03
Fix shapes_for to incldue list and map members

alextwoods:
*Description of changes:*
Fix shapes_for to incldue list and map members

### DIFF
--- a/gems/smithy/lib/smithy/vise/structure_parser.rb
+++ b/gems/smithy/lib/smithy/vise/structure_parser.rb
@@ -60,7 +60,7 @@ module Smithy
           parse_member(shape['key'], shapes, visited)
           parse_member(shape['value'], shapes, visited)
         else
-          # covers structure, union
+          # covers all other shapes including structure, union, enum
           shape['members']&.each_value { |member| parse_member(member, shapes, visited) }
         end
       end
--- a/gems/smithy/lib/smithy/vise/structure_parser.rb
+++ b/gems/smithy/lib/smithy/vise/structure_parser.rb
@@ -49,7 +49,16 @@ module Smithy
         shape = @shapes[target]
         shapes[target] = shape
         visited[target] = true
-        shape['members']&.each_value { |member| parse_member(member, shapes, visited) }
+        case shape['type']
+        when 'list'
+          parse_member(shape['member'], shapes, visited)
+        when 'map'
+          parse_member(shape['key'], shapes, visited)
+          parse_member(shape['value'], shapes, visited)
+        else
+          # covers structure, union
+          shape['members']&.each_value { |member| parse_member(member, shapes, visited) }
+        end
       end
     end
   end
--- a/gems/smithy/lib/smithy/vise/structure_parser.rb
+++ b/gems/smithy/lib/smithy/vise/structure_parser.rb
@@ -49,6 +49,10 @@ module Smithy
         shape = @shapes[target]
         shapes[target] = shape
         visited[target] = true
+        parse_members(shape, shapes, visited)
+      end
+
+      def parse_members(shape, shapes, visited)
         case shape['type']
         when 'list'
           parse_member(shape['member'], shapes, visited)
--- a/gems/smithy/spec/fixtures/service_index/model.smithy
+++ b/gems/smithy/spec/fixtures/service_index/model.smithy
@@ -56,9 +56,7 @@ operation DeleteResource {
 }
 
 @readonly
-operation ListResources {
-    output: Structure
-}
+operation ListResources {}
 
 operation ResourceOperation {
     input: Structure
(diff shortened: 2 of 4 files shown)


gems/smithy/lib/smithy/vise/structure_parser.rb
```diff
@@ -49,7 +49,16 @@ def parse_member(member_ref, shapes, visited)
...
         visited[target] = true
-        shape['members']&.each_value { |member| parse_member(member, shapes, visited) }
+        case shape['type']
+        when 'list'
+          parse_member(shape['member'], shapes, visited)
+        when 'map'
+          parse_member(shape['key'], shapes, visited)
+          parse_member(shape['value'], shapes, visited)
+        else
+          # covers structure, union
```
jterapin: This covers also enum and intEnum as well

gems/smithy/spec/fixtures/service_index/model.smithy
```diff
@@ -4,7 +4,7 @@ namespace smithy.ruby.tests
 
 service ServiceIndex {
     version: "1.0.0"
-    operations: [Operation]
+    operations: [Operation, MapOperation]
```
mullermp: Can we keep 1 operation?

gems/smithy/spec/fixtures/service_index/model.smithy
```diff
@@ -56,7 +56,11 @@ operation DeleteResource {
 }
 
 @readonly
-operation ListResources {}
+operation ListResources {
+    output:= {
```
mullermp: Are we using new syntax or old syntax? I've been using old/explicit.

gems/smithy/spec/fixtures/service_index/model.smithy
```diff
@@ -68,13 +72,38 @@ operation NestedResourceOperation {
...
 structure Structure {
     @required
     id: String
     name: String
 }
 
+list StructureList {
+    member: StructureMember
+}
+
```
mullermp, line 87: ListMember? And could also target string

jterapin (approved):
Nice!

gems/smithy/spec/fixtures/service_index/model.smithy
```diff
@@ -68,13 +72,38 @@ operation NestedResourceOperation {
...
+map StructureMap {
+    key: MapKey
+    value: MapValue
+}
+
+string MapKey
+
+structure MapValue {
```
mullermp: Key and value could just be shapes that target strings for simplicity.

gems/smithy/spec/fixtures/service_index/model.smithy
```diff
@@ -68,13 +72,38 @@ operation NestedResourceOperation {
     input: Structure
 }
 
+operation MapOperation {
+    input:= {
+        map: StructureMap
```
mullermp: Map could probably be in put resource.

gems/smithy/spec/fixtures/service_index/model.smithy
alextwoods: Sure, I'll just add Map to the Operation

gems/smithy/spec/fixtures/service_index/model.smithy
alextwoods: Shouldn't really matter, these aren't even used by the test.  I generally prefer the newer/shorter syntax for readability though.

gems/smithy/spec/fixtures/service_index/model.smithy
alextwoods: It can't that won't test the case.  It needs to be a shape that is not referenced anywhere else.

gems/smithy/spec/fixtures/service_index/model.smithy
alextwoods: See comment on list.  They must target shapes that are not referenced anywhere else.

gems/smithy/spec/fixtures/service_index/model.smithy
alextwoods: Map didn't really make sense in any of the resources methods conceptually to me.

gems/smithy/lib/smithy/vise/structure_parser.rb
alextwoods: Oh yeah, good call

gems/smithy/spec/fixtures/service_index/model.smithy
jterapin: I think we should pick one or the other for consistency.

gems/smithy/spec/fixtures/service_index/model.smithy
mullermp: I was saying I think the member should be called ListMember, that targets a string?

gems/smithy/spec/fixtures/service_index/model.smithy
mullermp: I prefer old syntax as it's explicit and when I search for ListResourcesOutput, I can find it in the model.

PR #239, opened 2025-01-03
Rubocop Polish and clean ups

alextwoods:
*Description of changes:*
Minor cleanups and fixes to endpoints.


projections/weather/spec/weather/endpoint_provider_spec.rb
```diff
@@ -7,5 +7,32 @@
 module Weather
   describe EndpointProvider do
     subject { EndpointProvider.new }
+
+    context "Endpoint set" do
```
jterapin: Surprised that rubocop didn't yell about double quotes here.

projections/weather/spec/weather/endpoint_provider_spec.rb
```diff
@@ -7,5 +7,32 @@
...
+      let(:expected) do
+        {"endpoint"=>{"url"=>"https://example.com"}}
+      end
+
+      it 'produces the expected output from the EndpointProvider' do
+        params = EndpointParameters.new(**{:endpoint=>"https://example.com"})
+        endpoint = subject.resolve_endpoint(params)
+        expect(endpoint.uri).to eq(expected['endpoint']['url'])
+        expect(endpoint.headers).to eq(expected['endpoint']['headers'] || {})
+        expect(endpoint.properties).to eq(expected['endpoint']['properties'] || {})
```
jterapin, line 21: I guess we will populate `expected` with `headers` and `properties` in the future?

projections/weather/spec/weather/endpoint_provider_spec.rb
```diff
@@ -7,5 +7,32 @@
...
+    context "Endpoint not set" do
+      let(:expected) do
+        {"error"=>"Endpoint is not set - you must configure an endpoint."}
+      end
+
+      it 'produces the expected output from the EndpointProvider' do
+        params = EndpointParameters.new(**{})
+        expect do
+          subject.resolve_endpoint(params)
+        end.to raise_error(ArgumentError, expected['error'])
```
jterapin, line 34: Can't we just do `"Endpoint is not set - you must configure an endpoint."` instead of retrieving from the hash? Maybe I'm not seeing something very obvious here.  I'm thinking we are setting this up to populate more expectations in the future.

projections/weather/spec/weather/endpoint_provider_spec.rb
alextwoods: It does :-(. However, this is arbitrary input so could contain characters that require double quotes.  We could add a view method called `quote` that woudl inspect a given string and decide how to quote it.... but I'm not 100% sure thats worth the effort vs say, running rubocop as a "polish"

projections/weather/spec/weather/endpoint_provider_spec.rb
alextwoods: These are based no the endpoint tests.  The tests for this ruleset dont have any headers or properties so nothing to test.

projections/weather/spec/weather/endpoint_provider_spec.rb
alextwoods: We could do that - this was just what we did in V3 since it simplifies the code for the the let(:expected) bit.

jterapin: Disregard, ditto.

projections/weather/spec/weather/endpoint_provider_spec.rb
jterapin: Just realized that these are projections, haha. Not enough coffee. Please disregard.

mullermp (approved):

gems/smithy/lib/smithy/polishes/rubocop.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Polishes
+    # Runs RuboCop autocorrect-all on the generated projections.
+    class RuboCop < Polish
+      def polish(_artifact)
```
alextwoods: A nit - this is an enumerable right? Should this be `artifacts`? (obviously doesn't matter here since unused, just a general question on polishes).

gems/smithy/lib/smithy/polishes/rubocop.rb
```diff
@@ -0,0 +1,20 @@
...
+        require 'rubocop'
```
alextwoods, line 8: We need to now add this to the gemspec as a dependency I think.

gems/smithy/lib/smithy/polishes/rubocop.rb
```diff
@@ -0,0 +1,20 @@
...
+        puts "Running RuboCop --autocorrect-all on #{destination_root}"
+        rubocop = ::RuboCop::CLI.new
+        args = [
+          '--autocorrect-all',
+          '--display-only-fail-level-offenses',
+          destination_root
+        ]
+        rubocop.run(args)
```
alextwoods, line 18: I really like this as a general approach - but I think we should not use it when running specs (it adds unessecary time + extra std out output).  I think theres probably a few ways we can do that.

PR #240, opened 2025-01-06
Cleanup spec stdout, no rubocop on specs

alextwoods:
*Issue #, if available:*

*Description of changes:*
* cleanup spec stdout (dont log test sdk file generation)
* no rubocop on specs


gems/smithy/spec/spec_helper.rb
```diff
@@ -56,7 +69,8 @@ def create_plan(modules, model, type, options)
       plan_options = {
         gem_name: options[:gem_name] || Smithy::Tools::Namespace.gem_name_from_namespaces(modules),
         gem_version: options[:gem_version] || '1.0.0',
-        destination_root: options[:destination_root] || Dir.mktmpdir
+        destination_root: options[:destination_root] || Dir.mktmpdir,
+        skip_polishes: options.fetch(:skip_polishes, true)
```
mullermp: Can you use || or make all of these fetch?

gems/smithy/spec/spec_helper.rb
```diff
@@ -46,6 +50,15 @@ def cleanup(module_names, tmpdir)
 
     private
 
+    def with_captured_stdout
+      original_stdout = $stdout  # capture previous value of $stdout
```
mullermp: Comments are not necessary here.. it's clear what is going on.

gems/smithy/spec/spec_helper.rb
```diff
@@ -21,7 +22,10 @@ def generate(modules, type, options = {})
       model = load_model(modules, options)
       plan = create_plan(modules, model, type, options)
 
-      Smithy.smith(plan)
+      with_captured_stdout do
```
mullermp: Single line

gems/smithy/lib/smithy/plan.rb
```diff
@@ -14,8 +14,18 @@ def initialize(model, type, options = {})
 
       Welds.load!(self)
       Polishes.load!(self)
-      @welds = Welds.for(model)
-      @polishes = Polishes.for(model)
+      @welds =
+        if options[:skip_welds]
```
mullermp: Can you add --welds [only, this] and --no-welds to CLI?

gems/smithy/spec/spec_helper.rb
alextwoods: || doesn't work well for booleans, so switched to using fetch for everything.

gems/smithy/lib/smithy/plan.rb
alextwoods: add welds/polishes cli boolean options (which allows you to use either `--no-welds` or `--skip-welds`)

gems/smithy/lib/smithy/plan.rb
```diff
@@ -14,8 +14,18 @@ def initialize(model, type, options = {})
 
       Welds.load!(self)
       Polishes.load!(self)
-      @welds = Welds.for(model)
-      @polishes = Polishes.for(model)
+      @welds =
+        if options[:welds]
```
mullermp: I think it's fine to ternary these if they are simple.

gems/smithy/spec/smithy/plan_spec.rb
```diff
@@ -5,7 +5,7 @@ module Smithy
     let(:fixture) { File.expand_path('../fixtures/weather/model.json', __dir__) }
     let(:model) { JSON.load_file(fixture) }
     let(:type) { :client }
-    let(:options) { {} }
+    let(:options) { { welds: true, polishes: true } }
```
mullermp: Is there a way to invert this or default in plan initialize? New plans should not have these opt in but rather opt out.

no-welds and no-plan will set nil and not false so maybe there's a distinction there.

mullermp (approved):

gems/smithy/spec/smithy/plan_spec.rb
alextwoods: Theres not a good way to invert this with Thor's boolean arguments that I could find.  I guess I could add a `--no-weld` boolean argument instead?

alextwoods: Edit: Nvm - I was able to just add negative options for these instead, which I agree is preferable.

PR #241, opened 2025-01-06
Response syntax example and operation examples

mullermp:
Reuse the request syntax example


gems/smithy/lib/smithy/anvil/client/operation_examples.rb
```diff
@@ -0,0 +1,87 @@
...
+
+        def format_hash(shape, newline_indent = '  ')
+          return '{}' unless shape
+
+          json = JSON.pretty_generate(
+            underscore_keys(shape),
+            object_nl: "\n#{newline_indent}",
+            array_nl: "\n#{newline_indent}"
+          )
+          json.gsub(/"([^"]+)":/, '\1:')
```
alextwoods, line 76: Using json generate seems like a slightly janky way to do this - I'd worry a little bit about escape character edge cases here.

gems/smithy/spec/interfaces/client/client_syntax_example_spec.rb
```diff
@@ -0,0 +1,149 @@
...
+          },
+          complex_map: {
+            "String" => {
+              member: "String"
+            }
+          },
+          structure: {
+            member: "String"
+          },
+          union: TODO: union
```
alextwoods, line 100: Are we tracking these TODOs anywhere?

projections/weather/lib/weather/client.rb
```diff
@@ -32,6 +44,14 @@ def get_city(params = {}, options = {})
     #   params = {}
     #   options = {}
     #   output = client.get_current_time(params, options)
+    #
+    # @example Response structure with placeholder values
+    #
+    #   output.to_h #=>
+    #   {
+    #     time: Time.now, # required
```
alextwoods, line 52: the "Time.now" is a little weird - is that what we do in V3?  I guess it does do a better job of conveying the type of the returned value though than what would actually be returned.

mullermp: We do for modeled API examples. I think sample return data is more useful in these cases. RBS can provide typing.

gems/smithy/spec/interfaces/client/client_syntax_example_spec.rb
mullermp: Other than searching for TODO, no. We could implement a todo task like rails does if we care.

gems/smithy/lib/smithy/anvil/client/operation_examples.rb
mullermp: It seems janky but I think it handles every case and quite simple? Is there a case you can think of? gsub will only remove the quotes on keys.

alextwoods: What about a map with string keys? Or keys with `\"`?  I think this is probably reasonable for now though.

alextwoods (approved):

gems/smithy/lib/smithy/anvil/client/operation_examples.rb
mullermp: You are correct, it doesn't handle maps. I've added a todo for later but right now it does 99% of the work.

PR #243, opened 2025-01-08
Normalize module names

mullermp:
We are not allowed to have fun anymore.


gems/smithy/lib/smithy/views/client.rb
```diff
@@ -0,0 +1,30 @@
...
+
+module Smithy
+  module Views
+    # @api private
+    module Client; end
+  end
+end
+
+# helper classes
+
```
alextwoods: Style nit: extra line?

alextwoods (approved):
Nice - I like the Anvil -> Views + flattening and I think model makes more sense than vice.

In terms of other names: 

I think polish is pretty clear on what it does - though I'm also not 100% convinced it should be separate functionality from Welds (ie, why not just have a `polish` or `post_process` hook on weld.

For `Weld` - Naming for this sort of functionality is hard.  "Plugin" is already very overloaded.  I think an alternative would be "Integration" or "Extension" or something like that, but its also not necessarily more descriptive than "Weld".  "Weld" also feels some what similar to "Railtie", so I think its reasonable to keep it.

gems/smithy/lib/smithy.rb
```diff
@@ -3,16 +3,16 @@
 require 'rails/railtie'
 require 'thor'
 
-require_relative 'smithy/tools'
```
jterapin, line 6: I actually don't mind that utils are called tools. I felt it was pretty straightforward there.

jterapin (approved):
Lookin good 👍

gems/smithy/lib/smithy/views/client.rb
mullermp: I'll change it but I prefer attaching docstrings to code if it's for the code and I use new lines for separating sections.

gems/smithy/lib/smithy/rivets/rubocop.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Rivets
+    # Runs RuboCop autocorrect-all on the generated projections.
+    class Rubocop < Rivet
+      TEST_NAMESPACES = [
```
mullermp: Since polishes and welds are now the same concept, we can't skip them selectively in tests, because we will still want to load them. We can instead use for?(service) (instead of model) as a way to select (can also look at the plan if needed). We can make all of the tests use smithy.ruby.tests (requires manual sync) or perhaps some upstream fixes to make the tests follow standard naming.

gems/smithy/lib/smithy/views/view.rb
```diff
@@ -22,7 +22,7 @@ def inherited(subclass)
         attr_accessor :template_file
       end
 
-      def hammer
+      def render
```
alextwoods, line 25: Nice, I like `render` for this.

gems/smithy/lib/smithy/views/client/endpoint_plugin.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Views
+    module Client
+      # @api private
+      class EndpointPlugin < View
```
alextwoods, line 7: Nice, I like this flattening.  I think the total number of views we'll have is fairly limited, so the flatter structure makes sense.

gems/smithy/lib/smithy/command.rb
```diff
@@ -27,11 +27,6 @@ def self.gem_options!
         method_option :gem_namespace, type: :string,
                                       desc: 'The namespace of the gem to generate, e.g. `MyGem::Namespace`.' \
                                             'If not provided, the gem name will be used to infer the namespace.'
-        method_option :skip_welds, type: :boolean, default: false,
```
alextwoods, line 30: Is there not a new skip_rivets?

gems/smithy/lib/smithy/rivet.rb
```diff
@@ -0,0 +1,49 @@
...
+    # Pre-process the model. Called before the model is loaded.
+    # @param [Hash] model
+    def pre_process(model)
+      model
+    end
+
+    # Post-process the artifacts after they are generated.
+    # @param artifacts [Enumerator<String, String>] The artifacts that were generated.
+    #  The key is the path of the file and the value is the content of the file.
+    def post_process(artifacts)
```
alextwoods: Nice - I like the integration of Polish into here.

gems/smithy/lib/smithy/rivet.rb
```diff
@@ -0,0 +1,49 @@
...
+      @plan = plan
+      # Necessary for Thor::Base and Thor::Actions
+      self.options = { force: true }
+      self.destination_root = plan.options[:destination_root]
+    end
+
+    # Called to determine if the rivet should be applied for this model.
+    # @param [Hash] service Service shape
+    # @return [Boolean] (true) True if the rivet should be applied, false otherwise.
+    def for?(service) # rubocop:disable Lint/UnusedMethodArgument
```
alextwoods: Is this just the Service shape? If so, I'm not sure its sufficient?  

EDIT: Nvm - I guess you have access to the `@plan` and therefore the model as well right?

gems/smithy/lib/smithy.rb
```diff
@@ -7,11 +7,9 @@
 require_relative 'smithy/forge'
 require_relative 'smithy/model'
 require_relative 'smithy/plan'
-require_relative 'smithy/polish'
-require_relative 'smithy/polishes'
+require_relative 'smithy/rivet'
```
alextwoods: I like the integration of the polish/post-process into weld, but why the rename to rivet?  I like name weld.

Though conceptually both are about equally meaningful/applicable to the functionality, so I'm fine with the rename just curious why.

gems/smithy/lib/smithy/command.rb
mullermp: I don't think it's necessary is it?

gems/smithy/lib/smithy/rivet.rb
mullermp: It's just the service shape. We can improve this if we need more scope. I had removed `plan` from the instance variable. Why else would you need the full model?

gems/smithy/lib/smithy/command.rb
alextwoods: You asked for those to be added in the first place.  I think they're generally useful.  But I'm fine removing it from the CLI options as long as its still supported in the internal plan.

gems/smithy/lib/smithy/plan.rb
```diff
@@ -12,10 +12,8 @@ def initialize(model, type, options = {})
       @options = options
       @service = find_service(model['shapes'])
 
-      Welds.load!(self)
-      Polishes.load!(self)
-      @welds = options[:skip_welds] ? [] : Welds.for(model)
-      @polishes = options[:skip_polishes] ? [] : Polishes.for(model)
+      Rivets.load!(self)
+      @rivets = Rivets.for(@service)
```
alextwoods: I think we still need the ability to skip these (how are they getting skipped in tests now?)

gems/smithy/lib/smithy/command.rb
mullermp: I did as a way to skip polishes. But we need welds for like endpoint rules. So the option doesn't make sense anymore with weld and polish as one thing. So now we have to skip based upon `for?`

gems/smithy/lib/smithy/plan.rb
mullermp: Talked offline

alextwoods (approved):

PR #244, opened 2025-01-08
Implement operationInput specs + fix bindings + endpoints documentation

alextwoods:
*Description of changes:*
*Implement operation input specs
* fixes various parameter binding issues.
* endpoint resolver/params docs.


gems/smithy/lib/smithy/anvil/client/endpoint_parameter.rb
```diff
@@ -73,21 +83,14 @@ def client_context_doc
         # @return [value, source].  source may be one of [operation, config, default]
         def endpoint_parameter_value(operation)
           unless operation.nil?
-            source = 'operation'
             value = static_context_param(operation)
-            value ||= context_param_value(operation)
-            value ||= operation_context_param_value(operation)
+            value = context_param_value(operation) if value.nil?
```
alextwoods: These changes are required to correctly handle "false" boolean values

gems/smithy/lib/smithy/anvil/client/templates/specs/endpoint_provider_spec.erb
```diff
@@ -27,6 +40,42 @@ module <%= namespace %>
...
+
+<% test_case.operation_inputs.each do |operation_input| -%>
+      it 'produces the correct output from the client when calling <%= operation_input.operation_name %>' do
+        client = Client.new(
+<% operation_input.client_params.each do |p| -%>
+          <%= p.param %>: <%= p.value %>,
+<% end -%>
+           plugins: [stub_send]
+        )
+ <% if test_case.expect_error? -%>
```
jterapin: Empty left space

projections/weather/lib/weather/endpoint_parameters.rb
```diff
@@ -4,7 +4,10 @@
 
 module Weather
   # Endpoint parameters used to resolve endpoints per request.
-  # TODO: Documentation
+  # @!attribute endpoint
+  #   Endpoint
+  #
+  #   @return [String]
```
jterapin, line 10: `@return` should align with `@!attribute` - I think

jterapin (approved):

projections/weather/lib/weather/endpoint_parameters.rb
mullermp: It's correct as is, because it's sub documentation under attribute. However the documentation "Endpoint" is really vague and not clear that it's from documentation shape.

gems/smithy/lib/smithy/anvil/client/endpoint_parameter.rb
mullermp: Makes sense

gems/smithy/lib/smithy/anvil/client/templates/endpoint_parameters.erb
```diff
@@ -4,7 +4,14 @@
 
 module <%= namespace %>
   # Endpoint parameters used to resolve endpoints per request.
-  <%= documentation %>
+<% parameters.each do |param| -%>
+  # @!attribute <%= param.name %>
+<% if param.documentation -%>
+  #   <%= param.documentation %>
+<% end -%>
+  #
```
mullermp: Can probably remove this blank line for compactness.

gems/smithy/lib/smithy/anvil/client/templates/endpoint_parameters.erb
```diff
@@ -4,7 +4,14 @@
 
 module <%= namespace %>
   # Endpoint parameters used to resolve endpoints per request.
-  <%= documentation %>
+<% parameters.each do |param| -%>
+  # @!attribute <%= param.name %>
+<% if param.documentation -%>
+  #   <%= param.documentation %>
```
mullermp: What happens when documentations have newline characters in them? Most of the AWS docs have this. General thought for all of documentation rendering. We probably want to always iterate documentation (this method should split by "\n" and return an array?) and if ends up empty, the "if" case is handled, and if it is a single or multiline string, it's handled.

gems/smithy/lib/smithy/anvil/client/templates/plugins/endpoint.erb
```diff
@@ -33,7 +34,7 @@ module <%= namespace %>
       # @api private
       class Handler < Smithy::Client::Handler
         def call(context)
-          params = EndpointParameters.create(context.config)
+          params = EndpointParameters.create(context.config, context.operation_name, context.params)
```
mullermp: This is a private method right? Why not just take "context" here and have the method unpack these?

gems/smithy/lib/smithy/anvil/client/templates/specs/endpoint_provider_spec.erb
```diff
@@ -6,6 +6,19 @@ require_relative '../spec_helper'
 
 module <%= namespace %>
   describe EndpointProvider do
+    # TODO: Can be replaced by stub_responses config once implemented
+    let(:stub_send) do
+      Class.new(Smithy::Client::Plugin) do
+        def add_handlers(handlers, _config)
+          handlers.add(Class.new(Smithy::Client::Handler) do
```
mullermp: HandlerBuilder has a handle method that accepts a block already, so you can skip the add_handlers call.

mullermp (approved):

projections/weather/lib/weather/endpoint_parameters.rb
alextwoods: This is from our own endpoint weld - I'll update the documentation.

gems/smithy/lib/smithy/anvil/client/templates/endpoint_parameters.erb
alextwoods: I think all existing Endpoint parameter docs are single line, but I dont *think* theres smithy level validation for that, so I'll add handling.

gems/smithy/lib/smithy/anvil/client/templates/endpoint_parameters.erb
alextwoods: Blank lines are required for yard to render correctly (and in fact after testing, I needed to add more)

gems/smithy/lib/smithy/anvil/client/templates/plugins/endpoint.erb
alextwoods: Its an api private method.  I guess I can just take context, but I had wanted to avoid coupling the endpoint parameters to context directly.

alextwoods: I've updated it to context.

gems/smithy/lib/smithy/anvil/client/templates/specs/endpoint_provider_spec.erb
alextwoods: Updated.

gems/smithy/lib/smithy/templates/client/endpoint_parameters.erb
```diff
@@ -4,7 +4,17 @@
 
 module <%= namespace %>
   # Endpoint parameters used to resolve endpoints per request.
-  <%= documentation %>
+<% parameters.each do |param| -%>
+  # @!attribute <%= param.name %>
+<% if param.documentation -%>
+<% param.documentation.split("\n").each do |line| -%>
```
mullermp: I was suggesting that param.documentation should return an array of strings, then you can just iterate it here, without having to check if param.documentation exists?

gems/smithy/lib/smithy/views/client/endpoint_provider_spec.rb
```diff
@@ -61,6 +64,116 @@ def expect_error?
             !@expect['error'].nil?
           end
         end
+
+        # OperationInput test
```
mullermp: Self explanatory .. can be api private for both of these classes..

mullermp (approved):

gems/smithy/lib/smithy/views/client/endpoint_provider_spec.rb
```diff
@@ -61,6 +64,116 @@ def expect_error?
...
+              transform_list(shape, value)
+            when 'map'
+              transform_map(shape, value)
+            else
+              value
+            end
+          end
+
+          def transform_map(shape, value)
+            member_shape = @model['shapes'][shape['value']['target']]
```
mullermp, line 122: Should be able to use the shape resolver after my PR

projections/weather/lib/weather/client.rb
```diff
@@ -98,7 +98,7 @@ def get_forecast(params = {}, options = {})
     #       }
     #     ]
     #   }
-    #
+    #
```
mullermp: Should be able to rebuild after my PR to remove diffs here.

gems/smithy/lib/smithy/anvil/client/templates/endpoint_parameters.erb
mullermp: Huh really? Maybe that's true for attribute but it shouldn't be for other tags.

mullermp: In v3 I don't see spaces between attribute and return.

PR #245, opened 2025-01-09
Update shapes spec to next instead of skip

jterapin:
Instead of skipping unrelated shape specs, we are `next`-ing to not crowd the spec outputs.

### DIFF
--- a/gems/smithy/spec/interfaces/client/shapes_spec.rb
+++ b/gems/smithy/spec/interfaces/client/shapes_spec.rb
@@ -20,15 +20,15 @@ describe 'Component: Shapes' do
     expect(subject).to be_a(Module)
   end
 
-  context 'generated shapes' do
+  context 'shapes' do
     shape_tests['shapes'].each do |id, shape|
-      context 'a generated shape' do
+      context "a generated shape: #{id}" do
         next if %w[operation service].include?(shape['type'])
 
         let(:shape_name) { Smithy::Vise::Shape.relative_id(id) }
         let(:generated_shape) { Object.const_get("#{subject}::#{shape_name}") }
 
-        it 'is a shape of expected shape type and has an id' do
+        it 'is of a expected shape type and has an id' do
           expected_shape_type =
             Object.const_get("#{shapes_module}::#{shape['type'].camelize}Shape")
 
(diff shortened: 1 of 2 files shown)


gems/smithy/spec/interfaces/client/shapes_spec.rb
```diff
@@ -20,45 +20,47 @@
...
           expect(generated_shape).to be_a(expected_shape_type)
           expect(generated_shape.id).to eq(id)
         end
 
-        it 'has a type variation of the shape when applicable' do
-          unless %w[structure union].include?(shape['type'])
-            skip("Test does not expect the generated #{id} to have a type")
-          end
+        context 'type' do
+          next unless %w[structure union].include?(shape['type'])
```
alextwoods: general style thought - rather than always defining this context, maybe use an if outside of it?  IE:

```ruby
if %w[structure union].include?(shape['type'])
  context 'type' do
    it 'is set on the shape' .... 
  end
end
```

gems/smithy/spec/interfaces/client/shapes_spec.rb
```diff
@@ -90,11 +92,9 @@
         end
 
         context 'key and value members' do
-          it 'are shapes of expected member names, shapes and contains traits when applicable' do
-            if shape['key'].nil? && shape['value'].nil?
-              skip("Test does not expect the generated #{id} to have a key/value members")
-            end
+          next if shape['key'].nil? && shape['value'].nil?
 
+          it 'are shapes of expected member names, shapes and contains traits when applicable' do
```
alextwoods: not really related to this PR, but this name reads a bit weirdly.  what about `it 'has correct key and value members '` or something like that

jterapin: yeah I definitely think my wording here could be better overall - let me make some adjustments

gems/smithy/spec/interfaces/client/shapes_spec.rb
mullermp: Yes you would have to do that above too. Otherwise you get an empty context section.

mullermp (approved):
This is fine, but it would be preferred to have explicit behavior tests rather than dynamic tests over the model. (Dynamic tests are fine with explicit input/output).

PR #246, opened 2025-01-10
Model resolver

mullermp:
Add a small model resolver that handles prelude shapes.


gems/smithy/lib/smithy/model.rb
```diff
@@ -9,28 +9,38 @@
...
-      smithy.api#PrimitiveFloat
-      smithy.api#PrimitiveInteger
-      smithy.api#PrimitiveLong
-      smithy.api#PrimitiveShort
-      smithy.api#Short
-      smithy.api#String
-      smithy.api#Timestamp
-      smithy.api#Unit
-    ].freeze
+    PRELUDE_SHAPES = {
```
alextwoods, line 14: I know you've mentioned before that you were worried about smithy prelude shapes changing... Could we have this be a dynamic list (json smithy ast?) that we pull from a specific version of smithy?  Maybe we check in a file and it references the smithy version it was generated from?

gems/smithy/lib/smithy/model.rb
```diff
@@ -9,28 +9,38 @@
...
+      'smithy.api#PrimitiveInteger' => { 'type' => 'integer', 'traits' => { 'smithy.api#default' => 0 } },
+      'smithy.api#PrimitiveLong' => { 'type' => 'long', 'traits' => { 'smithy.api#default' => 0 } },
+      'smithy.api#PrimitiveShort' => { 'type' => 'short', 'traits' => { 'smithy.api#default' => 0 } },
+      'smithy.api#Short' => { 'type' => 'short' },
+      'smithy.api#String' => { 'type' => 'string' },
+      'smithy.api#Timestamp' => { 'type' => 'timestamp' },
+      'smithy.api#Unit' => { 'type' => 'structure', 'members' => {}, 'traits' => { 'smithy.api#unitType' => {} } }
+    }.freeze
+
+    def self.shape(model, target)
```
alextwoods, line 41: I'm finding this a little weird - I think its a little weird to have a module `shape` method that takes a model rather than a model class with a `shape` method.  It seems like we want a class for model instead of just a Hash.

gems/smithy/lib/smithy/model.rb
mullermp: Yeah, that is a worry, but I think it's less of a worry if this is in code generation as a synthetic shape rather than added to the model. I don't know if we want to make this dynamic by shelling out to smithy ast (also, how? I can't get that to work without a smithy build json). I can add a comment on what version was used.

gems/smithy/lib/smithy/model.rb
mullermp: It would be an instance that we would have to create in many places and only provide a value. It doesn't require any state. This is similar to Api.shape in v3 code generation.

gems/smithy/lib/smithy/model.rb
alextwoods: Why is it less of a worry in code generation than as synthetic shapes? The behavior during code generation is the same.

jterapin: If we decide to include a dynamic way to create prelude shapes in the codegen-side of things,  just a note to pull prelude shapes definitions in smithy-client gem

gems/smithy/lib/smithy/model.rb
alextwoods: This to me is still a code smell that using Hash is the wrong way to represent model if we then need static methods to actually use it.

But I'll disagree and commit on it.

gems/smithy/lib/smithy/model.rb
mullermp: It's because we may or may not care about prelude shapes for some parts of generation, and that distinction is whether we inspect the model or not. The model can include prelude shapes if explicitly given, and that check happens first before checking prelude shapes fallback. This actually gives us an out if the prelude shapes are updated before code generation.

alextwoods: Why would we ever not want prelude shapes?  (I guess the only answer is a model which does not use them all, and in that case, they would get shaken out when using the service index to get the service's shapes closure anyway)

mullermp: Yeah. I think the key though is that we are not writing the model with an old copy of prelude shapes. We prefer the new ones in the model if provided. If not given, we have an older copy, and then an additional decision point of using the resolver or inspecting shapes in the model. It gives us flexibility on all levels.

alextwoods: I still dont see how thats different than a weld that adds them to the model if they don't exist.  

It gives us flexility => it also gives you a foot gun.  You need to know which one to use.

But anyway, I'll disagree and commit on this as well.

alextwoods (approved):

gems/smithy/lib/smithy/model.rb
mullermp: We tried wrapping the model with Vise and it didn't work out too well from the preprocess standpoint. Even if the model was wrapped, trying to fetch the shape would be the same result, either it would be another hash/object or failure.

gems/smithy/lib/smithy/model.rb
mullermp: Adding it to weld if it doesn't exist would work but then we have to worry about ordering of weld (pre process wouldn't be guaranteed across all welds). I think this is a good compromise.

PR #248, opened 2025-01-14
Smithy client rbs

alextwoods:
*Issue #, if available:*

*Description of changes:*


Rakefile
```diff
@@ -47,4 +47,22 @@ namespace 'smithy-client' do
...
+    sh('bundle exec rbs -I gems/smithy-client/sig validate')
+  end
+
+  task 'rbs:test' do
+    env = {
+      'RUBYOPT' => '-r bundler/setup -r rbs/test/setup',
+      'RBS_TEST_RAISE' => 'true',
+
+      'RBS_TEST_OPT' => '-I gems/smithy-client/sig',
+      'RBS_TEST_TARGET' => '"Smithy,Smithy::*,Smithy::Client,Smithy::Client::*"'
```
mullermp: Does Smithy::* not already capture those?

gems/smithy-client/lib/smithy-client/base.rb
```diff
@@ -23,10 +23,11 @@ def initialize(plugins, options)
         after_initialize(plugins)
       end
 
-      # @return [Configuration<Struct>]
+      # @return [Struct]
       attr_reader :config
 
       # @return [HandlerList]
+      # @api private
```
mullermp: I think this is and should be public.

gems/smithy-client/lib/smithy-client/configuration.rb
```diff
@@ -48,6 +48,7 @@ module Client
     #     configuration.build!(price: 100)
     #     #=> raises an ArgumentError, :price was not added as an option
     #
+    # @api private
```
mullermp: I think this is public also.

gems/smithy-client/sig/smithy-client/base.rbs
```diff
@@ -0,0 +1,8 @@
+module Smithy
+  module Client
+    # Base class for all service clients.
+    class Base
+      attr_reader config: Struct[untyped]
```
mullermp, line 5: Just as a general comment I think it's entirely reasonable to run rbs prototype on all the public classes and we can fill them in as we go.

Rakefile
alextwoods: I've been playing around with this - I tried every possible combo.  I was able to remove `Smithy::Client::*` and still get the correct errors.  But if I remove any others then it does not work.  I don't understand why, but maybe you can try as well.

gems/smithy-client/lib/smithy-client/base.rb
alextwoods: I'm not really sure we want to expose the handler list as public - I'd generally prefer to keep the public surface area as small as possible to start and if theres a reason or if its useful later, we can open it up.

gems/smithy-client/lib/smithy-client/configuration.rb
alextwoods: I dont think it needs to be - users never interact with Configuration

gems/smithy-client/lib/smithy-client/base.rb
mullermp: I think it's public to anyone writing plugins or abstractions. We use it to remove handlers in presigning for example.

gems/smithy-client/lib/smithy-client/configuration.rb
mullermp: It is public I think. I believe option() is a shortcut but add_options and after client initialization hooks interact with this?

Rakefile
mullermp: I'm sure it's fine. Weird that it doesn't include children modules.

gems/smithy-client/lib/smithy-client/base.rb
alextwoods: As far as I can tell, only the Configuration::DefaultResolver is actually exposed when developing plugins (through the default block on an option), but I'll remove the api private from it.

gems/smithy-client/lib/smithy-client/configuration.rb
alextwoods: As far as I can tell, only the Configuration::DefaultResolver is actually exposed when developing plugins (through the default block on an option), but I'll remove the api private from it.

After initialize only gets an instance of the client, which has the anonymous class for config (ie a resolved config). Before initialize gets the options Hash.

And `add_options` is not a class method - its an instance method that shouldn't get used by plugin developers I think?

mullermp: I believe it lets you conditionally define options based upon if config has an option already. Actually I think this might be quite useful for exposing secondary http config options if http plugin options exist. We probably should have been using this more in v3 development.

alextwoods: Ah I see - you mean when defining a Plugin, you can override the `add_options` method which will get [called during client initialization](https://github.com/smithy-lang/smithy-ruby/blob/decaf/gems/smithy-client/lib/smithy-client/base.rb#L64)?

This generally seems like a pretty dangerous way to add options and very order dependent.

mullermp: The intent seems to be that plugins can implement that method and define new config options dynamically which would make this config class public. I'm not sure about ordering. I suppose that makes sense if config is added based on config. It could be based on ruby environment or other components which would not be order dependent.

rbs_collection.yaml
```diff
@@ -0,0 +1,18 @@
...
+
+gems:
+  - name: stringio
+    ignore: false
+  - name: tempfile
+    ignore: false
+  - name: uri
+    ignore: false
+  - name: digest
+    ignore: false
```
jterapin: nit - alphabetize?

jterapin: Also - should we include manifest.yaml for future distribution? ([context](https://dev.to/pocke/rbs-introduced-manifestyaml-2idc))

gems/smithy-client/sig/smithy-client/configuration.rbs
```diff
@@ -0,0 +1,15 @@
+module Smithy
+  module Client
+
+    class Configuration
+      def initialize: () -> void
+      def add_option: (Symbol, ?untyped) ?{ (DefaultResolver) -> untyped } -> self
+      def build!: (?::Hash[untyped, untyped] options) -> Struct[untyped]
+
+      class DefaultResolver
```
jterapin, line 9: Ahh, so even if `DefaultResolver` is a private class - we are still defining this bit here due to the way `add_option` is defined above? Just wanted to clarify

gems/smithy-client/lib/smithy-client/block_io.rb
```diff
@@ -19,6 +19,7 @@ def initialize(headers = nil, &block)
       # @return [Integer]
       def write(chunk)
         @block.call(chunk, @headers)
+        chunk.bytesize
```
mullermp, line 22: I think adding this fixes an issue I ran into before in the read() method below. I don't remember what that was but it was related to return types.

gems/smithy-client/sig/smithy-client/block_io.rbs
```diff
@@ -0,0 +1,13 @@
+module Smithy
+  module Client
+    class BlockIO
+
```
mullermp, line 4: What happens if you run rubocop against rbs files? Extra spacing in a few files.

gems/smithy-client/spec/smithy-client/handler_builder_spec.rb
```diff
@@ -41,7 +41,7 @@ def initialize
           expect(handler_class.new.call('context')).to eq('context')
         end
 
-        it 'accepts a step with the block' do
+        it 'accepts a step with the block', rbs_test: :skip do
```
mullermp, line 44: What fails here?

gems/smithy-client/spec/smithy-client/handler_list_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 module Smithy
   module Client
-    describe HandlerList do
+    describe HandlerList, rbs_test: :skip do
```
mullermp, line 5: What fails with these?

mullermp (approved):

gems/smithy-client/sig/smithy-client/http/request.rbs
```diff
@@ -0,0 +1,18 @@
...
+  module Client
+    module HTTP
+      class Request
+
+        def initialize: (?Hash[Symbol, untyped]) -> void
+
+        attr_reader endpoint: endpoint_url?
+        attr_accessor http_method: String
+        attr_reader headers: Headers
+        attr_reader body:  _ReadableIO?
```
jterapin, line 11: since we have stringio defined in our rbs collection - we could use it here

gems/smithy-client/sig/smithy-client/http/request.rbs
```diff
@@ -0,0 +1,18 @@
...
+
+        def body=: (_ReadableIO | String | nil) -> void
```
jterapin, line 13: Same here as above

gems/smithy-client/spec/smithy-client/handler_list_spec.rb
alextwoods: Most tests fail since they try to add strings as handlers, eg: `subject.add('validate', step: :validate)`.

Rather than trying to rewrite every one of these many tests, I thought skipping would be easiest for now.

gems/smithy-client/spec/smithy-client/handler_builder_spec.rb
alextwoods:
```
       TypeError: [Smithy::Client::Handler#call] ArgumentTypeError: expected `::Smithy::Client::HandlerContext` but given `[:validate, :build, :sign, :send]`, [Smithy::Client::Handler#call] ReturnTypeError: expected `::Smithy::Client::Output` but returns `[:validate, :build, :sign, :send]`
```

I wasn't sure how to easily modify this test, so just decided to skip it.

gems/smithy-client/sig/smithy-client/block_io.rbs
alextwoods: Can you run rubocop on rbs files? I wasn't able to get it working.

gems/smithy-client/lib/smithy-client/block_io.rb
alextwoods: Yeah, this was returning the result of calling the block before which was incorrect.

jterapin (approved):
I noticed some .rbs files did not have some public methods defined. 

I'm thinking there's a boundary of whether something is useful to define vs not. 

Is there a set of RBS guidelines that we could agree as a team? (like, no need to define private methods, etc. if the input could be anything, set it untyped vs Object)

gems/smithy-client/sig/smithy-client/block_io.rbs
mullermp: I've never tried. That's what I was asking.

gems/smithy-client/lib/smithy-client/block_io.rb
mullermp: I think the read() code below can be simplified then.

gems/smithy-client/spec/smithy-client/handler_list_spec.rb
mullermp: Ah ok. I'm fine with skipping these. Is there an easy way to make rbs_test work like pending?

gems/smithy-client/spec/smithy-client/handler_builder_spec.rb
mullermp: The last handler needs to return an Output.new(context) and then the test can check output.context. (change all the resp -> output)

alextwoods: All of the context's need to be correctly typed, so we would need to create a HandlerContext and use the metadata for this.  I'm not sure its worth it, but can be done later.

rbs_collection.yaml
alextwoods: Good call on both!

gems/smithy-client/sig/smithy-client/http/request.rbs
alextwoods: The ReadbleIO interface is actually I think more correct here since we do allow any "io-like" object that supports read.

gems/smithy-client/sig/smithy-client/http/request.rbs
alextwoods: Ditto here on the ReadbleIO interface.  It *also* has a weird case where it will convert a String to a StringIO for you (and String ofc does not meet the ReadableIO requirements, so needs to be added explicitly to the type signature).

alextwoods:
> I'm thinking there's a boundary of whether something is useful to define vs not.

Yeah, you are right, some public methods are not covered with signatures.  I tried to use some general judgement in getting a first pass of signatures working and useful.  I find fully untyped signatures (ie, whats created by rbs prototype) to be useful - It doesn't add any value to define methods without at least some non-untyped signature, so I avoided doing that.  

We can/should keep adding to these signatures as we develop!

PR #249, opened 2025-01-14
Remove instance plugins

mullermp:
Removes instance level plugins. This simplifies the configuration experience by only allowing plugin adding and removing at the class level. Instance plugins would have to similarly support remove_plugins but also there was a nasty edge case about recursion where plugins could add other plugins, which would go away.

### DIFF
--- a/gems/smithy/lib/smithy/templates/client/endpoint_provider_spec.erb
+++ b/gems/smithy/lib/smithy/templates/client/endpoint_provider_spec.erb
@@ -39,11 +39,11 @@ module <%= namespace %>
 
 <% test_case.operation_inputs.each do |operation_input| -%>
       it 'produces the correct output from the client when calling <%= operation_input.operation_name %>' do
+        Client.add_plugin(stub_send)
         client = Client.new(
 <% operation_input.client_params.each do |p| -%>
           <%= p.param %>: <%= p.value %>,
 <% end -%>
-           plugins: [stub_send]
         )
 <% if test_case.expect_error? -%>
         expect do
--- a/gems/smithy/lib/smithy/templates/client/endpoint_provider_spec.erb
+++ b/gems/smithy/lib/smithy/templates/client/endpoint_provider_spec.erb
@@ -67,6 +67,7 @@ module <%= namespace %>
         expected['endpoint'].fetch('headers', {}).each do |k,v|
           expect(resp.context.request.headers[k]).to eq(v)
         end
+        Client.remove_plugin(stub_send)
 
         # TODO: expect auth
 <% end -%>
--- a/gems/smithy-client/lib/smithy-client/base.rb
+++ b/gems/smithy-client/lib/smithy-client/base.rb
@@ -59,7 +59,6 @@ module Smithy
       def build_config(plugins, options)
         config = Configuration.new
         config.add_option(:schema)
-        config.add_option(:plugins)
         plugins.each do |plugin|
           plugin.add_options(config) if plugin.respond_to?(:add_options)
         end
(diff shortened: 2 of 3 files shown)


alextwoods (commented):
We support instance plugins in V3 so this would be a breaking change from V3.  I also thing instance plugins are a great feature so I'd rather see us support it and allow instance level removal of plugins.

alextwoods:
What is the recursive edge case with plugins adding plugins that you mention?

Plugins can implement the `before_initialize` method which can modify `options`, but the list of plugins is already determined before that, so adding plugins to that list doesn't change anything.  

However,  plugin's `before_initialize`  is *also* called with the client class and could technically call `add_plugin` on that class.  However, again, the list of plugins for that particular instance has already been built, so it wouldn't affect it. It would affect future instances of the class, which seems pretty bad - should we remove the passing of client class to `before_initialize`? Are there any cases where we need that class?

mullermp:
I highly doubt anyone is using instance plugins. We added it 7 months ago and there are no known use cases. I think instance plugins complicate a lot of use cases and I also think it's completely unnecessary. When would you want your configuration option and handler to only apply in one instance of a client?

Instance plugins can add other plugins to options or to the class in before initialize. That is where the complexity is. In v3 we try to do this but it's actually not even working. I think before initialize is necessary as is for the global config case for example.

mullermp:
I would be more concerned about cutting the global config case, before initialize case, or class level plugins, than I would for instance plugins.

gems/smithy/lib/smithy/templates/client/endpoint_provider_spec.erb
```diff
@@ -39,11 +39,11 @@ module <%= namespace %>
 
 <% test_case.operation_inputs.each do |operation_input| -%>
       it 'produces the correct output from the client when calling <%= operation_input.operation_name %>' do
+        Client.add_plugin(stub_send)
```
alextwoods, line 42: This doesn't look correct - this will break behavior for any other specs trying to use the client

mullermp: Everything currently passes, but I agree, I would like to remove this when stub responses is implemented.

mullermp:
I think if we made a compatibility gem we could easily have a plugin that looks for plugins as an option and adds them to the client in a before initialize.

gems/smithy/lib/smithy/templates/client/endpoint_provider_spec.erb
```diff
@@ -67,6 +67,7 @@ module <%= namespace %>
         expected['endpoint'].fetch('headers', {}).each do |k,v|
           expect(resp.context.request.headers[k]).to eq(v)
         end
+        Client.remove_plugin(stub_send)
```
alextwoods, line 70: This isn't thread safe - clients created between the add and removal will get the stub_send.

alextwoods:
> Instance plugins can add other plugins to options or to the class in before initialize

I'm not quite sure they can currently.  In `new` we do:
```
plugins = build_plugins(self.plugins + options.fetch(:plugins, []))
plugins = before_initialize(plugins, options)
```

And before_initialize:
```
        def before_initialize(plugins, options)
          plugins.each do |plugin|
            plugin.before_initialize(self, options) if plugin.respond_to?(:before_initialize)
          end
        end
```

Plugins can modify the options or the class, but the plugin list has already been resolved, so they won't be used.

The most dangerous part of this is that a plugin could modify the class by adding a new plugin to the class.  That plugin would NOT be used for this instance, but would be used in future instances.  That behavior is unrelated to instance plugins.

gems/smithy/lib/smithy/templates/client/endpoint_provider_spec.erb
mullermp: I don't think rspec runs in threads.

mullermp:
I am referring to https://github.com/aws/aws-sdk-ruby/blob/e079ced8f91c34935b0fa760f4efc59e5a265832/gems/aws-sdk-core/lib/seahorse/client/base.rb#L215 which I left out.

alextwoods (approved):
I still disagree with this direction - I think instance plugins are a powerful and useful user feature, but will disagree and commit.

PR #250, opened 2025-01-15
Improve errors and error documentation

mullermp:
Adds documentation to error classes and methods, and improves error messaging which previously didn't work correctly. Also implements retryable and throttling on errors.


gems/smithy/spec/support/be_in_documentation_matcher.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+require 'rdoc'
+
+RSpec::Matchers.define :be_in_documentation do |file, klass, method|
```
alextwoods, line 5: Nice

alextwoods (approved):

PR #251, opened 2025-01-15
Improve endpoint specs + endpoint fixture models

alextwoods:
*Description of changes:*
Adds specs for getAttr - these cases cannot be fully covered with the non-AWS endpoint standard libs, so its useful to have coverage in the specs in smithy-client.  Additionally, fix rubocop autocorrect issue with attr name.

A few minor fixes in endpoint fixture models:
* fix version mismatches (reset all to 1.0)
* sync the latest string-array tests
* add an endpoint test case to the getAttr model.

Adds a rake task for syncing the fixture smithy model files to the json format.  This can be used in the future to create a fixture validation CI task (but requires smithy cli to be installed).


Rakefile
```diff
@@ -38,6 +38,14 @@ namespace :smithy do
   end
 
   task 'spec' => %w[spec:unit spec:endpoints]
+
+  desc 'Convert all fixture smithy models to JSON AST representation'
+  task 'sync-fixtures' do
```
mullermp, line 43: I'm fine with sync, but we could consider a validation task instead (one that doesn't overwrite but just checks)

gems/smithy-client/lib/smithy-client/endpoint_rules.rb
```diff
@@ -105,7 +105,9 @@ def self.attr(value, path)
         if parts.size == 1
           val
         else
-          attr_reader(val, parts.slice(1..-1).join('.'))
+          # rubocop:disable Style/Attr
```
mullermp: You can single line comment this and not need to disable/enable but only disable.

mullermp (approved):

PR #252, opened 2025-01-15
Implement Cbor Codec

jterapin:
Implementation of Cbor Codec based on caffeinated smithy-ruby, includes test updates to the shapes component in the smithy gem.

**Future ToDos:**
* support cbor specific traits
* support union shapes in cbor 
* support enum shapes in cbor
* support event streams and events in cbor
* add json, xml and query codecs


mullermp:
Cbor -> CBOR please

jterapin:
> Cbor -> CBOR please

Yeah I was going to ask about that since Cbor namespace format in V3 and Hearth were both different. I can advise.

gems/smithy-client/lib/smithy-client/cbor_codec.rb
```diff
@@ -0,0 +1,99 @@
+# frozen_string_literal: true
+
+require 'base64'
+
+module Smithy
+  module Client
+    # @api private
+    class CBORCodec
```
alextwoods: Worth a little discussion - While the aws query compatible *trait* is an AWS specific trait, a `query_compatible` configuration option here is *generic* and as far as codec is concerned, it just means that lists/maps with the [xmlFlattened trait](https://smithy.io/2.0/spec/protocol-traits.html#xmlflattened-trait) (which is generic, not aws specific) we parse an empty list/map instead of nil.

So - I *think* we could build this behavior in to the codec.  However - I'm also happy to punt that to later - we don't need to implement that now.

alextwoods (commented):
Nice! This is looking good so far.

gems/smithy-client/lib/smithy-client/cbor_codec.rb
```diff
@@ -0,0 +1,112 @@
...
+module Smithy
+  module Client
+    # Codec that serializes and deserializes in CBOR format.
+    class CBORCodec
+      include Shapes
+
+      # @option options [Boolean] :xml_flattened (nil)
+      def initialize(options = {})
+        # TODO: Need to add implementation to handle this trait
+        @xml_flattened = options[:xml_flattened] || false
```
alextwoods: I'd leave the `TODO` in place, but I think remove the actual option here..... I think we also would want the name to be some thing other than xml_flattened.   Either `query_compatible` or maybe more specifically on the behavior, which is `default_empty_flattened` or something like that.

jterapin:
Reverted changes relating to JSON Codec since we want the CBOR Codec to move forward for the protocol implementation.

.rubocop.yml
```diff
@@ -10,15 +10,31 @@ AllCops:
...
+Metrics/CyclomaticComplexity:
+  Exclude:
+    - 'gems/smithy-client/lib/smithy-client/cbor/encoder.rb'
+    - 'gems/smithy-client/lib/smithy-client/cbor/decoder.rb'
+    - 'gems/smithy-client/spec/smithy-client/cbor_spec.rb'
 
 Metrics/MethodLength:
   Max: 20
+  Exclude:
+    - 'gems/smithy-client/spec/smithy-client/cbor_spec.rb'
```
mullermp: Specs should already be excluded? If not you should do this generally (see BlockLength)

gems/smithy-client/lib/smithy-client.rb
```diff
@@ -44,6 +44,14 @@
 # endpoints
 require_relative 'smithy-client/endpoint_rules'
 
+# codecs
+require_relative 'smithy-client/codec/cbor'
+
+# cbor
+require_relative 'smithy-client/cbor'
+require_relative 'smithy-client/cbor/encoder'
```
mullermp: cbor is already requiring these. Either remove these or remove them from cbor.rb

gems/smithy-client/lib/smithy-client/cbor/decoder.rb
```diff
@@ -0,0 +1,317 @@
+# frozen_string_literal: true
```
mullermp, line 1: Is there a summary of what changed between this and v3/v4?

gems/smithy-client/lib/smithy-client/cbor/encoder.rb
```diff
@@ -0,0 +1,255 @@
+# frozen_string_literal: true
+
+require 'bigdecimal'
+
+module Smithy
+  module Client
+    module CBOR
+      # Pure ruby implementation of CBOR encoder.
+      class Encoder
```
mullermp, line 9: Ditto on summary if at all?

gems/smithy-client/lib/smithy-client/codec/cbor.rb
```diff
@@ -0,0 +1,114 @@
+# frozen_string_literal: true
+
+require 'base64'
+
+module Smithy
+  module Client
+    module Codec
```
mullermp: module Codecs plural imo (would have to fix directory)

gems/smithy-client/lib/smithy-client/codec/cbor.rb
```diff
@@ -0,0 +1,114 @@
...
+      # Codec that serializes and deserializes in CBOR format.
+      # TODO:
+      #   * Support (de)serializing union shapes once union is supported
+      #   * Support handling of typed documents when it is supported
+      #   * Update implementation to handle event streams
+      #   * Handle query_compatible trait
+      class CBOR
+        include Shapes
+
+        def initialize(options = {}); end
```
mullermp: Remove if this is not doing anything?

gems/smithy-client/lib/smithy-client/codec/cbor.rb
```diff
@@ -0,0 +1,114 @@
...
+          values.each_pair.with_object({}) do |(key, value), data|
+            if shape.member?(key) && !value.nil?
+              member = shape.member(key)
+              data[member.name] = format_data(value, member.shape)
+            end
+          end
+        end
+
+        def parse_data(value, shape, type = nil)
+          if value.nil?
```
mullermp: guard clause

gems/smithy-client/lib/smithy-client/shapes.rb
```diff
@@ -68,12 +68,12 @@ def initialize(options = {})
...
         end
 
-        # @return [Hash<String, MemberShape>]
+        # @return [Hash<Symbol, MemberShape>]
         attr_accessor :members
 
         # @return [MemberShape]
         def add_member(name, shape, traits)
-          @members[name] = MemberShape.new(name, shape, traits)
+          @members[name.to_sym] = MemberShape.new(name, shape, traits)
```
mullermp: I'm currently fixing this, but the fix should be in code gen. 

```
          def add_member_method(shape)
            case shape
            when 'ListShape'
              "set_member(#{@shape}, #{@traits})"
            when 'MapShape'
              "set_#{@name}(#{@shape}, #{@traits})"
            else
              "add_member(:#{@name}, #{@shape}, #{@traits})"
            end
          end
```

gems/smithy-client/lib/smithy-client/shapes.rb
```diff
@@ -147,7 +147,17 @@ def initialize(options = {})
...
```
mullermp: Ditto, no need to convert to symbol if you fix code gen.

I have these methods also. I don't think you need to raise (it can return nil just like a hash)

```
        # @return [Boolean]
        def member?(name)
          @members.key?(name)
        end

        # @return [MemberShape, nil]
        def member(name)
          @members[name]
        end
```

gems/smithy-client/sig/smithy-client/shapes.rbs
```diff
@@ -60,9 +60,11 @@ module Smithy
       end
 
       class StructureShape < Shape
-        attr_accessor members: Hash[String, MemberShape]
+        attr_accessor members: Hash[Symbol, MemberShape]
         attr_accessor type: untyped
         def add_member: (String, Shape, Hash[String, untyped]?) -> void
+        def member?: (String | Symbol) -> bool
```
mullermp:
```
        def member?: (Symbol) -> bool
        def member: (Symbol) -> MemberShape?
```

gems/smithy-client/spec/smithy-client/cbor/encoder_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+          end
+
+          it 'encodes arrays' do
+            expect(cbor64_encode([])).to eq('gA==')
+            expect(cbor64_encode([1, 2, 3])).to eq('gwECAw==')
+            expect(cbor64_encode([1, [2, 3], [4, 5]])).to eq('gwGCAgOCBAU=')
+          end
+
+          it 'encodes maps' do
+            expect(cbor64_encode({ 'a' => 1,
```
mullermp: formatting borked

gems/smithy-client/spec/smithy-client/cbor/encoder_spec.rb
```diff
@@ -0,0 +1,110 @@
+# frozen_string_literal: true
+
+module Smithy
```
mullermp, line 3: Are these specs all the same?

gems/smithy-client/spec/smithy-client/codec/cbor_spec.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    module Codec
+      describe CBOR do
+        subject { CBOR.new }
```
mullermp: subject is already CBOR.new

gems/smithy-client/spec/smithy-client/codec/cbor_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+        let(:string_shape) { Shapes::StringShape.new(id: 'string') }
+
+        let(:list_shape) do
+          shape = Shapes::ListShape.new(id: 'list')
```
mullermp: Don't have to make changes now, but you can consider yielding self on all of the complex shapes so that constructing can be like:
```
Shapes::ListShape.new(id: 'list') do |list|
  list.set_member(...)
end
```

gems/smithy-client/spec/smithy-client/shapes_spec.rb
```diff
@@ -100,7 +100,7 @@ module Shapes
           it 'adds a member' do
             member_name = 'FOO'
             subject.add_member(member_name, StringShape.new, {})
-            expect(subject.members[member_name]).to be_kind_of(MemberShape)
+            expect(subject.members[member_name.to_sym]).to be_kind_of(MemberShape)
```
mullermp: Won't need this change

gems/smithy/spec/fixtures/shapes/model.json
```diff
@@ -1,6 +1,21 @@
 {
     "smithy": "2.0",
     "shapes": {
+        "smithy.ruby.tests#BigDecimal": {
```
mullermp, line 4: Consider using the AllShapes fixture for your test?

gems/smithy/spec/interfaces/client/shapes_spec.rb
```diff
@@ -20,81 +19,174 @@
...
+        expect(subject::List.member).to be_a(shapes_module::MemberShape)
+        expect(subject::List.member.shape).to be_a(shapes_module::StringShape)
+        expect(subject::List.member.name).to eq('member')
+        expect(subject::List.member.shape.id).to eq(expected_member['target'])
+      end
+    end
+
+    context 'map' do
+      let(:expected_shape) do
+        fixture['shapes'].select { |_, s| s['type'] == 'map' }
```
mullermp: nit, carry over chaining to next line
```
fixture['shapes']
  .select ..
  .thing
```

gems/smithy/spec/interfaces/client/shapes_spec.rb
```diff
@@ -20,81 +19,174 @@
     expect(subject).to be_a(Module)
   end
 
-  context 'shapes' do
-    shape_tests['shapes'].each do |id, shape|
-      next if %w[operation service].include?(shape['type'])
+  context 'generated shapes' do
```
mullermp, line 22: nice

gems/smithy-client/spec/smithy-client/codec/cbor_spec.rb
alextwoods: Yeah, I do like that and I think we already do that for ServiceShape? 

But generally, wouldn't be useful for the way we generate shapes.rb since we need to generate all shape objects first so they can be referenced when setting members.  (Well unless we changed the approach to do basically a topological sort.  But that seems overkill).

gems/smithy-client/lib/smithy-client.rb
alextwoods: Yeah - I dont have real preference on this for now, but long term I think this behavior depends on what we do for engines.

gems/smithy-client/lib/smithy-client/codec/cbor.rb
alextwoods: Yeah, I guess we have `Plugins` so should follow plural for these.

gems/smithy-client/lib/smithy-client/codec/cbor.rb
alextwoods: I'd say we should keep it - we know we'll need to add an option for handling flattened lists/maps.

.rubocop.yml
jterapin: Yeah, it wasn't but I'll update this.

gems/smithy-client/lib/smithy-client.rb
jterapin: For now - I'll have cbor module require the encoder/decoder.

gems/smithy-client/lib/smithy-client/codec/cbor.rb
mullermp: That can be added when it's added?

gems/smithy-client/spec/smithy-client/codec/cbor_spec.rb
mullermp: Yeah, it's just useful for tests and one-off construction in implementations.

gems/smithy-client/lib/smithy-client/cbor/decoder.rb
jterapin: Sure.
* V3 implementation had some raises such as [this example](https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-core/lib/aws-sdk-core/cbor/decoder.rb#L126-L127) but V4 ignores it. The bulk of cbor decoder and encoder is from V4 implementation. I made changes to appease the rubocop violations.
  * After chatting with Alex, I went by V4 implementation.  
* Cbor codec are heavily based on V3 since caffeinated smithy-ruby had builders and parsers.
* Added `codecs/cbor.rb` which was something not covered in V3.

gems/smithy-client/lib/smithy-client/cbor/encoder.rb
jterapin: See comment above under decoder.

gems/smithy-client/lib/smithy-client/codec/cbor.rb
jterapin: Yeah I'll plan to add this in once I get a firmer idea on what needs to be done with JSONCodec (has more traits that influence the shape serde behaviors). I would like to leave it in. ( I will circle back once protocol is done)

gems/smithy-client/lib/smithy-client/shapes.rb
jterapin: Ahh, I was wondering what was going on - that makes sense.

gems/smithy-client/lib/smithy-client/shapes.rb
jterapin: Thanks. I'll copy that over to stay aligned.

gems/smithy-client/spec/smithy-client/cbor/encoder_spec.rb
jterapin: Yup, these are pretty copy-pasta.

gems/smithy-client/spec/smithy-client/codec/cbor_spec.rb
jterapin: Yeah, we currently do that for `ServiceShape` and `OperationShape`. I agree with Alex that it may not be useful since we need to define all the shapes prior to member adding. We could always change this up.

gems/smithy-client/spec/smithy-client/shapes_spec.rb
jterapin: Thanks, fixed up specs.

gems/smithy/spec/fixtures/shapes/model.json
jterapin: I did consider that but noticed that simple shapes defined within AllShapes fixture will reference prelude shapes. However, I'm open to updating to use AllShapes fixture (if we'd like that, I can include that in my next change)

alextwoods (approved):
Nice - lgtm.

gems/smithy/spec/interfaces/client/shapes_spec.rb
```diff
@@ -20,81 +19,175 @@
...
+        expect(subject::IntEnum.members[:baz].shape).to be_a(shapes_module::StructureShape)
+        expect(subject::IntEnum.members[:baz].name).to eq('baz')
+        expect(subject::IntEnum.members[:baz].traits).to eq(expected_member['traits'])
+        expect(subject::IntEnum.members[:baz].shape.id).to eq(expected_member['target'])
+      end
+    end
+
+    context 'list' do
+      let(:expected_member) do
+        fixture['shapes'].select { |_, s| s['type'] == 'list' }
```
mullermp: Please fix for all others (thought that went without saying..)

gems/smithy-client/lib/smithy-client/shapes.rb
```diff
@@ -68,12 +68,12 @@ def initialize(options = {})
...
         end
 
-        # @return [Hash<String, MemberShape>]
+        # @return [Hash<Symbol, MemberShape>]
         attr_accessor :members
 
         # @return [MemberShape]
         def add_member(name, shape, traits)
-          @members[name] = MemberShape.new(name, shape, traits)
+          @members[name] = MemberShape.new(name.to_s, shape, traits)
```
mullermp: You should not need to convert symbol to string for this to work.

gems/smithy/spec/interfaces/client/shapes_spec.rb
jterapin: My bad - missed that, will add now

gems/smithy-client/lib/smithy-client/shapes.rb
jterapin: Per our discussion - previously, i was thinking that we need a way to reference the `name` back but thinking a bit further, this is not necessary. Might be useful if its a unique member shape id or something but that's not quite it so moved to remove these. (always could go back to add them IF we need them)

PR #254, opened 2025-01-16
Plugin loading and documentation

mullermp:
Loads plugins and documentation.


gems/smithy-client/lib/smithy-client.rb
```diff
@@ -31,7 +31,6 @@
 
 require_relative 'smithy-client/plugins/logging'
 require_relative 'smithy-client/plugins/raise_response_errors'
-require_relative 'smithy-client/plugins/net_http'
```
alextwoods, line 34: I think it would simplify a few places if we followed an autoload strategy for these instead of not requiring it at all here.  I get not wanting to load a potentially optional plugin, but I think using autoload gives us the best of both (and then means we don't have to worry about requiring it in many other places).

gems/smithy/lib/smithy.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'rails/railtie'
+require 'smithy-client'
```
alextwoods, line 4: Should we require smithy client only when we are generating a client?

gems/smithy/lib/smithy/templates/client/client.erb
```diff
@@ -2,16 +2,23 @@
...
     self.schema = Shapes::SCHEMA
 
-<% plugins.each do |p| -%>
-    add_plugin(<%= p %>)
+<% add_plugins.each do |plugin_class| -%>
+    add_plugin(<%= plugin_class %>)
 <% end -%>
 
-    # TODO!
+<% docstrings.each do |docstring| -%>
```
alextwoods, line 19: should this be named like `initialize_docstrings` or something like that since there are also operation ones?

gems/smithy/lib/smithy/views/client/plugin.rb
```diff
@@ -0,0 +1,61 @@
...
+          @relative_path = options.fetch(:relative_path, false)
+          @requirable = options.fetch(:requirable, false)
+        end
+
+        attr_reader :class_name, :require_path
+
+        def docstrings
+          docstrings = []
+          options = Object.const_get(@class_name).options
+          options.each do |option|
```
alextwoods, line 21: In V3 we sometime used "undocumented" options - we may want to support that here as well and do a check to see if the option is documented.

projections/weather/lib/weather/client.rb
```diff
@@ -2,15 +2,87 @@
...
+  # See {#initialize} for a full list of supported configuration options.
   class Client < Smithy::Client::Base
     self.schema = Shapes::SCHEMA
 
     add_plugin(Smithy::Client::Plugins::NetHTTP)
-    add_plugin(Plugins::Endpoint)
+    add_plugin(Weather::Plugins::Endpoint)
 
-    # TODO!
+    # @param [Hash] options
```
alextwoods, line 16: nice!

gems/smithy/lib/smithy/weld.rb
```diff
@@ -33,16 +33,26 @@ def post_process(artifacts)
...
 
-    # Called when constructing a plan
+    # Called when constructing the endpoint provider. Any bindings defined here will
+    # be merged with other function bindings.
     # @return [Hash<String, String>] endpoint function bindings for use in endpoint rules.
     def endpoint_function_bindings
       {}
     end
+
+    # Called when constructing the client. Any plugins defined here will be merged
```
alextwoods, line 50: we'll also want the ability to add dependencies to the gemspec, but no need to put it in now.

gems/smithy/lib/smithy/views/client/plugin_list.rb
```diff
@@ -0,0 +1,84 @@
...
+            "#{namespace}::Plugins::Endpoint" => "#{gem_dir}/lib/#{gem_name}/plugins/endpoint.rb"
+          }
+          plugins.map do |class_name, require_path|
+            Plugin.new(class_name: class_name, require_path: require_path)
+          end
+        end
+
+        def transport_plugins(protocol)
+          plugins = {
+            'http' => { 'Smithy::Client::Plugins::NetHTTP' => 'smithy-client/plugins/net_http' }
```
alextwoods, line 66: This feels like it should be a constant somewhere.... and long term should maybe be extensible by welds.  But I think fine to leave as a constant for now.

gems/smithy/lib/smithy/views/client/plugin_list.rb
```diff
@@ -0,0 +1,84 @@
...
+          plugins.map do |class_name, require_path|
+            Plugin.new(class_name: class_name, require_path: require_path, relative_path: true, requirable: true)
+          end
+        end
+
+        def define_namespaces
+          parent = Object
+          namespace.split('::') do |mod|
+            child = mod
+            parent.const_set(child, ::Module.new) unless parent.const_defined?(child)
```
alextwoods, line 77: Is there a possible issue where we define something as a module that should be a class?

eg: `Module1::Class1::MyPlugin1` - we would define `Class1` as a module instead of a class.

I think for codegen...this shouldn't matter and I dont think it would break anything.

gems/smithy-client/lib/smithy-client.rb
mullermp: Yes I think we should go back and do autoloading of clients in a separate PR.

gems/smithy/lib/smithy.rb
mullermp: We will have to require it for types too, and since we have no server, I think this is fine for now. Same for rails/railtie, probably only for server, but active_support for all.

gems/smithy/lib/smithy/templates/client/client.erb
mullermp: I could - but operation.docstrings is already what is called, and docstrings under client seems sufficient for now?

gems/smithy/lib/smithy/views/client/plugin_list.rb
mullermp: Yeah, that is a limitation and would blow up. I'm not sure how else how to handle this. We could consider making documentation a post process? So after the gem is built, load the whole thing?

gems/smithy/lib/smithy/views/client/plugin_list.rb
mullermp: Transport would be configured somehow and index into this. I think protocol would work the same way. I agree it's fine for now..

gems/smithy/lib/smithy/weld.rb
mullermp: Yes, we need a way to add to the gemspec. We could even create a gemspec object and pass it to weld, then add to it directly, then write it out.

gems/smithy/lib/smithy/views/client/plugin.rb
mullermp: Fixed

gems/smithy/lib/smithy/weld.rb
alextwoods: oh yeah, a gemspec object would be awesome for that! but we can figure that out when we get there w/ AWS I think.

gems/smithy/lib/smithy/views/client/plugin_list.rb
alextwoods: I think its unlikely an issue - this would apply only for generated plugins, which we fully control and I think in general we won't/shouldn't have many generated plugins.  I think its also rare that you would want a plugin class to be defined in the namespace of another class instead of a module, so I'm not sure how much we should try to work around this.

alextwoods (approved):

PR #255, opened 2025-01-17
Generated RBS

alextwoods:
*Description of changes:*
Adds generation of all RBS signatures for: client, types, errors, shapes, endpoint provider and endpoint parameters.

Add a new CI and rake task for testing generated specs - enables spy tests for our adhoc generated fixture gems during normal unit test execution.  In the future we run validation and spy tests on gems generated for protocol tests and endpoint specs.

Type optionality: For members on our structs, they must always be optional regardless of the rules for [structure member optionality](https://smithy.io/2.0/spec/aggregate-types.html#structure-member-optionality) because of the way we build/parse (probably).


gems/smithy/lib/smithy/model.rb
```diff
@@ -35,16 +35,56 @@ module Model
...
+          model['shapes'][target]
+        elsif PRELUDE_SHAPES.key?(target)
+          PRELUDE_SHAPES[target]
+        else
+          raise ArgumentError, "Shape not found: #{target}"
+        end
+      end
+
+      # rubocop:disable Metrics/CyclomaticComplexity
+      def rbs_type(model, id, shape)
```
mullermp: I like this logic, but I'm debating if it belongs here. I understand that we probably have to reuse this for client.rbs? What else?

Worst case, I think Model can have an rbs type or even a doc type method, but maybe the logic belongs in a private class (similar to how I implemented the indexes) and can also drop the rbs / type prefix/suffix and act more like a visitor (not exactly a visitor pattern but this is simple).

alextwoods: Yeah, I wasn't really sure where to put this - I do think we'll need to re-use it, at least in client.rbs (and maybe other places? like paginators/waiters?).  I could leave it in types_rbs.rb for now and when we get to re-using it, refactor it out.

What do you mean by acting more like a visitor? You mean a class that defines methods for each smithy type effectively + a method that delegates to each of those?

Rakefile
```diff
@@ -63,6 +63,15 @@ namespace :smithy do
       raise 'Fixture models are out of sync.  Run bundle exec rake smithy:sync-fixtures to correct'
     end
   end
+
+  task 'rbs:unit' do
```
mullermp, line 87: Should we instead just hook up rbs to normal unit test runs or could that get annoying?

Rakefile
```diff
@@ -63,6 +63,15 @@ namespace :smithy do
...
+    env = {
+      'SMITHY_RUBY_RBS_TEST' => 'true'
+    }
+    sh(env, 'bundle exec rake smithy:spec:unit')
```
mullermp, line 91: sh seems wrong. Won't we be executing bundle exec already? Maybe this is a task invoke instead?

gems/smithy/lib/smithy/model.rb
mullermp: I think regardless of whether this method lives in Model or not, the rbs logic should live somewhere else, either in a private class (like the parsers) and callable here, or a public class. Maybe a class called RBS instead of Model with a self.type(model, id, shape) method?

gems/smithy/lib/smithy/views/client/module_rbs.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
```
mullermp, line 1: Haven't put much thought into this, but we could create a separate sig/lib namespace under views if you want. Since views is more flat than before it's bearable. Otherwise we can do what you did here.

gems/smithy/lib/smithy/views/client/types_rbs.rb
```diff
@@ -0,0 +1,63 @@
...
+          @types ||= Model::ServiceIndex
+                     .new(@model)
+                     .shapes_for(@plan.service)
+                     .select { |_key, shape| %w[structure union].include?(shape['type']) }
+                     .map { |id, structure| Type.new(id, structure, @model) }
+        end
+
+        # @api private
+        class Type
+          def initialize(id, structure, model)
```
mullermp: Super ridiculous nit and feel free to ignore but put model first (convention with other classes)

gems/smithy/lib/smithy/views/client/types_rbs.rb
```diff
@@ -0,0 +1,63 @@
...
+            @model = model
+          end
+
+          def name
+            Model::Shape.name(@id)
+          end
+
+          def members
+            @structure['members'].map do |name, member|
+              Member.new(name, member['target'], Model.shape(@model, member['target']), @model)
```
mullermp: I think you can resolve the shape inside the member class instead. I would just pass model, name, and member target.

gems/smithy/lib/smithy/views/client/types_rbs.rb
```diff
@@ -0,0 +1,63 @@
...
+            end
+          end
+        end
+
+        # @api private
+        class Member
+          def initialize(name, id, target, model)
```
mullermp: Id and target naming are different than the class that calls this. I think it should be name, target, shape?

gems/smithy/spec/spec_helper.rb
```diff
@@ -47,6 +51,18 @@ def cleanup(module_names, tmpdir)
       Object.send(:remove_const, module_names.first)
     end
 
+    def load_rbs_environment(sdk_dir, load_collection: true)
+      require 'rbs'
+
+      loader = RBS::EnvironmentLoader.new(core_root: RBS::EnvironmentLoader::DEFAULT_CORE_ROOT)
+      loader.add(path: Pathname(File.join(__dir__, '../../smithy-client/sig')))
```
mullermp: Dir to_s

mullermp (approved):

gems/smithy/lib/smithy/model.rb
alextwoods: Yeah, that makes sense, I moved it into an RBS class.

Rakefile
alextwoods: What do you mean?  Like whenever we run specs we run with RBS?

It does take a fair bit longer (a total of around 30 seconds for me) to run these tests - I think I'd generally prefer to keep them separate so that when I'm running unit tests normally (not working on signatures) my test will be fast.

Rakefile
alextwoods: I used `sh` because I'm setting ENV - I don't want the ENV settings to apply to anything else (granted, we currently aren't chaining this, but I could see us doing something like a `check` task that runs everything).

gems/smithy/lib/smithy/views/client/module_rbs.rb
alextwoods: I think for now I'll leave rbs flat - theres a fairly limited number of them total and I think keeping the namespace flatter is nice.  But we can move/re-evaluate if it does get clunky.

gems/smithy/lib/smithy/views/client/types_rbs.rb
alextwoods: yea, good call.  Fixed.

gems/smithy/lib/smithy/views/client/types_rbs.rb
alextwoods: Yeah this is a good call - cleans it up a little.

gems/smithy/lib/smithy/views/client/types_rbs.rb
alextwoods: cleaned up based on above.

Rakefile
mullermp: Then either I think we can remove the bundle exec or if you do task invoke you can add an ensure to clear this env?

gems/smithy/spec/spec_helper.rb
```diff
@@ -79,6 +97,64 @@ def smith(plan)
...
+
+      spy_modules = [modules.join('::'), 'Smithy::Client']
+      spy_classes = []
+      spy_modules.each do |module_name|
+        spy_classes += classes_to_spy(Object.const_get(module_name), env)
+      end
+
+      spy_classes.each do |spy_class|
+        tester.install!(spy_class, sample_size: RBS::Test::SetupHelper::DEFAULT_SAMPLE_SIZE,
+                                   unchecked_classes: unchecked_classes)
```
jterapin: Could do some alignment here for the args.

gems/smithy/spec/spec_helper.rb
```diff
@@ -79,6 +97,64 @@ def smith(plan)
         with_captured_stdout { Smithy.smith(plan) }
       end
     end
+
+    def load_collection(loader)
+      collection_config_path = RBS::Collection::Config.find_config_path
+      lock_path = RBS::Collection::Config.to_lockfile_path(collection_config_path)
+      if lock_path.file?
+        lock = RBS::Collection::Config::Lockfile.from_lockfile(lockfile_path: lock_path,
+                                                               data: YAML.load_file(lock_path.to_s))
```
jterapin: Alignment here.

gems/smithy-client/sig/smithy-client/errors.rbs
```diff
@@ -2,7 +2,7 @@ module Smithy
   module Client
     module Errors
       class ServiceError < RuntimeError
-        def initialize: (HandlerContext, String?, untyped?) -> void
+        def initialize: (HandlerContext?, String?, untyped?) -> void
```
jterapin, line 5: May want to update the documentation for this - not clear that these are supposed to be optional,

jterapin (commented):
I was thinking that maybe you could include a README regarding running the RBS tests (whether separately or together - kind of like how we have the other tests commands)

Rakefile
mullermp: Talked offline, sh seems correct here, since we need to unset ENV anyway, and it should run in a separate process.

Rakefile
mullermp: Talked offline. This happens because each time we generate an SDK, we have to inject spies. It would be efficient to generate all SDKs for tests up front then do this, and it would lower time, but that's outside of this scope. Running rbs outside of regular tests is sufficient.

alextwoods: Yeah - we could change our testing strategy to generate all fixture SDKs up front.  This has some tradeoffs though.  Running a single test still requires generating all SDKs.  But, we gain some benefit when running all tests as no SDKs need to be generated multiple times for different tests that use them.  It does mean we would need to be more careful about the namespaces uses by those tests - currently the sdk is generated, loaded, tested, then removed (by after blocks), so only tests in given scope have access to it.

I did some investigation into maintaining a dynamic RBS Environment and adding to it - around 80% of the time taken in the spy test construction is in loading the ENV and most of that time is from loading smithy::client, stdlib and collection signatures.  However, theres not straightforward ways to dynamically add signatures to an environment.

gems/smithy-client/sig/smithy-client/errors.rbs
alextwoods: The documentation I don't think currently indicates its required?  In general we have a pretty weak story on nullability - almost everything, everywhere is nullable :-(

README.md
```diff
@@ -38,5 +38,10 @@ Build a fixture
 ```
 bundle exec smithy-ruby smith client --gem-name fixture --gem-version 1.0.0 <<< $(cat gems/smithy/spec/fixtures/endpoints/default-values/model.json)
+```
 
+Running RBS validations and tests:
```
mullermp, line 43: I think eventually the README should be a readme. `rake -T` already accomplishes this assuming we've added descriptions to these rake tasks.

Rakefile
```diff
@@ -63,6 +82,19 @@ namespace :smithy do
...
+  end
+
+  task 'rbs:endpoints' do
+    task('smithy:spec:endpoints').invoke('rbs_test')
```
mullermp, line 96: Does the asme approach as spec:unit not work here? Setting of rbs env and calling spec endpoints? Since spec helper is called by the task, spec helper can be responsible for inserting. I think it should be the same speed?

gems/smithy-client/sig/smithy-client/endpoint_rules.rbs
```diff
@@ -2,10 +2,10 @@ module Smithy
   module Client
     module EndpointRules
       class Endpoint
-        def initialize: (uri: String, ?properties: ::Hash[String, untyped], ?headers: ::Hash[String, String]) -> void
+        def initialize: (uri: String, ?properties: ::Hash[String, untyped], ?headers: ::Hash[String, String|Array[String]]) -> void
         attr_accessor uri: String
         attr_accessor properties: Hash[String, untyped]
-        attr_accessor headers: Hash[String, String]
+        attr_accessor headers: Hash[String, String|Array[String]]
```
mullermp, line 8: Can this really be an array of strings for headers? That seems wrong..

gems/smithy-client/sig/smithy-client/shapes.rbs
```diff
@@ -2,7 +2,7 @@ module Smithy
   module Client
     module Shapes
       class Shape
-        def initialize: (?Hash[Symbol, untyped]) -> void
+        def initialize: (?Hash[Symbol, untyped]) ?{ (self) -> untyped } -> void
```
mullermp: I think you can make untyped as Shape here since it yields self or does that not work?

gems/smithy/lib/smithy/model.rb
```diff
@@ -35,16 +36,18 @@ module Model
...
-    # @param [String] target Target shape
-    # @return [Hash] The shape
-    def self.shape(model, target)
-      if model['shapes'].key?(target)
-        model['shapes'][target]
-      elsif PRELUDE_SHAPES.key?(target)
-        PRELUDE_SHAPES[target]
-      else
-        raise ArgumentError, "Shape not found: #{target}"
+    class << self
```
mullermp: Can you revert this back to self.shape?

gems/smithy/lib/smithy/model/rbs.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Model
+    # @api private
+    class Rbs
```
mullermp: Rbs or RBS as convention? I'm ok with either. I lean towards RBS. I think views it has to be Yml or Rbs.

gems/smithy/lib/smithy/views/client/client_rbs.rb
```diff
@@ -0,0 +1,118 @@
...
+        def namespace
+          Util::Namespace.namespace_from_gem_name(@plan.options[:gem_name])
+        end
+
+        def option_types
+          # TODO: Ensure this order is correct when plugins override options
+          @plugins
+            .map(&:options)
+            .flatten
+            .uniq(&:name)
```
mullermp: I think to_h will effectively do uniq

gems/smithy/spec/smithy/model/service_index_spec.rb
```diff
@@ -44,7 +44,7 @@ module Model
           it 'handles recursive shapes' do
             service =
               fixture['shapes']
-              .select { |id, _| id == 'smithy.ruby.tests#Recursive' }
+              .slice('smithy.ruby.tests#Recursive')
```
mullermp: nice

gems/smithy/spec/spec_helper.rb
```diff
@@ -23,18 +23,26 @@ class << self
...
       plan = create_plan(modules, model, type, options)
+      sdk_dir = plan.options[:destination_root]
       smith(plan)
 
-      $LOAD_PATH << ("#{plan.options[:destination_root]}/lib")
+      $LOAD_PATH << ("#{sdk_dir}/lib")
       require "#{plan.options[:gem_name]}#{type == :types ? '-types' : ''}"
-      plan.options[:destination_root]
+
+      setup_rbs_spytest(modules, sdk_dir) if options.fetch(:rbs_test, ENV.fetch('SMITHY_RUBY_RBS_TEST', nil))
```
mullermp, line 32: I think the environment variable can be the flag. I don't think we need rbs_test as an actual option (we would never call this in specs right?)

gems/smithy/spec/spec_helper.rb
```diff
@@ -23,18 +23,26 @@ class << self
...
+      sdk_dir
+    rescue StandardError => e
```
mullermp, line 34: Does the after{} block not already do this? When does this error happen?

gems/smithy/lib/smithy/views/client/endpoint_parameter.rb
```diff
@@ -25,15 +25,25 @@ def docstrings
         end
 
         def documentation_type
-          case @data['type']
-          when 'stringArray' then 'Array<String>'
+          case @data['type'].downcase
+          when 'stringarray' then 'Array<String>'
```
mullermp, line 29: I'd prefer not downcasing if we know the casing (these are straight from the endpoints specification and not user input right?)

mullermp (approved):

Rakefile
alextwoods: The issue is that we need to construct `RBS_TEST_OPTS` that include paths to every generated sig directory (which we only know inside the task) and `RBS_TEST_TARGET` (which we only know inside the task).  I couldn't find a clean way to break down that task other than adding an argument to it.

gems/smithy-client/sig/smithy-client/endpoint_rules.rbs
alextwoods: Yes - think of Signing region set which is a list of strings.

gems/smithy-client/sig/smithy-client/shapes.rbs
alextwoods: Tne `untyped` here is the return of the provided block.  The `self` is already what we yield. 

I could change the untyped to `void` to indicate that the return value isn't *used* by our code here.

gems/smithy/lib/smithy/model.rb
alextwoods: Sure.

gems/smithy/lib/smithy/model/rbs.rb
alextwoods: I've used `Rbs` everywhere so far because to me `ClientRbs` reads slightly better than `ClientRBS`.  But not a strong opinion and I'm happy to change all of them.

gems/smithy/spec/spec_helper.rb
alextwoods: When an exception is raised, the method does not return a value, meaning that the `@tmpdir` in specs does not get set.  The `after` block does get called, but it gets called with `nil` for directory.  This ensures that the directory is cleaned up correctly.

These errors can happen for a few reasons:
1. errors during code generation.
2. errors requiring generated code.
3. invalid rbs (causes loading the rbs env to fail).

With the changes I've made we now get nice clean error messages/stack traces and no tmp dirs are left around!

gems/smithy/lib/smithy/views/client/endpoint_parameter.rb
alextwoods: We made a mistake in smithy endpoints by allowing any casing for these :-(

gems/smithy/lib/smithy/model/rbs.rb
mullermp: I think RBS for this module and Rbs for views is fine

alextwoods: Updated to RBS

mullermp:
@ksss @soutaro if either of you are interested in reviewing RBS and RBS testing strategy for generic generated clients

PR #256, opened 2025-01-21
Parameter Validation

mullermp:
Adds parameter validation based on API rules.


gems/smithy-client/lib/smithy-client.rb
```diff
@@ -40,13 +41,11 @@
...
-
-# endpoints
-require_relative 'smithy-client/endpoint_rules'
+require_relative 'smithy-client/param_validator'
 
 module Smithy
   # Base module for a generated Smithy gem.
   module Client
-    VERSION = File.read(File.expand_path('../VERSION', __dir__)).strip
+    VERSION = File.read(File.expand_path('../VERSION', __dir__.to_s)).strip
```
alextwoods: Whats this change for?

gems/smithy-client/lib/smithy-client/param_validator.rb
```diff
@@ -0,0 +1,190 @@
...
+            shape(member_shape.shape, value, errors, context + "[#{name.inspect}]")
+          else
+            errors << "unexpected value at #{context}[#{name.inspect}]"
+          end
+        end
+      end
+
+      def list(shape, values, errors, context)
+        # ensure the value is an array
+        unless values.is_a?(Array)
```
alextwoods, line 90: Is this the validation we do in V3? Technically all we really need is an enumerable probably

gems/smithy-client/lib/smithy-client/param_validator.rb
```diff
@@ -0,0 +1,190 @@
...
+        values.each.with_index do |value, index|
+          next unless value
+
+          shape(shape.member.shape, value, errors, context + "[#{index}]")
+        end
+      end
+
+      def map(shape, values, errors, context)
+        unless values.is_a?(Hash)
+          errors << expected_got(context, 'a hash', values)
```
alextwoods: casing nit, hash -> Hash.

gems/smithy-client/lib/smithy-client/param_validator.rb
```diff
@@ -0,0 +1,190 @@
...
+        # validate members
+        values.each.with_index do |value, index|
+          next unless value
+
+          shape(shape.member.shape, value, errors, context + "[#{index}]")
+        end
+      end
+
+      def map(shape, values, errors, context)
+        unless values.is_a?(Hash)
```
alextwoods, line 103: Ditto here - I think we could technically work with anything (ie a struct) that responds to `[]`

gems/smithy-client/lib/smithy-client/param_validator.rb
```diff
@@ -0,0 +1,190 @@
...
+        when StringShape
+          errors << expected_got(context, 'a String', value) unless value.is_a?(String)
+        when IntegerShape
+          errors << expected_got(context, 'an Integer', value) unless value.is_a?(Integer)
+        when FloatShape
+          errors << expected_got(context, 'a Float', value) unless value.is_a?(Float)
+        when TimestampShape
+          errors << expected_got(context, 'a Time object', value) unless value.is_a?(Time)
+        when BooleanShape
+          errors << expected_got(context, 'true or false', value) unless [true, false].include?(value)
```
alextwoods, line 153: What about boolish/truthy?

alextwoods (commented):
Overall approach makes sense and looks good.

gems/smithy-client/lib/smithy-client.rb
mullermp: Apparently dir can be nil so I have been calling to_s on it always.

gems/smithy-client/lib/smithy-client/param_validator.rb
mullermp: A struct is enumerable.

gems/smithy-client/lib/smithy-client/param_validator.rb
mullermp: I don't think we should allow a struct for maps. We don't even generate structs for maps. I think the majority of cases this would just be wrong and likely an issue with composing the request (maybe I forgot a key)

gems/smithy-client/lib/smithy-client/param_validator.rb
mullermp: But truthy things would not be serialized as true or false?

I think generally the responds_to approach is not correct here but except for blob where we have to check things they can read.

gems/smithy-client/lib/smithy-client/shapes.rb
```diff
@@ -91,7 +91,30 @@ def member(name)
       class IntegerShape < Shape; end
 
       # Represents an IntEnum shape.
-      class IntEnumShape < EnumShape; end
+      class IntEnumShape < Shape
```
mullermp, line 94: Inheriting from EnumShape and StructureShape makes is_a? checks order dependent for visiting. While this is duplicating, it's safer.

gems/smithy/lib/smithy/views/client/shapes.rb
```diff
@@ -117,17 +117,17 @@ def filter_traits(traits)
...
 
           SHAPE_TYPES_MAP[type]
         end
 
-        def shape_type_from_id(id)
+        def shape_name_from_id(id)
           return PRELUDE_SHAPES_MAP[id] if PRELUDE_SHAPES_MAP.key?(id)
 
-          Model::Shape.relative_id(id)
+          Model::Shape.name(id).camelize
```
mullermp, line 130: This fixes a bug where constants in add_member are not camelized and do not exist.

gems/smithy/spec/support/client_helper.rb
```diff
@@ -4,95 +4,18 @@
 
 module ClientHelper
```
mullermp, line 5: I went back and forth on this but I'm happy with the result - I went too far with the refactor and dialed it back some. I think that the smithy spec_helper/client_helper should work only on fixtures. The purpose of those is to verify that the generated code/files are correct. I prefer the explicit module name declaration in the tests for smithy's spec_helper. I think having smithy-client rely on another gem's spec helper was funky, so I pulled out sample_shapes into smithy-clients client_helper, and I think it's ok to have two. At this point, when using this helper, we can assume the generated code is correct. The quick runtime client generation is great for smithy-client tests and does not need fixtures in my opinion, as we probably want to change model values during tests.

gems/smithy-client/lib/smithy-client/shapes.rb
alextwoods: If we wanted to reduce the duplicaiton, we could do use an `include` with some sort of ShapeWithMembers module or something.  But the duplication is minimal and fine.

gems/smithy-client/spec/smithy-client/param_validator_spec.rb
```diff
@@ -0,0 +1,317 @@
...
+          validate(enum: 'abc')
+          validate({ enum: 123 },
+                   'expected params[:enum] to be a String, got class Integer instead.')
+        end
+      end
+
+      describe 'floats' do
+        it 'accepts a float' do
+          validate(float: 123.0)
+          validate({ float: 123 },
```
alextwoods, line 165: This validation feels weird to me - it seems like we should accept an integer here?

gems/smithy-client/spec/support/client_helper.rb
```diff
@@ -43,54 +63,55 @@ def sample_shapes
...
       model = options[:model] ||= model(options)
-      namespace = options[:gem_module] || next_sample_module_name
-      plan = create_plan(model, namespace, options)
+      plan = create_plan(module_name, model, options)
       source = Smithy.source(plan)
       Object.module_eval(source)
-      Object.const_get(namespace)
-    rescue Exception => e # rubocop:disable Lint/RescueException
+      Object.const_get(module_name)
+    rescue LoadError => e
```
alextwoods, line 105: Does LoadError cover all of the syntax errors that could happen during eval?

alextwoods (approved):
Nice - looks good overall - just minor comments.

gems/smithy-client/spec/smithy-client/param_validator_spec.rb
mullermp: In v3 we do both param validation and param conversion. Param validation should ensure strictness-ish and param conversion can convert integers given to float shapes as floats. This makes error messaging cleaner maybe. Unless you have a different idea.

gems/smithy-client/spec/support/client_helper.rb
mullermp: I think so. At this point we assume generated code is valid.

PR #257, opened 2025-01-24
General clean ups to shape and schema

mullermp:
General clean ups.

Moved Prelude shapes in their own model so that Param validation can succeed.

### DIFF
--- a/gems/smithy-client/lib/smithy-client/shapes.rb
+++ b/gems/smithy-client/lib/smithy-client/shapes.rb
@@ -2,9 +2,9 @@
 
 module Smithy
   module Client
-    # Represents shape types from Smithy Model
+    # Represents shape types from the Smithy Model.
     module Shapes
-      # A base shape that all shapes inherits from
+      # A base shape that all shapes inherits from.
       class Shape
         def initialize(options = {})
           @id = options[:id]
@@ -18,7 +18,7 @@ module Smithy
         attr_accessor :traits
       end
 
-      # Represents a slim variation of the Service shape
+      # Represents a slim variation of the Service shape.
       class ServiceShape < Shape
         def initialize(options = {})
           super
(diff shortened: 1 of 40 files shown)


gems/smithy-client/lib/smithy-client/shapes.rb
```diff
@@ -181,51 +180,54 @@ def initialize(shape, traits)
...
-        traits: { 'smithy.api#default' => 0 }
-      )
-      Short = IntegerShape.new(id: 'smithy.api#Short')
-      String = StringShape.new(id: 'smithy.api#String')
-      Timestamp = TimestampShape.new(id: 'smithy.api#Timestamp')
-      Unit = StructureShape.new(
-        id: 'smithy.api#Unit',
-        traits: { 'smithy.api#unitType' => {} }
-      )
+      # Prelude shape definitions.
```
jterapin, line 183: Nice, thanks for doing this.

gems/smithy/lib/smithy/views/client/operation_examples.rb
```diff
@@ -99,7 +99,7 @@ def structure(json, shape, indent)
             entry = value(value, member_shape, "#{indent}  ")
             lines << "#{indent}  #{key.underscore}: #{format(entry)},"
           end
-          lines.last.chomp!(',')
+          lines.last.chomp!(',') if lines.last.end_with?(',')
```
alextwoods, line 102: Do you need `&.chomp`? can it ever be empty?  Also, why do you need to check if it ends_with?  chomp!, `'test'.chomp(',') just returns the same string`

alextwoods (approved):

gems/smithy/lib/smithy/views/client/operation_examples.rb
mullermp: If there are no members, lines is just '{' and it's a frozen string literal, so there is a failure.

jterapin (approved):

gems/smithy-client/lib/smithy-client/shapes.rb
mullermp: Yes. I did not consider the namespace issue.

PR #258, opened 2025-01-25
Refactor for source generation

mullermp:
Refactor of endpoints for source generation using client helper.

We should not support source only generation, so we must iterate source files and load into source (this improves runtime when testing API related changes). We use this testing in many places in aws-sdk-core. When we iterate source files, endpoint weld is not loaded, so instead we leave welds to be outside integration only and keep defaults within component generation.


mullermp:
Going to try a different approach by keeping welds

PR #259, opened 2025-01-28
Source only generation

mullermp:
This approach allows you to use Smithy.smith (or Smithy.generate/source, name TBD) to generate an in-memory SDK with only source files. This is advantageous both for testing of API without generating an entire artifact, or for replacing Aws.add_service in v2 and the hacked v3 wrapper.


gems/smithy/lib/smithy/generators/schema.rb
```diff
@@ -11,26 +11,47 @@ def initialize(plan)
...
         end
-        files
+      end
+
+      def source
+        source = source_files.map { |_file, content| content }.join("\n")
+        Object.module_eval(source)
+        Object.const_get(@plan.gem_namespace)
+      rescue Exception => e # rubocop:disable Lint/RescueException
+        puts "Error evaluating source:\n#{source}"
```
alextwoods: Wouldn't the source be large here?

gems/smithy/lib/smithy/generators/client.rb
```diff
@@ -11,41 +11,67 @@ def initialize(plan)
...
 
           create_file file, content
+          files << file
         end
-        files
+      end
+
+      def source
+        source = source_files.map { |_file, content| content }.join("\n")
+        Object.module_eval(source)
```
alextwoods: I would argue that generators should never be responsible for evaluating ruby code or getting the module.  Just generating the source.

I think something like the client helper should be responsible for evaluating code/returning the module.

gems/smithy/lib/smithy/generators/client.rb
```diff
@@ -11,41 +11,67 @@ def initialize(plan)
...
-      # rubocop:disable Metrics/AbcSize
-      def source_files
+      def gem_files
         Enumerator.new do |e|
           e.yield "#{@gem_name}.gemspec", Views::Client::Gemspec.new(@plan).render
           e.yield '.rubocop.yml', Views::Client::RubocopYml.new(@plan).render
 
-          e.yield "lib/#{@gem_name}.rb", Views::Client::Module.new(@plan).render
+          source_files.each { |file, content| e.yield file, content }
           e.yield "lib/#{@gem_name}/customizations.rb", Views::Client::Customizations.new.render
```
alextwoods, line 36: This feels a little weird to me as a special case - why isn't it considered a source file?

gems/smithy/lib/smithy/generators/schema.rb
```diff
@@ -11,26 +11,47 @@ def initialize(plan)
...
+        raise e
       end
 
       private
 
-      def source_files
+      def gem_files
         Enumerator.new do |e|
           e.yield "#{@gem_name}.gemspec", Views::Client::Gemspec.new(@plan).render
           e.yield '.rubocop.yml', Views::Client::RubocopYml.new(@plan).render
```
alextwoods, line 32: This isn't necessarily related to this change but why is the schema generator using client views?  Additionally - theres a lot of duplication between schema and client generators.

alextwoods (commented):
I think overall approach makes sense though I don't love the circular (test) dependency between smithy <-> smithy-client.  But I guess we implicitly have the same dependency in V3 right now with the build tools.

gems/smithy/lib/smithy/generators/schema.rb
mullermp: Possibly, but not as large as you would think, and I think the size is/will be the same with the current v3. If you're talking about the debugging, yeah, I think in a future change I plan to add a logger and this can be a debug (opt-in).

gems/smithy/lib/smithy/generators/client.rb
mullermp: Yeah, I went back and forth on this. On one hand it's nice to just be returned a module that is the loaded code, on the other, given a string, the user will have to load themselves (or us in the helper file). Let me think through the implications.

gems/smithy/lib/smithy/generators/client.rb
mullermp: I don't think customizations would be a source file because the file persists as a written file in the gem. There would be no way to load a customization if it were not from a file.

gems/smithy/lib/smithy/generators/schema.rb
mullermp: Originally I had separation of schema and client views, but they ended up being the same duplicated code. I think some duplication is ok (e.g. usages of these classes) but I think re-defining all of the views was not too necessary. The client and schema gems are effectively the same except the set of files to generate and what gets required. The generator base is responsible for setting up thor and destination file. Originally base shared more logic but with server generation, it doesn't seem the correct abstraction. I think it's correct for these classes to implement the interface of "generate" and "source"

gems/smithy/lib/smithy/generators/client.rb
mullermp: One downside of string only is we don't know the resolved namespace.

gems/smithy/lib/smithy/generators/client.rb
```diff
@@ -11,41 +11,67 @@ def initialize(plan)
...
+        gem_files.each_with_object([]) do |(file, content), files|
           next if file == "lib/#{@gem_name}/customizations.rb" && should_skip_customizations?
 
           create_file file, content
+          files << file
         end
-        files
+      end
+
+      def source
```
jterapin, line 23: I think it would be worthwhile to document what `generate` and `source` methods does somewhere, since it is touching different set of files (gem_files vs source_files)

gems/smithy/lib/smithy/generators.rb
```diff
@@ -16,5 +15,15 @@ def self.generate(plan)
         raise 'Not yet implemented'
       end
     end
+
+    def self.source(plan)
+      case plan.type
+      when :schema then Schema.new(plan).source
+      when :client then Client.new(plan).source
+      when :server then raise 'Will not be implemented'
```
alextwoods, line 23: If we will not implement it, why do we have this?

gems/smithy/lib/smithy/generators/client.rb
alextwoods: I guess I don't understand what a source file is then. I assume source => a ruby source file (ie, any file ending in .rb).

gems/smithy/lib/smithy/generators.rb
mullermp: This is for source only generation. I don't think we will write out an entire server implementation as a string. That may change depending on how that's implemented (maybe it's just a shim to include on rails app). The other condition is for unknown option.

alextwoods (approved):

gems/smithy/lib/smithy/generators/schema.rb
jterapin: When I think about the these views nested under client namespace, the organization is a bit weird - given that we are reusing them for schema generator (and probably server generator down the road). It could be worthwhile to make specific files such as `types.rb` and `shapes.rb` generic to all generators but I think at the time we were working on this - we were just focused on putting everything under `smithy-client`

Also these `types.rb` and `shapes.rb` has references to `Smithy::Client::...` which is a bit funky - now that we actually want to share the schema gem with early testers - maybe revisit the `smithy-core` gem?

gems/smithy/lib/smithy/generators/client.rb
mullermp: They are documented at their entry point in the Smithy module.

jterapin: Yeah, I saw that - might want to expand what is considered to be minimal Ruby code then. There appears to some confusion on what is considered a "source" file based on other comments.

gems/smithy-client/spec/support/client_helper.rb
```diff
@@ -0,0 +1,121 @@
...
+        }
+      }
+    end
+
+    def sample_client(options = {})
+      model = options[:model] ||= model(options)
+      namespace = options[:gem_module] || next_sample_module_name
+      plan = create_plan(model, namespace, options)
+      source = Smithy.source(plan)
+      Object.module_eval(source)
```
alextwoods, line 81: Since we are doing the eval on `Object` we are adding this to the global namespace.  We should add functions for cleaning up and managing these as well I think.

gems/smithy/lib/smithy/generators/client.rb
mullermp: It's source code. The source files are iterated just for their source. Do you have a better name?

gems/smithy/lib/smithy/generators/schema.rb
mullermp: Server generation would not use any of these I don't think. Schema generation is still a client-side view, just a subset of files. We could consider separate views for schema as we had before, but they will be the same files pretty much. I think the right boundary is here in the generator (separate generator to define the subset of files) and generate the same files.

Regarding types/shapes and smithy-client gem, since we control generation dependencies, we can easily swap this without concern, so I don't think it's necessary to do that right now.

gems/smithy-client/spec/support/client_helper.rb
mullermp: Sure

mullermp: As opposed to files, these will get garbage collected (next sample name will ensure they are unique). I will further refactor this and spec_helper to use something common but I will do it outside of this PR

alextwoods: I dont think garbage collection works on modules/classes that get defined?

mullermp: I will check, but I will also add a clean up when we have a test that uses the source. I will do some follow up changes.

PR #260, opened 2025-01-29
Implement Unions

alextwoods:
*Description of changes:*
Implements generation of tagged Unions following an approach similiar to Java/Hearth Smithy-Ruby: a "sealed classes" like approach with delegation.

Note: The to_h method needs to be generated on each member subclass because we don't want active-support as a dependnecy runtime.  If we did have it runtime, the Smithy::Client::Union class could define a to_h method that would do something like: `{}.tap { |h| h[self.class.name.split(::).last.underscore] = super(__getob__)`

The weather service in projections does not have any union types, so below is code generated for the `all_shapes` fixture which defines the following union:

```smithy
union Union {
    string: String,
    integer: Integer,
    structure: Structure,
    complexList: ComplexList,
    complexMap: ComplexMap
}
```

And generates the following in types.rb:
```rb
    class Union < Smithy::Client::Union

      class String < Union
        def to_h
          { string: super(__getobj__) }
        end
      end

      class Integer < Union
        def to_h
          { integer: super(__getobj__) }
        end
      end

      class Structure < Union
        def to_h
          { structure: super(__getobj__) }
        end
      end

      class ComplexList < Union
        def to_h
          { complex_list: super(__getobj__) }
        end
      end

      class ComplexMap < Union
        def to_h
          { complex_map: super(__getobj__) }
        end
      end

      class Unknown < Union
        def initialize(name:, value:)
          super({name: name || 'Unknown', value: value})
        end

        def to_h
          { unknown: super(__getobj__) }
        end
      end
    end
```

And generates the following in shapes.rb:
```ruby
    Union = UnionShape.new(id: 'smithy.ruby.tests#Union')

    Union.add_member(:string, Prelude::String, Types::Union::String)
    Union.add_member(:integer, Prelude::Integer, Types::Union::Integer)
    Union.add_member(:structure, Structure, Types::Union::Structure)
    Union.add_member(:complex_list, ComplexList, Types::Union::ComplexList)
    Union.add_member(:complex_map, ComplexMap, Types::Union::ComplexMap)
    Union.type = Types::Union
```

### DIFF
--- a/gems/smithy-client/lib/smithy-client/shapes.rb
+++ b/gems/smithy-client/lib/smithy-client/shapes.rb
@@ -138,10 +138,10 @@ module Smithy
           @type = nil
         end
 
-        # @return [Hash<String, MemberShape>]
+        # @return [Hash<Symbol, MemberShape>]
         attr_accessor :members
 
-        # @return [Struct]
+        # @return [Class]
         attr_accessor :type
 
         # @return [MemberShape]
@@ -164,7 +164,26 @@ module Smithy
       class TimestampShape < Shape; end
 
       # Represents both Union and EventStream shapes.
-      class UnionShape < StructureShape; end
+      class UnionShape < StructureShape
+        def initialize(options = {})
+          super
+          @member_types = {}
+        end
+
+        # @return [Symbol, Class]
+        attr_accessor :member_types
+
+        # @return [MemberShape]
+        def add_member(name, shape, type, traits: {})
+          @member_types[name] = type
+          super(name, shape, traits: traits)
+        end
+
+        # @return [Class, nil]
+        def member_type(name)
+          @member_types[name]
+        end
+      end
 
       # Represents a member shape.
       class MemberShape
--- a/gems/smithy-client/sig/smithy-client/union.rbs
+++ b/gems/smithy-client/sig/smithy-client/union.rbs
@@ -1,0 +1,7 @@
+module Smithy
+  module Client
+    class Union < ::SimpleDelegator
+      include Structure
+    end
+  end
+end
--- a/gems/smithy-client/spec/smithy-client/structure_spec.rb
+++ b/gems/smithy-client/spec/smithy-client/structure_spec.rb
@@ -17,9 +17,9 @@ module Smithy
         end
       end
 
-      let(:test_union_class) { Class.new(Union) }
+      let(:union_class) { Class.new(Union) }
       let(:struct_value_class) do
-        Class.new(test_union_class) do
+        Class.new(union_class) do
           def to_h
             { struct_value: super(__getobj__) }
           end
--- a/gems/smithy-client/spec/smithy-client/structure_spec.rb
+++ b/gems/smithy-client/spec/smithy-client/structure_spec.rb
@@ -10,9 +10,19 @@ module Smithy
           :hash_value,
           :value,
           :some_object,
+          :union_value,
           keyword_init: true
         ) do
           include Structure
+        end
+      end
+
+      let(:test_union_class) { Class.new(Union) }
+      let(:struct_value_class) do
+        Class.new(test_union_class) do
+          def to_h
+            { struct_value: super(__getobj__) }
+          end
         end
       end
 
@@ -24,6 +34,7 @@ module Smithy
             structure.new(value: 'bar')
           ],
           hash_value: { key: structure.new(value: 'value') },
+          union_value: struct_value_class.new(structure.new(value: 'value')),
           value: 'value',
           some_object: Object.new
         )
@@ -40,6 +51,7 @@ module Smithy
             hash_value: {
               key: { value: 'value' }
             },
+            union_value: { struct_value: { value: 'value' } },
             value: 'value',
             some_object: subject.some_object
           }
--- a/gems/smithy-client/spec/smithy-client/union_spec.rb
+++ b/gems/smithy-client/spec/smithy-client/union_spec.rb
@@ -3,9 +3,9 @@
 module Smithy
   module Client
     describe Union do
-      let(:test_union_class) { Class.new(Union) }
+      let(:union_class) { Class.new(Union) }
       let(:string_value_class) do
-        Class.new(test_union_class) do
+        Class.new(union_class) do
           def to_h
             { string_value: super(__getobj__) }
           end
--- a/gems/smithy-client/spec/smithy-client/union_spec.rb
+++ b/gems/smithy-client/spec/smithy-client/union_spec.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    class TestUnion < Union
+      class StringValue < TestUnion
+        def to_h
+          { string_value: super(__getobj__) }
+        end
+      end
+    end
+
+    describe Union do
+      subject { TestUnion::StringValue.new('union') }
+
+      it 'uses simple delegator and structure' do
+        expect(subject).to be_a(SimpleDelegator)
+        expect(subject).to be_a(Structure)
+      end
+
+      describe '#to_h' do
+        it 'serializes the value to a hash' do
+          expect(subject.to_h).to eq(string_value: 'union')
+        end
+      end
+
+      describe '#to_s' do
+        it 'returns a string representation' do
+          expect(subject.to_s)
+            .to eq('#<Smithy::Client::TestUnion::StringValue union>')
+        end
+      end
+    end
+  end
+end
--- a/gems/smithy-client/spec/smithy-client/union_spec.rb
+++ b/gems/smithy-client/spec/smithy-client/union_spec.rb
@@ -2,16 +2,22 @@
 
 module Smithy
   module Client
-    class TestUnion < Union
-      class StringValue < TestUnion
-        def to_h
-          { string_value: super(__getobj__) }
+    describe Union do
+      let(:test_union_class) { Class.new(Union) }
+      let(:string_value_class) do
+        Class.new(test_union_class) do
+          def to_h
+            { string_value: super(__getobj__) }
+          end
+
+          # anonymous class, need a class name to test to_s
+          def self.name
+            'TestUnion::StringValue'
+          end
         end
       end
-    end
 
-    describe Union do
-      subject { TestUnion::StringValue.new('union') }
+      subject { string_value_class.new('union') }
 
       it 'uses simple delegator and structure' do
         expect(subject).to be_a(SimpleDelegator)
@@ -27,7 +33,7 @@ module Smithy
       describe '#to_s' do
         it 'returns a string representation' do
           expect(subject.to_s)
-            .to eq('#<Smithy::Client::TestUnion::StringValue union>')
+            .to eq('#<TestUnion::StringValue union>')
         end
       end
     end
--- a/gems/smithy/lib/smithy/templates/client/types.erb
+++ b/gems/smithy/lib/smithy/templates/client/types.erb
@@ -8,6 +8,7 @@ module <%= namespace %>
 <% types.each do |type| -%>
 
     <%= type.documentation %>
+<% if type.type == 'structure' -%>
     <%= type.name %> = Struct.new(
 <% type.member_names.each do |member| -%>
       :<%= member %>,
@@ -16,6 +17,27 @@ module <%= namespace %>
     ) do
       include Smithy::Client::Structure
     end
+<% elsif type.type == 'union' -%>
+    class <%= type.name %> < Smithy::Client::Union
+<% type.members.each do |member| %>
+      class <%= member.name.camelize  %> < <%= type.name %>
+        def to_h
+          { <%= member.name.underscore %>: super(__getobj__) }
+        end
+      end
+<% end -%>
+
+      class Unknown < <%= type.name %>
+        def initialize(name:, value:)
+          super({name: name || 'Unknown', value: value})
+        end
+
+        def to_h
+          { unknown: super(__getobj__) }
+        end
+      end
+    end
+<% end -%>
 <% end -%>
 
   end
--- a/gems/smithy/lib/smithy/templates/client/types_rbs.erb
+++ b/gems/smithy/lib/smithy/templates/client/types_rbs.erb
@@ -2,6 +2,7 @@ module <%= namespace %>
   module Types
 <% types.each do |type| -%>
 
+<% if type.type == 'structure' -%>
     class <%= type.name %>
       include Smithy::Client::Structure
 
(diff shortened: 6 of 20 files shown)


gems/smithy-client/spec/smithy-client/union_spec.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    class TestUnion < Union
+      class StringValue < TestUnion
```
mullermp: I believe rspec/rubocop labels this anti pattern. I think Class.new(Union) inside of a let() block.

gems/smithy-client/sig/smithy-client/union.rbs
```diff
@@ -0,0 +1,7 @@
+module Smithy
+  module Client
+    class Union < ::SimpleDelegator
```
mullermp, line 3: Wondering if both union and structure in the future should be with shapes. (Do they apply to servers? I think so)

gems/smithy/lib/smithy/templates/client/types.erb
```diff
@@ -16,6 +17,27 @@ module <%= namespace %>
     ) do
       include Smithy::Client::Structure
     end
+<% elsif type.type == 'union' -%>
+    class <%= type.name %> < Smithy::Client::Union
+<% type.members.each do |member| %>
+      class <%= member.name.camelize  %> < <%= type.name %>
```
mullermp, line 23: Nice. Can we camelize types and whatever else while we are at it?

gems/smithy/lib/smithy/views/client/types.rb
```diff
@@ -39,8 +40,28 @@ def name
           end
 
           def member_names
-            @structure['members'].keys.map(&:underscore)
+            @shape['members'].keys.map(&:underscore)
           end
+
+          def members
+            @shape['members']
+              .map { |name, member| Member.new(name, member['target'], Model.shape(@model, member['target'])) }
```
mullermp: You can just resolve shape inside member.

gems/smithy/lib/smithy/templates/client/types.erb
```diff
@@ -8,6 +8,7 @@ module <%= namespace %>
 <% types.each do |type| -%>
```
mullermp, line 8: Not sure if this cleans up more, but instead of types, we can have both structures and unions separately and iterate both lists.

mullermp (approved):
Nice

mullermp:
We will probably want to add union shapes in shapes tests, as well as some documentation generation.

gems/smithy-client/spec/smithy-client/union_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+        def to_h
+          { string_value: super(__getobj__) }
+        end
+      end
+    end
+
+    describe Union do
```
jterapin: I'm not sure where we could test this but might be worth to test out an union member is of a structure shape .

gems/smithy/lib/smithy/templates/client/types.erb
```diff
@@ -16,6 +17,27 @@ module <%= namespace %>
     ) do
       include Smithy::Client::Structure
     end
+<% elsif type.type == 'union' -%>
```
jterapin, line 20: I think I forgot to cover this in the Doc but is there any changes that needs to make to `UnionShape` itself? Should it have type references to the subclasses from the POV of Union Shape Schema.

jterapin (commented):
Nice!

gems/smithy-client/spec/smithy-client/union_spec.rb
alextwoods: Yeah I'll update this - I just copied this from Hearth.

gems/smithy-client/spec/smithy-client/union_spec.rb
alextwoods: Will add testing of Structure

gems/smithy/lib/smithy/templates/client/types.erb
alextwoods: Good question - it does have a reference to the top level class which would let you get all of the member classes (but does require runtime inspection of the object space)... so maybe we should have a map of member to type?

gems/smithy/lib/smithy/templates/client/types.erb
alextwoods: I'll start a seperate PR for that - don't want to conflate naming with union implementation.

gems/smithy/lib/smithy/templates/client/types.erb
jterapin: Might be good to implement and useful for customers to access a subclass easily. I want to say the UnionShape in V3 had a way to access to  its member subclasses.

gems/smithy-client/sig/smithy-client/union.rbs
alextwoods: Summarizing offline discussion - long term these do probably belong in a smithy-core library that both smithy-client and smithy-server would depend on.  However, for now we will leave everything in smithy-client and consider a future refactoring.

gems/smithy/lib/smithy/templates/client/types.erb
alextwoods: https://github.com/smithy-lang/smithy-ruby/pull/261

gems/smithy/lib/smithy/views/client/types.rb
alextwoods: It requires passing the model in, but I guess thats consistent with what we're doing elsewhere so I'll change.

gems/smithy-client/spec/smithy-client/union_spec.rb
alextwoods: Note - ended up adding this as part of the structure_spec (since it has changes to the to_h method and I want to ensure that the Struct->Union->Struct case works correctly in to_h).

gems/smithy-client/lib/smithy-client/shapes.rb
```diff
@@ -164,7 +164,26 @@ def member(name)
...
+        attr_accessor :member_types
+
+        # @return [MemberShape]
+        def add_member(name, shape, type, traits: {})
+          @member_types[name] = type
+          super(name, shape, traits: traits)
+        end
+
+        # @return [Class, nil]
+        def member_type(name)
```
mullermp, line 183: Is members a good enough method name instead?

gems/smithy-client/spec/smithy-client/structure_spec.rb
```diff
@@ -10,12 +10,22 @@ module Client
...
           :value,
           :some_object,
+          :union_value,
           keyword_init: true
         ) do
           include Structure
         end
       end
 
+      let(:test_union_class) { Class.new(Union) }
```
mullermp: Nit: I would call these classes just union_class and union_structure_class? (And other tests in other files with similar name?)

gems/smithy-client/spec/smithy-client/union_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+  module Client
+    describe Union do
+      let(:test_union_class) { Class.new(Union) }
+      let(:string_value_class) do
+        Class.new(test_union_class) do
+          def to_h
+            { string_value: super(__getobj__) }
+          end
+
+          # anonymous class, need a class name to test to_s
```
mullermp, line 13: Self explanatory imo

gems/smithy-client/spec/smithy-client/union_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+          def self.name
+            'TestUnion::StringValue'
```
mullermp, line 15: Can drop Test prefix imo.

gems/smithy/lib/smithy/templates/client/types.erb
```diff
@@ -8,6 +8,7 @@ module <%= namespace %>
 <% types.each do |type| -%>
 
     <%= type.documentation %>
+<% if type.type == 'structure' -%>
     <%= type.name %> = Struct.new(
```
mullermp, line 12: Did we add camelize to this one?

gems/smithy/lib/smithy/views/client/request_response_example.rb
```diff
@@ -111,6 +111,17 @@ def structure(structure_shape, indent, visited)
           lines.join("\n")
         end
 
+        def union(union_shape, indent, visited)
+          lines = []
+          lines << '{'
+          union_shape['members']&.first&.tap do |member_name, member_shape|
```
mullermp: Instead of first, I would instead list all of the members, with a top level comment of "One of: " like we do with enums?

gems/smithy/lib/smithy/views/client/shapes.rb
```diff
@@ -69,41 +69,42 @@ def build_shape(id, shape)
...
+        def build_map_members(id, shape)
           %w[key value].map do |m_name|
             m_shape = shape[m_name]
-            build_member_shape(m_name, m_shape['target'], m_shape['traits'])
+            build_member_shape(id, m_name, m_shape['target'], m_shape['traits'])
           end
         end
 
-        def build_member_shape(name, id, traits)
+        def build_member_shape(parent_id, name, id, traits)
```
mullermp, line 105: This was a lot of plumbing. I think parent id should be id and the current id renamed to target?

gems/smithy/lib/smithy/views/client/shapes.rb
```diff
@@ -178,6 +179,7 @@ def initialize(options = {})
         # Member Shape represents members of Smithy shape
         class MemberShape
           def initialize(options = {})
+            @parent_id = options[:parent_id]
```
mullermp, line 182: Ditto on rename

gems/smithy/spec/fixtures/all_shapes/model.smithy
```diff
@@ -76,4 +76,7 @@ map ComplexMap {
 union Union {
     string: String,
```
mullermp, line 77: You can remove integer here, I just needed two values. I think we also want simple list and simple map (for documentation tests?)

gems/smithy/spec/interfaces/client/types_spec.rb
```diff
@@ -4,27 +4,46 @@
   %i[schema client].each do |plan_type|
     context "#{plan_type} generator" do
       before(:all) do
-        @tmpdir = SpecHelper.generate(['Weather'], plan_type)
+        @tmpdir = SpecHelper.generate(['AllShapes'], plan_type)
```
mullermp, line 7: This is fine in general. Weather was useful to get started. We have shapes and all_shapes. I eventually want one of them.. so no action needed

mullermp (commented):
Nice

gems/smithy-client/lib/smithy-client/shapes.rb
alextwoods: members already exists to return the MemberShape - I wanted this to be distinct.

We could return instead a tuple of [MemberShape, Class] but that seems clunky and is a different interface from the way StructureShape works.

gems/smithy/lib/smithy/templates/client/types.erb
alextwoods: the type? No - thats the raw smithy type.

If you mean the name on the next line - yes.

gems/smithy/lib/smithy/views/client/request_response_example.rb
alextwoods: Yeah that makes sense, will do.

gems/smithy/lib/smithy/views/client/shapes.rb
alextwoods: calling it `id` for the parent shapes id is a little confusing to me - theres technically three "ids" we're talking about for a member: 1) the parent shapes id 2) the member shape's id and 3) the target of the member shape's id.

I'd prefer to keep the parent_id explicit.

mullermp: Isn't member shapes id the same as name here? I thought members didn't have IDs, only the shapes with members and their targets. It's otherwise fine as is.

alextwoods: the member shape id includes the name eg: `smithy.example.foo#ExampleShapeName$memberName` see: https://smithy.io/2.0/spec/model.html#shape-id).  I don't think our codegen would ever really use member ids though, but the smithy concept is there.

I could rename the `id` parameter here as `target_id` as well?

mullermp: I think target is preferred naming either way, it's what we've been using.

gems/smithy-client/spec/smithy-client/union_spec.rb
jterapin: I don't think it hurts to keep it around (I could definitely use the reminder)

gems/smithy/lib/smithy/templates/client/types_rbs.erb
```diff
@@ -16,7 +17,18 @@ module <%= namespace %>
...
 <% end -%>
     end
+<% elsif type.type == 'union' -%>
+    class <%= type.name %> < Smithy::Client::Union
+<% type.members.each do |member| %>
+      class <%= member.name.camelize  %> < <%= type.name %>
+      end
 <% end -%>
 
+      class Unknown < <%= type.name %>
```
jterapin, line 27: I was looking at the generated shapes example on the PR description and noticed that the `Union.add_member(...)` is not covering the Unknown type. Would this be useful for a UnionShape to have access to when deserializing an unknown member? 

In V3 - we did add the unknown type as member subclasses ([example](https://github.com/aws/aws-sdk-ruby/blob/b9cc6c32da6e28e9563a3abbbb6ea3d5ea866de6/gems/aws-sdk-apptest/lib/aws-sdk-apptest/client_api.rb#L289)).

jterapin (approved):
This looks great!

gems/smithy/lib/smithy/templates/client/types_rbs.erb
alextwoods: Yeah - this is a good question! It is useful to have for sure - though not strictly nessecary.  We have the top level class and we know that there will always be an `Unknown` subclass so we can do `union_shape.type.const_get(:Unknown)`

I was working on adding it, but realized we'd then need to give it a Shape as a target.  We could give it Unit, but I don't think its quite right.  I'm actually leaning towards *not* including it in the list of members since Unknown isn't technically a member.

mullermp: Unknown is something we just parse into for forwards compatibility. I also don't even think we need member types at all (when would you need to know these?) If this isn't actually used at runtime, documentation that we already generated is sufficient for discovery.

alextwoods: I think we need the member types for deserialization unless we want active support to be a runtime dependency (because would need to go from member_name to MemberClass using the exact same logic as we do during code generatoion)

PR #261, opened 2025-01-30
Use camelize on class names

alextwoods:
*Description of changes:*
Applies `camelize` to all generated class/constant names - in the future we may need a view utility to handle shape IDs such as `__123Struct` and to handle name conflicts (ie, a reserved word list).  


mullermp (approved):

jterapin (approved):
I think we could use a TODO somewhere so we don't forget about having this. I will have to wrap up the pending list soon

PR #262, opened 2025-02-02
Options refactoring

mullermp:
This change does three things:
1) Adds a real quiet option to disable output instead of captured stdout. Makes use of say_status in thor classes.
2) Do not allow an options hash in plan and instead resolve accessors on initialize, making them immutable and also easier to use.
3) Adds extra options for name/identifier and squishes destination_root and source_only.


gems/smithy/spec/smithy/plan_spec.rb
```diff
@@ -21,11 +21,7 @@ module Smithy
       end
     end
 
-    describe '#options' do
-      it 'returns the options' do
-        expect(subject.options).to eq(options)
-      end
-    end
+    # TODO: unit tests for accessors in Plan
```
mullermp: Will add tests if the accessors and their defaults make sense.

gems/smithy/lib/smithy/plan.rb
```diff
@@ -9,9 +9,17 @@ class Plan
...
       @model = model
       @type = type
-      @options = options
       @service = find_service(model['shapes'])
 
+      @name = options.fetch(:name, default_name(service))
+      @identifier = @name.downcase
+      @module_name = options.fetch(:module_name, @name)
+      @gem_name = options.fetch(:gem_name, default_gem_name(@module_name, @type))
+      @gem_version = options.fetch(:gem_version)
```
jterapin, line 26: Since there's no default value set on `:gem_version`, this will raise `KeyError` - should there be a default in place to prevent this from happening? The initialize params does not indicate that this is a required in the options hash

jterapin (approved):
This looks great, I like that we pushed majority of the "naming" related logic to Plan to figure out from the get-go.

gems/smithy/lib/smithy/generators/base.rb
```diff
@@ -10,8 +10,9 @@ class Base
       # @param [Plan] plan The plan to generate.
       def initialize(plan)
         # Necessary for Thor::Base and Thor::Actions
-        self.options = { force: true }
-        self.destination_root = plan.options[:destination_root]
+        self.options = { force: true, quiet: plan.quiet }
+        self.destination_root = plan.destination_root
+        shell.base = self
```
alextwoods, line 15: Whats this doing?

gems/smithy/lib/smithy/plan.rb
```diff
@@ -6,35 +6,62 @@ class Plan
     # @param [Hash] model The API model as a JSON hash.
     # @param [Symbol] type The type of code to generate, either :client, :server, or :schema.
     # @param [Hash] options The options passed to the generator.
+    # rubocop:disable Metrics/AbcSize
     def initialize(model, type, options = {})
```
alextwoods, line 18: Should we document the options for initialize?

gems/smithy/lib/smithy/views/client/client.rb
```diff
@@ -12,16 +12,16 @@ def initialize(plan, code_generated_plugins)
           super()
         end
 
-        def namespace
-          @plan.gem_namespace
+        def module_name
```
alextwoods, line 15: Every single view we have needs to do this - maybe we should move this to our [View](https://github.com/smithy-lang/smithy-ruby/blob/decaf/gems/smithy/lib/smithy/views/view.rb) class that all of these inherit?

gems/smithy/spec/spec_helper.rb
```diff
@@ -85,21 +77,14 @@ def load_model(modules, options)
 
     def create_plan(modules, model, type, options)
       plan_options = {
-        gem_name: options.fetch(:gem_name, Smithy::Util::Namespace.gem_name_from_namespaces(modules)),
-        gem_version: options.fetch(:gem_version, '1.0.0'),
-        destination_root: options.fetch(:destination_root, Dir.mktmpdir)
+        module_name: modules.join('::'),
+        gem_version: options.fetch(:gem_version, '0.1.0'),
+        destination_root: options.fetch(:destination_root, Dir.mktmpdir),
+        quiet: ENV.fetch('SMITHY_RUBY_DEBUG', 'false') == 'false'
```
alextwoods: It feels a little weird for the ENV to not match the name.

gems/smithy/lib/smithy/welds/endpoints.rb
```diff
@@ -8,7 +8,7 @@ def pre_process(model)
         id, service = model['shapes'].select { |_k, s| s['type'] == 'service' }.first
         return if service['traits'] && service['traits']['smithy.rules#endpointRuleSet']
 
-        puts "Adding default endpoint rules to #{id}"
+        say_status :insert, "Adding default endpoint rules to #{id}", :yellow
```
alextwoods, line 11: Probably a broader discussion than this PR - but does thor give us the ability to support log levels? Rather than just an on/off we may want to be able to do like debug/info/error or something.

gems/smithy/spec/spec_helper.rb
```diff
@@ -69,14 +69,6 @@ def load_rbs_environment(sdk_dir, load_collection: true)
 
     private
 
-    def with_captured_stdout
```
alextwoods, line 72: Nice!

gems/smithy/lib/smithy/plan.rb
mullermp: I think we should explicitly require gem_version - this prevents overwrites of existing versions.

gems/smithy/lib/smithy/generators/base.rb
mullermp: Typically when you inherit from Thor, it does some setup, but this is the minimal setup for including thor actions and base. The thor shell is what controls output. The shell's base (the class) is nil, so quiet? method ends up returning false always. Now that base is not nil, quiet option is respected.

gems/smithy/lib/smithy/plan.rb
mullermp: Yeah, I think we should. They are also on the command line, but I see cases where we want them here too.

gems/smithy/spec/spec_helper.rb
mullermp: You mean SMITHY_RUBY_QUIET? I think that's less descriptive about what it does.. I'll think about it.

gems/smithy/lib/smithy/welds/endpoints.rb
mullermp: No, it doesn't. I investigated having a Logger and a configured log level. We could do that instead - but create_file will use say_status, so we would have to quiet everything always, and do our own logging. I'm not sure if we care to add logging output since generations are quick one-time invokes every so often. Would this be loaded in an always-running process you think?

gems/smithy/lib/smithy/views/client/client.rb
mullermp: Good question. I'm not sure. Originally I had that in View but pulled it out. I don't know how View will impact servers and whether these methods would exist. We could stuff them in View, but also, it might be good for each view to be explicit about what methods they need.

gems/smithy/lib/smithy/generators/base.rb
alextwoods: Ah I see!  I think that was the key piece I was missing when I tried to get quiet to work.

gems/smithy/spec/spec_helper.rb
alextwoods: We could also rename the cli parameter/plan option.  I personally think `SMITHY_RUBY_QUIET` is more descriptive than debug though, but I do think the env name, cli parameter and plan option shoudl all match.

gems/smithy/lib/smithy/welds/endpoints.rb
alextwoods: Other than for testing - no I dont really expect usage in an always running process.  I think most usage would come through invoking the CLI or through our own release process.  

I'm just thinking through what our release would look like - I think we generally don't want to log the create for every file, that ends up being huge for all of AWS services, but we likely want to be able to log warnings of some sort (like unknown traits? invalid formatting in docs?  I'm not quite sure).  

For now - I think just having the quiet option is reasonable.

mullermp: I'll look into options in the future. There is also a `say_error` that we could conditionally use based on options. I could put a logger on plan.

gems/smithy/spec/spec_helper.rb
```diff
@@ -80,7 +80,7 @@ def create_plan(modules, model, type, options)
         module_name: modules.join('::'),
         gem_version: options.fetch(:gem_version, '0.1.0'),
         destination_root: options.fetch(:destination_root, Dir.mktmpdir),
-        quiet: ENV.fetch('SMITHY_RUBY_DEBUG', 'false') == 'false'
+        quiet: ENV.fetch('SMITHY_RUBY_QUIET', 'true') == 'true'
```
alextwoods, line 83: Should this be defaulting to true?  I think we'd want to default it to false.

alextwoods (approved):
Nice - looks good.

gems/smithy/spec/spec_helper.rb
mullermp: In tests don't we want no debug output?

gems/smithy/lib/smithy/plan.rb
jterapin: Yeah, that makes sense to avoid overwrites - may want to document that is `gem_version` is required to build the gem

gems/smithy/spec/spec_helper.rb
alextwoods: Ahhh, right.  nvm.

PR #263, opened 2025-02-02
Rename examples models; improve shape specs

mullermp:
Diffed against new-options PR

Improves specs around shapes.


gems/smithy-client/lib/smithy-client/shapes.rb
```diff
@@ -75,6 +75,16 @@ def initialize(options = {})
         def add_member(name, shape, traits: {})
           @members[name] = MemberShape.new(shape, traits: traits)
         end
+
+        # @return [Boolean]
+        def member?(name)
```
alextwoods, line 80: Should these have rbs updates as well?

gems/smithy/spec/interfaces/client/shapes_spec.rb
```diff
@@ -11,188 +11,431 @@
...
+      expect(subject::Blob.traits).to eq(shape['traits'])
+    end
+  end
 
-      it 'has traits and does not contain omitted traits' do
-        expected_traits =
-          expected_shape
-          .values
-          .first['traits']
-          .reject { |k, _| k.include?('documentation') }
```
alextwoods, line 37: Are we still testing that some traits are omitted?  I didn't see this same test preserved, but its also a lot of new tests so may have missed it.

gems/smithy/spec/interfaces/client/shapes_spec.rb
```diff
@@ -11,188 +11,431 @@
...
-      expect(subject::Blob).to be_a(shapes_module::BlobShape)
+    it 'has traits' do
+      _, shape = fixture['shapes'].find { |_, s| s['type'] == 'blob' }
+      expect(subject::Boolean.traits).to eq(shape['traits'])
     end
+  end
 
-    it 'boolean is an instance of BooleanShape' do
-      expect(subject::Boolean).to be_a(shapes_module::BooleanShape)
+  context 'string' do
```
alextwoods, line 51: It feels like a lot of duplicated test code here - maybe a shared example or helper for these?

gems/smithy-client/lib/smithy-client/shapes.rb
mullermp: Yes. I'll add that.

gems/smithy/spec/interfaces/client/shapes_spec.rb
mullermp: I removed the test that traits are omitted. Since adding those traits are benign, it's just a matter of cutting down generated code. I didn't think it was necessary to test that.

gems/smithy/spec/interfaces/client/shapes_spec.rb
mullermp: I agree. I considered a loop but forgot about shared example. Good idea.

gems/smithy/spec/fixtures/shapes/model.smithy
```diff
@@ -2,96 +2,119 @@ $version: "2"
 
 namespace smithy.ruby.tests
 
-@paginated(inputToken: "nextToken", outputToken: "nextToken", pageSize: "pageSize")
+@trait
```
jterapin, line 5: TIL about `@trait`

gems/smithy-client/spec/smithy-client/shapes_spec.rb
```diff
@@ -212,11 +290,78 @@ module Shapes
...
+          end
+
+          it 'can set traits' do
+            subject = MemberShape.new(Shape.new, traits: { 'trait' => 'value' })
+            expect(subject.traits).to eq({ 'trait' => 'value' })
+          end
+        end
+      end
+
+      describe Prelude do
```
jterapin, line 337: Nice!

jterapin (approved):

gems/smithy/spec/fixtures/shapes/model.smithy
mullermp: Yeah. This allows us to test traits on all shapes.

gems/smithy/spec/interfaces/client/shapes_spec.rb
alextwoods: Including them won't break behavior, but I think its an important part of the code generation logic and so that behavior should get tested somewhere.

alextwoods (approved):

PR #264, opened 2025-02-03
Implement protocol plugin and rpcv2 protocol

jterapin:
## Description
This PR aims to provide the following features:
* Ability to swap protocol for another protocol.
* Resolve for default protocol based on what welds are applied
* Allow customers to pass in their preferred protocol configurations if they choose to

Differences between current state of V3 and smithy-ruby:
* There is a default protocol plugin that is codegenerated for every `client` gem generation
* A new configuration, `protocol` on the client
* Protocols are no longer their own individual plugins with handlers
* `ContentTypePlugin` in V3 is no longer needed - moved all the logic to `build_request` method.

## Customer Experience

Customer already has a default protocol defined on client creation:
```
client = Service::Client.new
client.config.protocol
=> Smithy::Client::Protocols::RPCV2.... 
```

Customer could overwrite the default protocol:
```
# custom protocol with preferred settings
custom_protocol = Smithy::Client::Protocols::CustomRPCV2.new(query_compatible: true)

# sets the custom protocol as active protocol
client = Service::Client.new(protocol: custom_protocol)
client.config.protocol
=> Smithy::Client::Protocols::CustomRPCV2.... 
```

Future TODOs:
* Updating RPCV2 protocol
  * Handle eventstreams 
  * Handle error handling
  * Remove Content-Length header once Content-Length Plugin/handler is implemented
 * Make any changes once Welds are updated to handle ordering
  * Add tests to validate weld ordering 


gems/smithy-client/lib/smithy-client/codecs/cbor.rb
```diff
@@ -30,7 +30,8 @@ def serialize(data, shape)
         # @param [Struct] type
         # @return [Object, Hash]
         def deserialize(bytes, shape, type = nil)
-          return {} if bytes.empty?
+          return {} # for development purposes since default response body is StringIO
+          # return {} if bytes.empty?
```
jterapin: I noticed that in the smithy-decaf implementation for response class - we did not end up implementing `body_contents` method ([source](https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-core/lib/seahorse/client/http/response.rb#L45-L56)) - I was wondering if we should migrate that functionality over to cover eventstreams and converting io -> string. Thoughts?

gems/smithy-client/lib/smithy-client/protocols/rpcv2.rb
```diff
@@ -0,0 +1,50 @@
...
+
+        def parse_response(context)
+          puts 'PARSING RESPONSE FROM RPCV2 PROTOCOL'
+          # TODO: Need to refactor to handle eventstreams
+          output_shape = context.operation.output
+          codec = Client::Codecs::CBOR.new(codec_setting(output_shape))
+          codec.deserialize(context.response.body, output_shape, output_shape.type)
+        end
+
+        # TODO: Implement ContentType Handler specific to RPCV2
```
jterapin: Downsides of the idea behind "codegenerated plugins" imo, less control over protocol-specific handlers.  I haven't quite figured this part out that works with codegenerated plugin for protocol.

Maybe its better to not have codegenerated plugin.... still thinking things through but open to ideas.

gems/smithy/lib/smithy/templates/client/client.erb
```diff
@@ -37,12 +37,15 @@ module <%= module_name %>
 
     def build_input(operation_name, params)
       handlers = @handlers.for(operation_name)
+      operation = config.schema.operation(operation_name)
+      config.protocol.service_id = '<%= service_id %>'
```
jterapin: We need to the protocol to be aware of the service-id namespace due to typed document feature (in the future, there may be cases when the returned document might fall under a different service namespace but we still need to deserialize it accordingly)

Ideally, I would have set the service id when the protocol initializes ONCE but at protocol swap, i don't think its a good CX when customer has to set the service-id themselves before passing in the instance of their preferred protocol.... Thoughts? 

Alternatively, I could check to see if the service_id is set on the protocol - if not, i will retrieve the information from config schema to be set... (I actually think I like this alternative better now that I write it out)

(Also - I figured this might be fine since I noticed that gem_name and gem_version is set whenever the request_context is created).

gems/smithy/lib/smithy/views/client/client.rb
```diff
@@ -16,6 +16,10 @@ def module_name
           @plan.module_name
         end
 
+        def service_id
+          @plan.service.keys.first
+        end
```
jterapin: Note to self: I think it is supposed to be the full service-id namespace. I need to check docs / code.

gems/smithy/lib/smithy/views/client/protocol_plugin.rb
```diff
@@ -0,0 +1,45 @@
...
+
+        def module_name
+          @plan.module_name
+        end
+
+        def protocol_name
+          raise ArgumentError, 'Invalid protocol' if (protocol = resolve_protocol).nil?
+
+          protocol
+        end
```
jterapin: TBD on exact Smithy plan of determining what protocol is "default" but Dowling is expecting that there is always a default protocol set.  (not sure if this is a REQUIREMENT since you can still build out a model w/o protocol)

Right now, I just look for the very first protocol that starts with `smithy.protocols` namespace

This gets me thinking about - should client generation happen even if there is no protocol set? Or should there be a raise?

jterapin (commented):
Just an idea of codegenerated protocol plugin so may be scrapped.

I liked the idea in the beginning but there's some limitations that I need to think through and makes me lean toward V3 implementation for protocols. Open to thoughts!

gems/smithy-client/lib/smithy-client/handler_context.rb
```diff
@@ -20,6 +21,7 @@ def initialize(options = {})
         @config = options[:config]
         @request = options[:request] || HTTP::Request.new
         @response = options[:response] || HTTP::Response.new
+        @protocol = options[:protocol]
```
alextwoods: This should already be on config right?  I'm not sure we need to add effectively an acessor for it to context.

gems/smithy-client/lib/smithy-client/protocols.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    module Protocols
+      # Base protocol class that all protocols derived from
+      class ProtocolBase
+        def initialize(id)
+          @id = id
```
alextwoods: What is id?

gems/smithy-client/lib/smithy-client/protocols/rpcv2.rb
```diff
@@ -0,0 +1,50 @@
...
+        private
+
+        def apply_headers(request)
+          # TODO: Add X-Amz-Query-Mode header if aws.protocols#awsQueryCompatible is true
+          request.headers['Smithy-Protocol'] = 'rpc-v2-cbor'
+          # TODO: Remove after implementing ContentLength Plugin/Handler
+          request.headers['Content-Length'] = request.body.size
+        end
+
+        def codec_setting(_shape)
```
alextwoods: Are the codec settings based on the operation's shape or based on the service?

Thinking about the case of awsQueryCompatible, I think that is applied at the service level?

gems/smithy/lib/smithy/templates/client/client.erb
alextwoods: I think I actually lean towards getting the serviceID out of the schema rather than having it as an attribute on protocol.  I think smithy-java has serviceId on its protocol, but I think for our cases, it maybe makes more sense to get it from the schema.

gems/smithy/lib/smithy/templates/client/protocol_plugin.erb
```diff
@@ -0,0 +1,43 @@
...
+  module Plugins
+     # Protocol plugin
+     class Protocol < Smithy::Client::Plugin
+       option(
+         :protocol,
+         doc_default: 'Smithy::Client::Protocols::<%= protocol_name %>',
+         doc_type: 'Smithy::Client::Protocols::ProtocolBase',
+         docstring: 'TODO'
+         ) do |_cfg|
+           Smithy::Client::Protocols::<%= protocol_name %>.new
```
alextwoods: I understand why we're using a codegen plugin to determine the default protocol here and this is pretty reasonable.

I was trying to think through alternatives to this - one option would be a `ProtocolResolver` that takes the schema (and its associated, ordered protocol traits) and then loads all of the descendants of ProtocolBase and finds the first matching one.  Sort of like engines, this would let users dynamically pick protocols based on 1) the preference of the service and 2) what they have available at runtime.

The downside of that is that the default protocol used would change based on what the user has available... and means we couldn't give it a good doc default either.

gems/smithy/lib/smithy/templates/client/protocol_plugin.erb
```diff
@@ -0,0 +1,43 @@
...
+         end
+
+       # @api private
+       class Build < Smithy::Client::Handler
```
alextwoods, line 41: I know these are super simple, but I think I'd lean towards having the handlers be in smithy-client and minimizing the amount of codegen.  (and we should maybe do that with endpoint plugin as well?)

gems/smithy/lib/smithy/views/client/client.rb
alextwoods: Yeah - I think it is.

gems/smithy/lib/smithy/views/client/protocol_plugin.rb
alextwoods: > Right now, I just look for the very first protocol that starts with smithy.protocols namespace

I think this is actually a problem we have in our model.  In the Java smithy codegen we could say, give me all protocol traits.  But in our model we don't have anything to tell us if a given trait is a protocol or not. (unless its a custom trait that we define ourselves).

We also need a way to register a Ruby protocol class to a protocol trait.  We could ptoentially solve both together by having welds "register" protocols - we would then ask all our welds for their protocols.  That gives us a list of all supported protocol traits and we can then match these up and find the first on the service and make that the default.

gems/smithy-client/lib/smithy-client/codecs/cbor.rb
mullermp: I'm not sure if we want body_contents. What are you trying to do here?

gems/smithy-client/lib/smithy-client/protocols.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    module Protocols
+      # Base protocol class that all protocols derived from
+      class ProtocolBase
```
mullermp: I personally don't think we should have a base class - we should just follow duck typing. We also don't get any inherited functionality.

gems/smithy-client/lib/smithy-client/protocols/rpcv2.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    module Protocols
+      # RPCV2 protocol
+      class RPCV2 < ProtocolBase
```
mullermp: RPCv2 (this becomes rpc_v2 in underscoring).

gems/smithy-client/lib/smithy-client/protocols/rpcv2.rb
mullermp: Yeah, I think at minimum we need to code generate the protocol default / list.

gems/smithy/lib/smithy/templates/client/client.erb
mullermp: I agree it should be a property of schema - or an identifier on the client. See v3's identifier.

gems/smithy/lib/smithy/templates/client/protocol_plugin.erb
mullermp: What would we minimize in the endpoint plugin? But I agree, I don't think this needs to be code generated, and we can resolve the default on the service shape or in a config option.

gems/smithy/lib/smithy/views/client/client.rb
mullermp: I think this exists as plan.name or at worst case we can add identifier.

gems/smithy/lib/smithy/views/client/protocol_plugin.rb
mullermp: Agree with registration - take the ID and map it to a class.

gems/smithy/lib/smithy/templates/client/client.erb
jterapin: Yes, smithy-java has a serviceId defined on the protocol. 

During deserialization process, we will always have access to the schema via Context so yeah - we could find the service id there instead. Let me reroute to that approach and see how it goes.

gems/smithy-client/lib/smithy-client/handler_context.rb
jterapin: Yeah, definitely could use `config.protocol` instead. I'll remove.

gems/smithy-client/lib/smithy-client/protocols.rb
jterapin: The `id` here is a the protocol shape id - might not be necessary to have.

gems/smithy-client/lib/smithy-client/protocols/rpcv2.rb
jterapin: Ahh yeah, `awsQueryCompatible` trait is applied at service-level. Hmm, thinking through the setting piece - there's technically....2 settings that affect SERDE process:
* Service-shape level
* Operation-shape level
Ideally, when the protocol instantiated - the protocol remembering the state of the service-level setting will be good, I think. If not - we could always simplify this by checking the context to retrieve both of the settings.

gems/smithy-client/lib/smithy-client/protocols.rb
alextwoods: I think instead of it being a member it should be a property/constant of the class - I think that will make mapping protocol traits to protocol implementations easier.

gems/smithy-client/lib/smithy-client/codecs/cbor.rb
jterapin: Nothing exactly. Just an observation I noticed while investigating how RPCV2 protocol was implemented in V3. It was interesting why in some cases we use `body` vs `body_contents` when deserializing. We could take a look deeper when we work on eventstreams.

gems/smithy-client/lib/smithy-client/protocols.rb
jterapin: My initial thinking is that we want to ensure that the methods such as `build_request` and `parse_response` is always implemented when a customer decides to implement their own protocol.

That being said - I could get rid of this "base" and just leave good documentation on what needs to be done if the customer wants to bring their own (like follow cbor protocol as an example and list out the methods that we need to support).

gems/smithy-client/lib/smithy-client/protocols/rpcv2.rb
jterapin: Thanks! I will update this and other file-related changes.

gems/smithy/lib/smithy/templates/client/protocol_plugin.erb
jterapin: I don't exactly recall how flexible protocol swapping should happen. It didn't occur to me until now that only protocol traits defined on the service shape could be allowed as protocol alternatives. I was under the impression that customers could bring ANY protocol of their choice - outside of the service shape model...

Let's say that service shape traits only lists one protocol trait. Generated client with using that protocol as default. Customer wrote their own protocol (follows to our protocol implementation guidance) and decides to pass it in the client config  - is this allowed?

gems/smithy/lib/smithy/templates/client/protocol_plugin.erb
jterapin: Yeah - after reading through the discussion points - agreed, these handlers are not necessary to be codegenerated. I will rework here.

gems/smithy/lib/smithy/views/client/client.rb
jterapin: I think I won't need this - if we are relying on the schema to tell us the service id shape (should we need it).

gems/smithy-client/lib/smithy-client/protocols/rpcv2.rb
mullermp: I think it's easier to just get the traits from the service shape in this case. I don't think it's related to protocol implementation other than checking for existance (because it's a trait applied to service)

gems/smithy/lib/smithy/templates/client/protocol_plugin.erb
mullermp: I think so. They could modify an existing protocol to have separate behavior. I don't think it's useful other than cases where we make a backwards incompatible change or something.

gems/smithy-client/lib/smithy-client/protocols/rpc_v2.rb
```diff
@@ -0,0 +1,57 @@
...
+      # RPCV2 protocol
+      # TODO:
+      #  * Refactor methods to handle eventstreams
+      #  * Decide whether we should recognize query traits here
+      #  * Add Documentation
+      class RPCv2
+        SHAPE_ID = 'smithy.protocols#rpcv2Cbor'
+
+        def initialize(options = {})
+          @query_compatible = options[:query_compatible]
```
jterapin, line 17: This "query-compatible" is very aws-specific but I want to demo how we could have "settings" on protocol.

I imagined the settings idea will be more relevant for other protocols like JSON (customer decides to whether to honor `jsonName` trait).

That being said - the "home" for RPCV2 protocol is in `smithy-client` gem and we need to figure out a way to recognize `query_compatible` trait when we use this protocol for AWS services. 

We could just recognize the query trait here but that goes against the grain of separating between generic and aws. Alternatively, maybe `aws-sdk-core` could have a RPCV2 protocol class that inherits this and layers "query-compatible" code a top of it? Open to any thoughts.

gems/smithy-client/lib/smithy-client/protocols/rpc_v2.rb
```diff
@@ -0,0 +1,57 @@
...
+          context.request.headers['Content-Type'] = 'application/cbor'
+          # TODO: Implement Content-Length Plugin/Handler
+          context.request.headers['Content-Length'] = context.request.body.size
+        end
+
+        def setting(context)
+          {}.tap do |h|
+            h[:query_compatible] = true if query_compatible?(context)
+          end
+        end
```
jterapin, line 73: I probably don't need to build out like this for a single setting but want to demo - once again.

gems/smithy-client/lib/smithy-client/protocols/rpc_v2.rb
```diff
@@ -0,0 +1,57 @@
...
+
+        def query_compatible?(context)
+          @query_compatible ||
+            context.config.schema.service.traits.one? { |k, _v| k == 'aws.protocols#awsQuery' }
+        end
```
jterapin: Notes about protocol setting priority order:
* Customer passes an instance of protocol with their preferred configuration to the client.
* For this specific case, if `@query_compatible` is set to `nil` - then check the service shape and its traits

gems/smithy/lib/smithy/templates/client/protocol_plugin.erb
jterapin: Since these handlers are very generic - I was thinking about moving them into its own plugin class within `smithy-client` level but... what do I call it? `BuildParsePlugin`? Naming is hard but ideally I want a single plugin that adds these generic handlers. But I was also thinking of doing separate plugins like `Build` or `Parse` - kind of like `Sign` plugin that we already have. 

 Alternatively - I could leave it here for starters.

gems/smithy/lib/smithy/weld.rb
```diff
@@ -62,5 +62,13 @@ def endpoint_function_bindings
     def plugins
       {}
     end
+
+    # Called to construct the protocol plugin. Any protocols defined here will be
+    # merged with other protocols. The key is the protocol trait ID and the value
+    # is the fully qualified class name of the protocol.
+    # @return [Hash<String, String>] protocols for use in protocol plugin.
```
jterapin: Self-Reminder to fix this value documentation here.

model/weather.smithy
```diff
@@ -2,7 +2,10 @@ $version: "2"
 
 namespace example.weather
 
+use smithy.protocols#rpcv2Cbor
+
 /// Provides weather forecasts.
+@rpcv2Cbor
```
jterapin: I added this trait for development. I could keep it in or remove it to its own fixture model when I write the tests.

projections/weather/lib/weather/plugins/protocol.rb
```diff
@@ -0,0 +1,59 @@
...
+        docstring: <<~DOCS) do |_cfg|
+          Allows you to overwrite default protocol. The given protocol
+          must provide the following functionalities:
+          - `build_request`
+          - `parse_response`
+          - `parse_error`
+          See existing protocols within Smithy::Client::Protocols for examples.
+        DOCS
+        Smithy::Client::Protocols::RPCv2.new
+      end
```
jterapin: Example projection when no protocol is found on service shape:
```
module Weather
  module Plugins
    # Protocol plugin - allows user to configure protocol on client.
    # @api private
    class Protocol < Smithy::Client::Plugin
      option(
        :protocol,
        default: nil,
        docstring: <<~DOCS)
          This configuration is required to build requests and parse responses.
          In Smithy, a protocol is a named set of rules that defines the syntax
          and semantics of how a client and server communicate. The given protocol
          must provide the following functionalities:
          - `build_request`
          - `parse_response`
          - `parse_error`
          See existing protocols within Smithy::Client::Protocols for examples.
        DOCS
...
```

gems/smithy-client/lib/smithy-client/protocols.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    # TODO: Documentation
+    # All the generic protocols
```
jterapin: Prob a good spot for adding information creating your own protocol.

gems/smithy-client/lib/smithy-client.rb
```diff
@@ -43,6 +43,10 @@
 require_relative 'smithy-client/cbor'
 require_relative 'smithy-client/codecs/cbor'
 
+# protocols
+require_relative 'smithy-client/protocols'
+require_relative 'smithy-client/protocols/rpc_v2'
```
alextwoods: maybe the protocols.rb should have this require?

gems/smithy-client/lib/smithy-client/protocols/rpc_v2.rb
alextwoods: Yeah, this is a tricky line between generic and AWS that rpcv2 blurs.  We never quite dealt with it in the java smithy-ruby (we just didn't implement the query compatible part).

I'd honestly lean towards leaving the setting and check against the trait in here.  While it does check against an AWS defined trait, its using only the string shape ID and doesn't bring in any additional dependencies (at codegen or runtime).

gems/smithy-client/lib/smithy-client/protocols/rpc_v2.rb
alextwoods: +1 - that seems like correct behavior.

gems/smithy/lib/smithy/templates/client/protocol_plugin.erb
alextwoods: As an alternative - you could just define the handler classes in smithy-client, but have the generated plugin add them.

gems/smithy/lib/smithy/views/client/protocol_plugin.rb
```diff
@@ -0,0 +1,44 @@
...
+
+        private
+
+        def service_traits
+          @service_traits ||= @plan.service.values.first['traits']
+        end
+
+        def weld_protocols
+          @weld_protocols ||= @plan.welds.map(&:protocols).reduce({}, :merge)
```
alextwoods, line 34: The order of these is important and we need to ensure that when we add AWS protocols we can control the order everything ends up in - maybe add a TODO here RE: weld ordering.

gems/smithy/lib/smithy/views/client/protocol_plugin.rb
```diff
@@ -0,0 +1,44 @@
...
+        end
+
+        def default_protocol
+          return if weld_protocols.empty? && service_traits.nil?
```
alextwoods: Should this be an || instead of &&?

gems/smithy/lib/smithy/views/client/protocol_plugin.rb
```diff
@@ -0,0 +1,44 @@
...
+        def protocol
+          return if default_protocol.nil?
+
+          weld_protocols[default_protocol]
+        end
+
+        private
+
+        def service_traits
+          @service_traits ||= @plan.service.values.first['traits']
```
alextwoods: could probably do a .fetch('traits', {}) to then avoid having to do a nil check later

model/weather.smithy
alextwoods: I'd vote for creating a seperate fixture model for this though I do like seeing the generated diff here (which is why I also liked the more complete white_label model).

projections/weather/lib/weather/plugins/protocol.rb
alextwoods: Nice.

gems/smithy/lib/smithy/templates/client/protocol_plugin.erb
```diff
@@ -0,0 +1,73 @@
...
+       class Error < Smithy::Client::Handler
+         def call(context)
+           @handler.call(context).on(300..599) do |response|
+             context.config.protocol.parse_error(context, response)
+           end
+         end
+       end
+
+       def add_handlers(handlers, config)
+         handlers.add(Build)
```
alextwoods, line 67: We should be adding these to specific steps right since the ordering is pretty critical?

alextwoods (commented):
Generally looking good - I like the weld based mapping of service traits to ruby protocol implementations.

gems/smithy/lib/smithy/weld.rb
alextwoods: A general thought on these - with the `plugins` weld method, it requires us to provide a `require_path` which we then add to the generated client.

Thinking about where protocols come from and how we (for AWS and generic customers) would add them.  This current approach assumes that the class will be available/defined already at runtime.  I think this is reasonable for now, but we may need an approach like plugins where we add a `require`.  

And beyond that we may also want the ability for both plugins and protocols defined by welds to *also* define additional gem dependencies that are added to the gemspec.

gems/smithy/lib/smithy/templates/client/protocol_plugin.erb
mullermp: They can exist on build where they are.

gems/smithy/lib/smithy/templates/client/protocol_plugin.erb
```diff
@@ -0,0 +1,73 @@
...
+         handlers.add(Parse)
+         # TODO: Requires error handling to be implemented
+         # handlers.add(Error, step: :sign)
```
mullermp, line 70: Why is error on sign? It would also be where parse is I think. I believe insertion order is maintained.

gems/smithy/lib/smithy/views/client/protocol_plugin.rb
mullermp: Why is order important?

gems/smithy/lib/smithy/welds/protocols.rb
```diff
@@ -0,0 +1,27 @@
...
+    #  * Smithy RPCv2 CBOR
+    #  * AWS JSON 1.0
+    #  * AWS JSON 1.1
+    #  * REST JSON
+    #  * REST XML
+    #  * AWS/Query
+    #  * EC2/Query
+    class Protocols < Weld
+      def protocols
+        { 'smithy.protocols#rpcv2Cbor' => Smithy::Client::Protocols::RPCv2 }
```
mullermp, line 24: I think we can add a priority system later that is like our handler registration.

projections/weather/lib/weather/plugins/protocol.rb
```diff
@@ -0,0 +1,59 @@
...
+        doc_type: 'Smithy::Client::Protocols::RPCv2',
+        docstring: <<~DOCS) do |_cfg|
+          Allows you to overwrite default protocol. The given protocol
+          must provide the following functionalities:
+          - `build_request`
+          - `parse_response`
+          - `parse_error`
+          See existing protocols within Smithy::Client::Protocols for examples.
+        DOCS
+        Smithy::Client::Protocols::RPCv2.new
```
mullermp: We should have a convenience mapping of :rpc_v2 or other symbols to default classes and document it takes either or.

projections/weather/lib/weather/client.rb
```diff
@@ -86,6 +88,13 @@ class Client < Smithy::Client::Base
     # @option options [Logger] :logger
     #  The Logger instance to send log messages to. If this option is not set,
     #  logging is disabled.
+    # @option options [Smithy::Client::Protocols::RPCv2] :protocol (Smithy::Client::Protocols::RPCv2)
+    #  Allows you to overwrite default protocol. The given protocol
+    #  must provide the following functionalities:
+    #  - `build_request`
```
mullermp: Does v3 use "build" "parse" and "error"? You can shorten if you want.

gems/smithy/lib/smithy/templates/client/protocol_plugin.erb
mullermp: I think one plugin with multiple handlers is fine. I would just call it the protocol plugin. I think the only thing that needs to be code generated is the option and its defaults.

mullermp (commented):
Can discuss offline later.

projections/weather/lib/weather/plugins/protocol.rb
alextwoods: This is an interesting idea.  I think it is convenient for Ruby users and follows patterns we have in v3 - I think the question is how we actually would implement that.  

We obviously could use an after_initialize hook to convert the symbol into the class, but how and where do we maintain this mapping (while remaining generic)? I suppose this could be part of the code generation - we would find all of the available protocols from welds that match the service and generate a mapping of symbol to class and then generate an after_initialize? (as a bonus the after initialize could initialize a class if provided and also validate that its a valid protocol?)

gems/smithy/lib/smithy/templates/client/protocol_plugin.erb
jterapin: I mimicked what V3 had in terms of handler ordering: https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-core/lib/aws-sdk-core/plugins/protocols/json_rpc.rb#L26

After this PR is completed, I could take a detailed look into how we could go about handling errors since we have to a chance to change up the way we handle errors in V3.

gems/smithy/lib/smithy/views/client/protocol_plugin.rb
jterapin: Nice, that makes things easier.

gems/smithy/lib/smithy/welds/protocols.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+require 'smithy-client/protocols/rpc_v2'
+
+module Smithy
+  module Welds
+    # Provides map of protocol trait id and its Ruby class name.
+    # TODO: Update Welds to have a functionality to control ordering since there
```
jterapin, line 8: Reason why weld ordering is important.

gems/smithy/lib/smithy/welds/protocols.rb
jterapin: Yeah I'll make note this possible solution.

gems/smithy/lib/smithy/views/client/protocol_plugin.rb
jterapin: Yeah, I think I meant to write the `||` operator here. However, no need to the latter check since `service_traits` will return an empty hash if there's no traits.

projections/weather/lib/weather/client.rb
jterapin: Sure, let me shorten them (V3 used a mix of `build_request`, `parse_response` and `error`)

projections/weather/lib/weather/plugins/protocol.rb
jterapin: For now - I will add a TODO for revisit.

gems/smithy/lib/smithy/views/client/protocol_plugin.rb
jterapin: See: https://github.com/smithy-lang/smithy-ruby/pull/264#discussion_r1949525714

gems/smithy/spec/fixtures/multi_protocol/model.smithy
```diff
@@ -0,0 +1,14 @@
...
+
+use smithy.protocols#rpcv2Cbor
+
+// Define a fake protocol trait for use.
+@trait
+@protocolDefinition
+structure testProtocol {}
+
+@testProtocol
+@rpcv2Cbor
```
mullermp: May be easier if you just define two test protocols here - then you don't have to worry about cbor existing for this test, which seems to just be a resolution test?

projections/weather/lib/weather/client.rb
```diff
@@ -86,6 +88,13 @@ class Client < Smithy::Client::Base
     # @option options [Logger] :logger
     #  The Logger instance to send log messages to. If this option is not set,
     #  logging is disabled.
+    # @option options [Smithy::Client::Protocols::RPCv2] :protocol (Smithy::Client::Protocols::RPCv2)
```
mullermp: Doc type looks wrong. You should use a response to for yardoc #build(), etc.

gems/smithy-client/lib/smithy-client/handlers.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    # A collection of generic handlers that are used by Smithy Clients.
+    # @api private
+    module Handlers
```
mullermp: I'm not a fan of the half and half. Mainly because plugin code is not tied to the handler in the same package, so drift can occur if we reorder or change code, making extra cases to consider. I still think the entirety of the protocol plugin can be generic. The alternative is fully code generated which you had before. The only part that needs to be code generated is configuration default right? I am also thinking about the endpoint case on how we can genericize that one. One option I don't think we considered would be to set default config values using the before_initialize hook, perhaps in a code generated default config plugin. We can chat offline on it.

gems/smithy-client/lib/smithy-client/protocols/rpc_v2.rb
```diff
@@ -0,0 +1,59 @@
...
+
+        def initialize(options = {})
+          @query_compatible = options[:query_compatible]
+        end
+
+        # @api private
+        def build(context)
+          codec = Client::Codecs::CBOR.new(setting(context))
+          context.request.body = codec.serialize(context.params, context.operation.input)
+          context.request.http_method = 'POST'
```
alextwoods, line 24: This also needs to set a path of `/service/{serviceName}/operation/{operationName}` on the endpoint, see: https://smithy.io/2.0/additional-specs/protocols/smithy-rpc-v2.html#requests

gems/smithy-client/lib/smithy-client/handlers.rb
jterapin: Agreed, the separation between codegeneration plugin that using these handlers and the handlers definition in a separate file does not feel "right" to me. Your idea on having a codegenerated default config plugin is a good idea. 

In the subject of codegenerated endpoint plugin, I was think making the Endpoint Handler class generic could be tricky since it uses the codegenerated EndpointsParameters defined within the gem: https://github.com/smithy-lang/smithy-ruby/blob/decaf/projections/weather/lib/weather/plugins/endpoint.rb#L29

(Could be different if we are able to make that portion generic - however)

gems/smithy/spec/fixtures/multi_protocol/model.smithy
jterapin: Yeah true, I could use another dummy protocol in place but I do like being able to check that protocols welds are working correctly here.

mullermp: Isn't that already checked when you run protocol tests and the plugin added?

jterapin: Ahhhh yeah, once protocol tests are done - that's true. I could modify this now to using fake protocol (and remove smithy-build.json) with anticipation of cbor protocol tests being done soon (but keep around to test weld ordering later)

gems/smithy-client/lib/smithy-client/protocols/rpc_v2.rb
jterapin: Ahhhhh good catch, I was so focused on the build/parse that I didn't migrate those portions over. Let me double check!

gems/smithy-client/lib/smithy-client/handlers.rb
alextwoods: The issue with differing versions of codegen vs client runtime is a good point on this.

I'd suggest the entire plugin be code generated in that case.

Runtime resolution of the default plugin doesn't seem like the right behavior to me.

> One option I don't think we considered would be to set default config values using the before_initialize hook, perhaps in a code generated default config plugin. We can chat offline on it.

This is interesting - but how is it different from setting a default block in an `option` (except for the fact we couldn't generate default doc option, which I think is important for customers).

gems/smithy/spec/fixtures/multi_protocol/model.smithy
alextwoods: I dont think the protocol tests will cover default resolution?  They only (currently) ever cover single protocol cases.  And even with multiple protocols defined on a service, they would never test the default resolution - we would instead need to configure the correct protocol for each test on the client.

gems/smithy-client/lib/smithy-client/handlers.rb
jterapin: Let's have a discussion offline either later today or tomorrow AM.

gems/smithy/spec/fixtures/multi_protocol/model.smithy
mullermp: I'm not suggesting that. This test can simply test resolution with two dummy protocols. Juli wants to test rpc v2 with welds. That is implicitly done through protocol test because they would not succeed otherwise.

gems/smithy-model/lib/smithy-model/shapes.rb
```diff
@@ -198,6 +211,7 @@ class StructureShape < Shape
         def initialize(options = {})
           super
           @members = {}
+          @members_by_name = {}
```
jterapin, line 220: Basically - any shapes that involves members - we want to retain their member names to reference for (de)serialization process. (AKA the location_name in Shape ref)

gems/smithy-model/lib/smithy-model/shapes.rb
```diff
@@ -207,19 +221,22 @@ def initialize(options = {})
...
+        def add_member(name, member_name, shape, traits: {})
+          @members_by_name[member_name] = name
+          @members[name] = MemberShape.new(member_name, shape, traits: traits)
         end
 
+        # @param [Symbol, String] name
         # @return [Boolean]
         def member?(name)
-          @members.key?(name)
+          @members.key?(name) || @members_by_name.key?(name)
```
jterapin, line 241: I want flexibility to check members existence whether by the symbolized member name (`:city_id`) vs the member name defined on the shape (`cityName`). I also did this for `member(name)` method

gems/smithy/spec/fixtures/multi_protocol/model.smithy
jterapin: I was thinking about putting in a dummy protocol to replace the first one but then I have to create a dummy weld as well. For now - I'm going to leave it as it is until we implement/consider weld ordering.

Some test cases we want to cover are:
* Case: There are two protocols set on the service model but the merged protocol welds only has one protocol
  * Expected: Generated protocol plugin should be using what protocol weld is bringing
  
* Case: There are two protocol sets on the service model and the weld protocols supports them as well
  * Expected: Generated protocol plugin should determine the default protocol by the protocol order priority

* Case: No protocol weld, but service model has protocols defined
  * Expected: default is nil

gems/smithy-client/lib/smithy-client/codecs/cbor.rb
```diff
@@ -52,20 +56,35 @@ def format_data(value, shape)
...
+            data[key] =
+              if value.nil? && sparse?(shape)
+                nil
+              else
+                format_data(value, shape.value.shape)
+              end
           end
         end
 
         def format_structure(values, shape)
```
jterapin, line 83: I'm going to implement the (de)serialization of union shapes in a separate PR since I have some questions before implementing. Will be a small follow-up PR most likely.

projections/weather/lib/weather/plugins/protocol.rb
```diff
@@ -0,0 +1,58 @@
...
+          resp = @handler.call(context)
+          resp.data = context.config.protocol.parse(context)
+          resp
+        end
+      end
+
+      # @api private
+      class Error < Smithy::Client::Handler
+        def call(context)
+          @handler.call(context).on(300..599) do |response|
```
mullermp, line 44: Unfortunately s3 has errors at 200 level, so this may need to be 0..599 or some way to control/customize this.

projections/weather/lib/weather/plugins/protocol.rb
```diff
@@ -0,0 +1,58 @@
...
+          context.config.protocol.build(context)
+          @handler.call(context)
+        end
+      end
+
+      # @api private
+      class Parse < Smithy::Client::Handler
+        def call(context)
+          resp = @handler.call(context)
+          resp.data = context.config.protocol.parse(context)
```
mullermp: This is technically called "output" now - I keep using resp too.

projections/weather/lib/weather/client.rb
```diff
@@ -93,6 +95,12 @@ class Client < Smithy::Client::Base
     # @option options [Logger] :logger
     #  The Logger instance to send log messages to. If this option is not set,
     #  logging is disabled.
+    # @option options [#build, #parse, #error] :protocol
+    #  This configuration is required to build requests and parse responses.
+    #  In Smithy, a protocol is a named set of rules that defines the syntax
+    #  and semantics of how a client and server communicate. The given protocol
+    #  must provide the following functionalities: `build`, `parse` and `error`.
```
mullermp, line 102: Probably want to reword this to "should respond to build(context), parse(context) etc"

gems/smithy/lib/smithy/welds/protocols.rb
mullermp: I think we can add another method (later) for protocol ordering, and in aws-sdk-core, we have a plugin that defines the order (it would also have a reference to any in smithy-client).

gems/smithy-model/lib/smithy-model/shapes.rb
mullermp: Yea. I think we should choose a different name though..

mullermp (approved):

gems/smithy-model/lib/smithy-model/shapes.rb
mullermp: I'm not sure if we want to do that. We may get something returned unexpectedly?

projections/weather/lib/weather/plugins/protocol.rb
jterapin: Yeah that makes sense - we should do some brainstorm work when we start looking into error handling.

projections/weather/lib/weather/client.rb
jterapin: I'll make the changes in a separate PR.

PR #265, opened 2025-02-03
Refactor generated rbs spy testing

alextwoods:
*Description of changes:*
1. Adds an option for opting test suites out of rbs testing. 
2. Refactor rbs related code out of spec_helper.


Rakefile
```diff
@@ -6,10 +6,11 @@ require 'rspec/core/rake_task'
 RuboCop::RakeTask.new
 
 namespace :smithy do
-  RSpec::Core::RakeTask.new('spec:unit') do |t|
+  RSpec::Core::RakeTask.new('spec:unit') do |t, _args|
```
mullermp: Args isn't necessary anymore is it?

gems/smithy/spec/spec_helper.rb
```diff
@@ -18,6 +19,7 @@ class << self
     #  `:schema`, `:client`, or `:server`.
     # @param [Hash] options Additional options to pass to the generator.
     # @option options [String] :fixture The name of the fixture to load.
+    # @option options [Boolean] :rbs_test (true) Enable RBS spy testing when the SMITHY_RUBY_RBS_TEST is set.
```
mullermp: I don't think this option is necessary any more eithe r- we can just call setup if ENV

mullermp (approved):
Nice. Did you want to merge this now or do the client helper refactor too?

Rakefile
alextwoods: Right, removed.

PR #266, opened 2025-02-04
Load plugins from welds

mullermp:
This change loads default plugins from welds. It also guarantees order of welds and applies plugin options alphabetically.

### DIFF
--- a/gems/smithy/lib/smithy/views/client/client.rb
+++ b/gems/smithy/lib/smithy/views/client/client.rb
@@ -35,10 +35,13 @@ module Smithy
         end
 
         def docstrings
-          docstrings = []
-          @plugins.options_docstrings.each_value do |value|
-            docstrings.concat(value)
+          options = @plugins.map(&:options).flatten.sort_by(&:name)
+          documentation = {}
+          options.each do |option|
+            documentation[option.name] = option_docstrings(option) if option.docstring
           end
+          docstrings = []
+          documentation.each_value { |value| docstrings.concat(value) }
           docstrings
         end
 
@@ -66,6 +69,24 @@ module Smithy
             parent.const_set(child, ::Module.new) unless parent.const_defined?(child)
             parent = parent.const_get(child)
           end
+        end
+
+        def option_docstrings(option)
+          docstrings = []
+          docstrings << option_tag(option)
+          documentation = option.docstring.split("\n").map { |line| " #{line}" }
+          docstrings.concat(documentation)
+          docstrings
+        end
+
+        def option_tag(option)
+          tag = StringIO.new
+          tag << '@option options'
+          tag << " [#{option.doc_type}]" if option.doc_type
+          tag << " :#{option.name}"
+          default = option.doc_default || option.default
+          tag << " (#{default})" if default
+          tag.string
         end
 
         # @api private
--- a/gems/smithy/lib/smithy/views/client/client.rb
+++ b/gems/smithy/lib/smithy/views/client/client.rb
@@ -25,16 +25,17 @@ module Smithy
         end
 
         def require_plugins
-          @plugins.select(&:requirable?).map(&:require_path)
+          @plugins.map do |plugin|
+            "require#{'_relative' if plugin.require_relative?} '#{plugin.require_path}'"
+          end
         end
 
         def add_plugins
-          @plugins.reject(&:default?).map(&:class_name)
+          @plugins.map(&:class_name)
         end
 
         def docstrings
           docstrings = []
-          docstrings << '@param [Hash] options'
           # TODO: ensure correct handling of duplicate option definitions
           @plugins.each do |plugin|
             docstrings.concat(plugin.docstrings)
@@ -56,7 +57,7 @@ module Smithy
           code_generated_plugins.each do |plugin|
             Object.module_eval(plugin.source)
           end
-          PluginList.new(plan).to_a + code_generated_plugins.to_a
+          PluginList.new(plan, code_generated_plugins)
         end
 
         def define_module_names
--- a/gems/smithy/lib/smithy/views/client/client.rb
+++ b/gems/smithy/lib/smithy/views/client/client.rb
@@ -8,7 +8,7 @@ module Smithy
         def initialize(plan, code_generated_plugins)
           @plan = plan
           @model = plan.model
-          @plugins = plugins(plan, code_generated_plugins)
+          @plugins = PluginList.new(plan, code_generated_plugins)
           super()
         end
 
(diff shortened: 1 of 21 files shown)


projections/weather/lib/weather/client.rb
```diff
@@ -10,10 +14,17 @@ module Weather
...
 
-    add_plugin(Smithy::Client::Plugins::NetHTTP)
     add_plugin(Weather::Plugins::Endpoint)
+    add_plugin(Smithy::Client::Plugins::Logging)
+    add_plugin(Smithy::Client::Plugins::RaiseResponseErrors)
+    add_plugin(Smithy::Client::Plugins::ResponseTarget)
+    add_plugin(Smithy::Client::Plugins::NetHTTP)
 
-    # @param [Hash] options
+    # @option options [Weather::EndpointProvider] :endpoint_provider
```
mullermp, line 25: We could stable sort these alphabetically.

gems/smithy/lib/smithy/weld.rb
```diff
@@ -34,25 +35,34 @@ def post_process(artifacts)
...
     # Called when constructing the client. Any plugins defined here will be merged
-    # with other plugins.
-    # @return [Hash<String, String>] a mapping of fully qualified class names as the
-    #  key, and the plugin's file location as the value.
-    # def plugins
-    #   {}
-    # end
+    # with other plugins. The key is the fully qualified class name of the plugin,
+    # and the value is a hash with any of the following keys:
+    # * :source - the source code of the plugin
```
alextwoods, line 57: I feel like the passing of random hashes for these makes it harder for users.  Why not a class?

gems/smithy/lib/smithy/welds.rb
```diff
@@ -1,18 +1,18 @@
...
 require_relative 'welds/endpoints'
+require_relative 'welds/plugins'
 require_relative 'welds/rubocop'
 
 module Smithy
   # @api private
   module Welds
     @welds = {}
 
+    # Weld.subclasses should be ordered by insertion order. Reverse the order
```
alextwoods: How is insertion order controlled?

gems/smithy/lib/smithy/weld.rb
```diff
@@ -34,25 +35,34 @@ def post_process(artifacts)
...
+    # * :path - the path to the plugin
```
alextwoods: Whast the difference between path and require path?

gems/smithy/lib/smithy/welds/plugins.rb
```diff
@@ -0,0 +1,40 @@
...
+    class Plugins < Weld
+      def for?(_service)
+        say_status :insert, 'Adding default plugins', @plan.quiet
+        true
+      end
+
+      # rubocop:disable Metrics/MethodLength
+      def plugins
+        {
+          'Smithy::Client::Plugins::Logging' => {
```
alextwoods: I'd almost argue that a weld should be responsible for ensuring all of the plugins its exposing are already loaded (ie, it should require all of them) -  I think we could then use the Class instead of a string for the name of it and then I think that simplifies getting the options from it as well.

gems/smithy/lib/smithy/views/client/client.rb
```diff
@@ -25,20 +25,23 @@ def gem_version
...
+          @plugins.map(&:class_name)
         end
 
         def docstrings
-          docstrings = []
-          docstrings << '@param [Hash] options'
-          # TODO: ensure correct handling of duplicate option definitions
-          @plugins.each do |plugin|
-            docstrings.concat(plugin.docstrings)
+          options = @plugins.map(&:options).flatten.sort_by(&:name)
```
alextwoods, line 38: How does this handle the order of multiple plugins adding the same option?

alextwoods (approved):
Nice - I like the default plugins -> weld approach.  I think that simplifies/removes special casing nicely.

gems/smithy/lib/smithy/weld.rb
mullermp: We could add a class later, sure.

gems/smithy/lib/smithy/welds.rb
mullermp: By when the class is defined. You should verify this. ObjectSpace had no such guarantee.

gems/smithy/lib/smithy/weld.rb
mullermp: Path is where the plugin lives, and require path is how it's required in client.

gems/smithy/lib/smithy/views/client/client.rb
mullermp: I believe it is ordered by weld and then by plugin. I verified this by adding a duplicate option and it was able to override. You should also do a verification.

gems/smithy/lib/smithy/welds/plugins.rb
mullermp: I had this originally but I wasn't sure. I guess that means we don't need to worry about how it gets required when generating options documentation or rbs?

alextwoods: Yeah - I think it would simplify that code and I think its a reasonable requirement of a weld

gems/smithy/lib/smithy/welds.rb
alextwoods: Haven't done a deep dive on Active Support's DescendantsTracker is that there are no specific guarantees, particularly across different ruby implementations (jruby, truffle, ect).

Long term we'll probably need some mechanism to actually control this ordering, but out of scope for now.

mullermp: This is not from active support, this is native ruby https://ruby-doc.org/core-3.1.0/Class.html#method-i-subclasses though it says the order is not guaranteed. In testing it appeared to be. I'm not sure what to do then.

alextwoods: Ah fair - but yeah either way, order is not guaranteed.   Long term I think we want/need some ability to control order.  There are a lot of ways we can do that: 1) `priority` field (as we did for RubyIntegrations and we do still for handlers) 2) some sort of configuration that allows order to be specified.  3) Some sort of relative ordering (through configuration or specified as an attirbute of a given weld).

But we don't need to figure that out now I think.

mullermp: Insertion order may not even be correct either because it can depend on whether you define the weld before or after you require smithy/welds.

gems/smithy/lib/smithy/welds/plugins.rb
```diff
@@ -9,32 +14,22 @@ def for?(_service)
         true
       end
 
-      # rubocop:disable Metrics/MethodLength
       def plugins
         {
-          'Smithy::Client::Plugins::Logging' => {
-            path: 'smithy-client/plugins/logging',
-            relative_path: true,
+          Smithy::Client::Plugins::Logging => {
```
alextwoods, line 19: Nice! I like this approach

alextwoods (approved):

PR #267, opened 2025-02-04
ClientHelper / SpecHelper improvement with source generation tests

mullermp:
Refactor spec helper to understand how to generate clients, types, servers, and use a client helper which uses dynamic or static code generation. Also adds source code tests to component tests.


gems/smithy/spec/support/client_helper.rb
```diff
@@ -0,0 +1,185 @@
+# frozen_string_literal: true
+
+require 'tmpdir'
+
+module ClientHelper
+  class << self
+    def sample_shapes
```
alextwoods, line 7: I think this model should work the same way our other smithy models do - we should have a model.smithy + model.json and we load that here. 
Its terrible to work in smithy json.   Its even worse to work in ruby hashes that are trying to look like json.

gems/smithy/spec/support/client_helper.rb
```diff
@@ -0,0 +1,185 @@
...
+
+      if ENV.fetch('SMITHY_RUBY_KEEP_GENERATED_SOURCE', 'false') == 'true'
+        puts "Leaving generated service in: #{tmpdir}"
+      else
+        FileUtils.rm_rf(tmpdir)
+      end
+      $LOAD_PATH.delete("#{tmpdir}/lib")
+    end
+
+    def source(type, options = {})
```
alextwoods: I find source a slightly weird name for this since we are doing the eval on it and not returning source.

gems/smithy/spec/support/client_helper.rb
```diff
@@ -0,0 +1,185 @@
...
+        module_name: next_sample_module_name,
+        gem_version: options.fetch(:gem_version, '0.1.0'),
+        quiet: ENV.fetch('SMITHY_RUBY_QUIET', 'true') == 'true'
+      }
+      plan = Smithy::Plan.new(model, type, plan_options)
+      source = Smithy.source(plan)
+      Object.module_eval(source)
+      Object.const_get(plan.module_name)
+      [plan.module_name]
+    rescue Exception => e # rubocop:disable Lint/RescueException
```
alextwoods: We may actually want to divide the error handling into generating code and evaluating that code.

gems/smithy/spec/support/client_helper.rb
```diff
@@ -0,0 +1,185 @@
...
+        module_name: module_names.join('::'),
+        gem_version: options.fetch(:gem_version, '0.1.0'),
+        destination_root: options.fetch(:destination_root, Dir.mktmpdir),
+        quiet: ENV.fetch('SMITHY_RUBY_QUIET', 'true') == 'true'
+      }
+      plan = Smithy::Plan.new(model, type, plan_options)
+      tmpdir = plan.destination_root
+      Smithy.generate(plan)
+      $LOAD_PATH << ("#{tmpdir}/lib")
+      require plan.gem_name
```
alextwoods: In theory maybe we should also clean this up off the load path?

gems/smithy/spec/interfaces/weld_spec.rb
```diff
@@ -43,7 +43,7 @@ def post_process(artifacts)
   # rubocop:enable Lint/UselessAssignment
 
   after(:all) do
-    SpecHelper.cleanup(['Weather'], @tmpdir)
+    SpecHelper.cleanup_generated(['Weather'], @tmpdir)
```
alextwoods: This isn't directly related to this rectoring, but is something I've been thinking about - rather than needing to do this hacky before(:all)/after(:all) - we could have a with_generated method that would setup  a before and after block for us.

mullermp: I think shared context might work for that, but I don't like wrapping all of the tests in yet another block. I think it would be nice to collect all of the generated clients and handle them at clean up automatically, perhaps by having state in the helper.

gems/smithy/spec/support/client_helper.rb
mullermp: That is done already in cleanup_generated

gems/smithy/spec/support/client_helper.rb
mullermp: Yes, it's really just a sourced client. I guess this could be "from_source" and "from_files"

gems/smithy/spec/support/client_helper.rb
mullermp: The advantage to having the model json in code, prior to generation, is that we can modify contents in the test. For example v3 code has tests like:

```
        shapes['StructureShape']['members']['String']['locationName'] = 'str'
        expect(json(string: 'abc')).to eq('{"str":"abc"}')
```

alextwoods: Yeah I like that ability - I guess I'm not suggesting that we can't have that - just that where the default comes from is a smithy model rather than an in code hash which is just kinda ugly and super annoying to edit.

mullermp: Ideally this is only written once, but yeah we can have a fixture default I suppose, but we need a way to modify it at runtime I think.

gems/smithy/spec/interfaces/weld_spec.rb
```diff
@@ -37,34 +37,57 @@ def post_process(artifacts)
...
   end
 
-  it 'can pre process the model' do
-    weld = Weather::Types::Weld.new
-    expect(weld).to be_a(Struct)
-    expect(weld.members).to be_empty
-    get_forecast_output = Weather::Types::GetForecastOutput.new
-    expect(get_forecast_output.members).to include(:chance_of_welds)
+  context 'generated client gem' do
+    include_context 'generated client gem', { fixture: 'weather' }
```
alextwoods, line 48: Nice!

gems/smithy/spec/interfaces/weld_spec.rb
```diff
@@ -37,34 +37,57 @@ def post_process(artifacts)
...
-    expect(other_weld).to be_a(Struct)
+  context 'generated schema gem' do
+    it 'pending'
   end
 
-  it 'does not apply welds that return false in #for?' do
-    expect(defined?(Weather::Types::WeldShouldNotExist)).to be nil
-    expect(defined?(Weather::Types::OtherWeldShouldNotExist)).to be nil
+  context 'source code' do
+    include_context 'generated client from source', { fixture: 'weather' }
```
alextwoods: I still struggle with this naming.  What about "generated in-memory client"

gems/smithy/spec/spec_helper.rb
```diff
@@ -73,6 +60,38 @@ def create_plan(modules, model, type, options)
       }
       Smithy::Plan.new(model, type, plan_options)
     end
+
+    ## end temporary
+
+    def generate_client_gem(options = {})
+      plan = ClientHelper.generate(:client, options)
+      RbsSpyTest.setup(modules, sdk_dir) if ENV.fetch('SMITHY_RUBY_RBS_TEST', false)
```
alextwoods: Does this work? Where does `modules` get defined?

gems/smithy/spec/spec_helper.rb
```diff
@@ -73,6 +60,38 @@ def create_plan(modules, model, type, options)
...
+      $LOAD_PATH << "#{plan.destination_root}/lib"
+      require plan.gem_name
+      RbsSpyTest.setup(plan.module_name, plan.destination_root) if ENV['SMITHY_RUBY_RBS_TEST']
+      plan
+    rescue LoadError => e
+      puts "Error loading gem: #{plan.gem_name}"
+      raise e
+    end
+
+    def generate_client_from_source(options = {})
```
alextwoods: from source I still find confusing  maybe `generate_in_memory_client` or something like that.

gems/smithy/spec/support/client_helper.rb
```diff
@@ -0,0 +1,182 @@
...
+        'smithy.ruby.tests#Union' => {
+          'type' => 'union',
+          'members' => {
+            'string' => { 'target' => 'smithy.api#String' }
+          }
+        }
+      }
+    end
+
+    # @param [Array<String>] module_names A list of module names for the
```
alextwoods: Is there a reason we need this as an array insead of as just a full name eg "Company::Weather"

gems/smithy/spec/support/client_helper.rb
```diff
@@ -0,0 +1,182 @@
...
+      plan = create_plan(model, type, options)
+      source = Smithy.source(plan)
+      [plan.module_name, source]
+    end
+
+    def cleanup_gem(module_name, tmpdir = nil)
+      undefine_module(module_name)
+      return unless tmpdir
+
+      if ENV.fetch('SMITHY_RUBY_KEEP_GENERATED_SOURCE', 'false') == 'true'
```
alextwoods, line 124: More general comment on how we're checking ENV flags.  Here we are expecting a case sensitive "true" only.  In my RBS rpy setup I only check if its set at all.   These are only used for our generation and not for customers configuring an SDK so its probably less critical, but we should probably be consistent.

gems/smithy/spec/spec_helper.rb
mullermp: It doesn't work. Sloppy editing.

gems/smithy/spec/support/client_helper.rb
mullermp: I am going for full name. The docs are not correct (sloppy editing)

gems/smithy/spec/support/client_helper.rb
mullermp: I am preferring downcasing. For customer experience we would probably force it downcase.

gems/smithy/spec/spec_helper.rb
```diff
@@ -1,77 +1,47 @@
...
 
-    def create_plan(modules, model, type, options)
-      plan_options = {
-        module_name: modules.join('::'),
-        gem_version: options.fetch(:gem_version, '0.1.0'),
-        destination_root: options.fetch(:destination_root, Dir.mktmpdir),
-        quiet: ENV.fetch('SMITHY_RUBY_QUIET', 'true') == 'true'
-      }
-      Smithy::Plan.new(model, type, plan_options)
+    def cleanup_eval_source(module_name)
```
alextwoods: this reads a little weirdly to me - isn't this cleaning up modules? could we just call it "cleanup_modules" or "remove_generated_modules"

alextwoods (approved):
Nice - looks good - I think nicely cleans up our specs!

PR #269, opened 2025-02-10
Flatten shapes when fetching with Model.shape

alextwoods:
See @mullermp comment (hijacked PR)


gems/smithy/lib/smithy/model.rb
```diff
@@ -36,16 +36,99 @@ module Model
...
+        else
+          raise ArgumentError, "Shape not found: #{target}"
+        end
+      end
+
+      # resolve mixins and "apply" types.
+      def preprocess(model)
+        # TODO: Deep dup of model
+        model['shapes'].each do |id, shape|
+          flatten_shape(id, model, shape)
```
jterapin: Since we are flattening the shape here (bringing members and its traits from the mixin shape and applying to this specific shape), do we keep around the mixin shape? 

I'm guessing once all the shapes that uses mixins are resolved - the shapes that are marked with mixins are not quite useful since they already fulfilled their purpose.  I'm not sure if we opt to remove them or skip over them during codegeneration (check if `mixin` trait is applied, etc)

alextwoods: Oh yeah, thats actually a good point.  The `flattten_shape` ends up removing the `mixin` property of shapes that has flattened, but the shapes themselves remain on the model.  Probably makes sense to remove those mixins as well.  In theory they shouldn't get picked up when we're using the service shape closure, but they could be picked up if we don't have a service (depending on how we end up implementing that down the road).

mullermp:
Returns a flattened shape by default (resolving mixins and apply) when using Model.shape.

Uses depth first recursion. Merging is based on [DeepMergable](https://github.com/rails/rails/blob/cf6ff17e9a3c6c1139040b519a341f55f0be16cf/activesupport/lib/active_support/deep_mergeable.rb)

gems/smithy/lib/smithy/model/flattener.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Model
```
jterapin, line 4: I _think_ we were considering either renaming this module name so it doesn't share namespaces with `smithy-model`  - I tried to dig in [smithy docs](https://smithy.io/2.0/spec/model.html) and internal docs  for ideas:
* Smithy::JsonModel (based on the fact that we are working with JSON AST)
* Smithy::DataModel 
* Smithy::ApiModel (I think you mentioned this previously in a discussion)

Could also go back and rename smithy-model gem into smithy-shapes since the gem contains building blocks to create shapes? Maybe not a good idea since "shapes" are very overloaded....

gems/smithy/lib/smithy/model/flattener.rb
```diff
@@ -0,0 +1,68 @@
...
+          elsif v1.is_a?(Array) && v2.is_a?(Array) && context != 'traits'
+            # Merge arrays, but only if the key is not a trait
+            v1 + v2
+          else
+            v2
+          end
+        end
+      end
+
+      def apply_member_traits(id, shape)
```
jterapin: `shape.fetch('members', [])`  will work for shapes like structures and union. I think members for list (`member`) and map shapes (`key`, `value`) are defined differently so need to cover those cases, right? 

I forget if this is already resolved from the get-go.

gems/smithy/lib/smithy/model/flattener.rb
mullermp: I will rename this module in another commit, probably named API. I think smithy-model is correct for types and shapes. We could consider smithy-api or smithy-schema for the other gem.

gems/smithy/lib/smithy/model/flattener.rb
mullermp: I don't know if mixins work for those shapes or not but that's a good call out, I will investigate.

Rakefile
```diff
@@ -59,19 +59,27 @@ namespace :smithy do
 
   desc 'Convert all fixture smithy models to JSON AST representation.'
   task 'sync-fixtures' do
+    smithy_build_files = Dir.glob('gems/smithy/spec/fixtures/**/smithy-build.json')
     Dir.glob('gems/smithy/spec/fixtures/**/model.smithy') do |model_path|
       out_path = model_path.sub('.smithy', '.json')
-      sh("smithy ast --aut #{model_path} > #{out_path}")
+      config_files = smithy_build_files.map do |file|
```
alextwoods, line 65: Would there ever be multiple smithy build files? That seems maybe invalid?

gems/smithy/lib/smithy/model.rb
```diff
@@ -37,15 +38,15 @@ module Model
...
-        model['shapes'][target]
-      elsif PRELUDE_SHAPES.key?(target)
-        PRELUDE_SHAPES[target]
+    # @param [String] id Shape ID
+    # @return [Hash]
+    def self.shape(model, id)
+      if model['shapes'].key?(id)
+        Flattener.new(model).shape(id)
+      elsif PRELUDE_SHAPES.key?(id)
+        PRELUDE_SHAPES[id]
```
alextwoods, line 47: Is there any possible interaction between mixins and prelude shapes? I don't think there would be, but just want to check.

gems/smithy/lib/smithy/model/flattener.rb
```diff
@@ -0,0 +1,68 @@
...
+
+      def shape(id)
+        shape = @model['shapes'][id]
+        return shape unless shape['mixins']
+
+        shape['mixins'].reverse_each do |mixin|
+          mixin_shape = shape(mixin['target'])
+          shape = deep_merge(mixin_shape, shape, exclude_traits(mixin_shape))
+          apply_member_traits(id, shape)
+          shape.delete('mixins')
```
alextwoods, line 19: Should this be configurable? I could see wanting to resolve the flattened shape, but leave in place mixins (thinking of cases for mixin interface)

gems/smithy/lib/smithy/model/flattener.rb
```diff
@@ -0,0 +1,68 @@
...
+      def deep_merge(hash1, hash2, exclude_traits = [], context = nil)
+        hash1 = hash1.dup
+        if hash1['traits']
+          hash1['traits'] = hash1['traits'].except(*exclude_traits)
+          hash1.delete('traits') if hash1['traits'].empty?
+        end
+        deep_merge!(hash1, hash2, exclude_traits, context)
+      end
+
+      def deep_merge!(hash1, hash2, exclude_traits, context)
```
alextwoods, line 43: The difference between deep_merge and deep_merge! is confusing - maybe these could use comments?

alextwoods:
Changes overall look good but its my own PR so can't leave an approve.

jterapin (approved):
I forgot to approve during my review so... here it is!

Rakefile
mullermp: The AST command takes a config file and it says multiple ones are merged. The intent was to allow this to be at all folder hierarchies.

gems/smithy/lib/smithy/model.rb
mullermp: I don't think so but I'll check.

gems/smithy/lib/smithy/model/flattener.rb
mullermp: This code returns a new hash. I think the shape() method can have flattened false option and it would just return the shape from the model without flattening. For now, I have it always happening. Wouldn't mixin interfaces not need a resolved shape? I would assume just the shapes defined and then an "include" on the mixin. This can always be refactored later.

gems/smithy/lib/smithy/model/flattener.rb
mullermp: One returns a new hash and the other modifies the current hash, just like merge and merge!. I got inspiration from the rails implementations. I can add that here.

gems/smithy/lib/smithy/model.rb
mullermp: I tried this: `Cannot apply smithy.api#mixin to an immutable prelude shape defined in
smithy.api.`

PR #270, opened 2025-02-11
Generate specs for protocol tests

alextwoods:
*Description of changes:*
Diffing against cbor protocol branch.

Generates protocol tests.  Full support for request, response and error tests. Most test code follows the patterns of codegen protocol test from java smithy-ruby.  

Steps to run tests:
Run the new `smithy:spec:protocols` command (or run `smithy:spec` to do everything)

## Steps for manually testing
To generate protocol tests:
```
bundle exec smithy-ruby smith client --gem-name rpcv2cbor --gem-version 1.0.0 <<< $(cat gems/smithy/spec/fixtures/protocol_tests/rpcv2_cbor/model.json)
```

Then to run them, from your IDE open up the `build/smithy/source/smithy-ruby` directory and run the tests from `build/smithy/source/smithy-ruby/spec/rpcv2cbor/protocol_spec.rb`


Rakefile
```diff
@@ -55,14 +55,62 @@ namespace :smithy do
     plans.each { |plan| SpecHelper.cleanup_gem(plan) }
   end
 
-  task 'spec' => %w[spec:unit spec:endpoints]
+  task 'spec:protocols', [:rbs_test] do |_t, args|
```
mullermp: This code looks very similar... Any refactor opportunities?

Rakefile
```diff
@@ -55,14 +55,62 @@ namespace :smithy do
...
   desc 'Convert all fixture smithy models to JSON AST representation.'
   task 'sync-fixtures' do
     Dir.glob('gems/smithy/spec/fixtures/**/model.smithy') do |model_path|
       out_path = model_path.sub('.smithy', '.json')
       sh("smithy ast --aut #{model_path} > #{out_path}")
     end
+
+    Dir.glob('gems/smithy/spec/fixtures/protocol_tests/**/smithy-build.json') do |smithy_build_path|
+      Dir.chdir(File.dirname(smithy_build_path)) do
+        sh('smithy ast --flatten > model.json')
```
mullermp: Is this supposed to take a file for an argument somewhere?

gems/smithy-client/lib/smithy-client/cbor/value_matcher.rb
```diff
@@ -0,0 +1,66 @@
+# frozen_string_literal: true
```
mullermp: I'm hesitant for this to live in smithy client because it would be requirable at runtime. What other options exist?

gems/smithy-client/lib/smithy-client/cbor/value_matcher.rb
```diff
@@ -0,0 +1,66 @@
...
+# Provides an rspec matcher for CBOR encoded values
+# rubocop:disable Metrics/BlockLength
+RSpec::Matchers.define :match_cbor do |expected|
+  match do |actual|
+    # identical values don't need more comparison
+    return true if actual == expected
+
+    expect(actual.class).to eq(expected.class)
+
+    def match_hash(actual, expected)
```
mullermp: Is it common to define methods like this in the match block or should these be scoped up under the matcher definition?

gems/smithy-client/lib/smithy-client/cbor/value_matcher.rb
```diff
@@ -0,0 +1,66 @@
+# frozen_string_literal: true
+
+require 'rspec/expectations'
+
+def expected_nan?(expected)
+  expected == 'NaN' || expected.nan?
+end
+
+def expected_infinite?(expected)
```
mullermp: Couldn't match_cbor handle these generally? Shouldn't these be defined with the other methods?

gems/smithy-client/lib/smithy-client/protocols/rpc_v2.rb
```diff
@@ -27,7 +27,7 @@ def build(context)
         def parse(context)
           output_shape = context.operation.output
           codec = Client::Codecs::CBOR.new(setting(context))
-          codec.deserialize(context.response.body.read, output_shape, output_shape.type)
+          codec.deserialize(context.response.body.read, output_shape, output_shape.type.new)
```
mullermp: Possibly an issue with codec interface. If output shape is already passed in, why also type? And maybe codec should do that initialization.

gems/smithy/lib/smithy/generators/client.rb
```diff
@@ -94,6 +95,12 @@ def code_generated_plugins
       def should_skip_customizations?
         Dir["#{destination_root}/**/*"].any? { |f| f.include?('/customizations.rb') }
       end
+
+      def protocol_tests?
```
mullermp: I'm assuming you did this to avoid generating an empty file. I don't like knowledge of these traits in the orchestrator. I would be ok with generating an empty file with maybe a single default test. We should smoke test at least something.

gems/smithy/lib/smithy/templates/client/protocol_spec.erb
```diff
@@ -0,0 +1,164 @@
...
+
+require_relative '../spec_helper'
+<% additional_requires.each do |r| -%>
+require '<%= r %>'
+<% end -%>
+
+
+module <%= module_name %>
+  # TODO: Can be replaced by stub_responses config once implemented
+  class StubSend < Smithy::Client::Plugin
```
mullermp, line 10: You could reuse the dummy send handler in spec helper couldn't you?

gems/smithy/lib/smithy/templates/client/protocol_spec.erb
```diff
@@ -0,0 +1,164 @@
...
+  describe Client do
+    before(:all) { Client.add_plugin(StubSend) }
+    after(:all) { Client.remove_plugin(StubSend) }
+
+    let(:client_options) do
+      {
+        # stub_responses: true,
+        # validate_input: false,
+        endpoint: 'http://127.0.0.1',
+        # retry_strategy: # disable?
```
mullermp, line 32: Stub responses plugin would disable retry.

gems/smithy/lib/smithy/templates/client/protocol_spec.erb
```diff
@@ -0,0 +1,164 @@
...
+
+<% all_operation_tests.each do |operation_tests| %>
+    describe '#<%= operation_tests.name %>' do
+<% unless operation_tests.request_tests.empty? -%>
+      describe 'requests' do
+<% operation_tests.request_tests.each do |test| %>
+<% test.comments.each do |line| -%>
+        # <%= line %>
+<% end -%>
+<% if test.skip? -%>
```
mullermp, line 45: I think you can clean this up by using an <%= and render extra string if test.skip? Or does that not work?

gems/smithy/lib/smithy/templates/client/protocol_spec.erb
```diff
@@ -0,0 +1,164 @@
...
+        it '<%= test.id %>', skip: '<%= test.skip_reason %>' do
+<% else -%>
+        it '<%= test.id %>' do
+<% end -%>
+<% if test['host'] -%>
+          client = Client.new(client_options.merge(endpoint: '<%= test.test_case.fetch('authScheme', 'http') %>://<%= test['host'] %>'))
+<% end -%>
+<% if test.idempotency_token_trait? -%>
+          allow(SecureRandom).to receive(:uuid).and_return('00000000-0000-4000-8000-000000000000')
```
mullermp, line 54: Is there a better way to match expectations here? This is very implementation specific. Do the protocol tests also force this value?

gems/smithy/lib/smithy/templates/client/protocol_spec.erb
```diff
@@ -0,0 +1,164 @@
...
+<% test.comments.each do |line| -%>
+        # <%= line %>
+<% end -%>
+<% if test.skip? -%>
+        it '<%= test.id %>', skip: '<%= test.skip_reason %>' do
+<% else -%>
+        it '<%= test.id %>' do
+<% end -%>
+<% if test['host'] -%>
+          client = Client.new(client_options.merge(endpoint: '<%= test.test_case.fetch('authScheme', 'http') %>://<%= test['host'] %>'))
```
mullermp: I think you can just merge client options because client has not been referenced or loaded yet. I would also consider putting endpoint in a method for the test or test case.

gems/smithy/lib/smithy/templates/client/protocol_spec.erb
```diff
@@ -0,0 +1,164 @@
...
+          allow(SecureRandom).to receive(:uuid).and_return('00000000-0000-4000-8000-000000000000')
+<% end -%>
+          resp = client.<%= operation_tests.name %>(<%= test.params %>)
+          request = resp.context.request
+          expect(request.http_method).to eq('<%= test['method'] %>')
+          expect(request.endpoint.path).to eq('<%= test['uri'] %>')
+<% if test['resolvedHost'] -%>
+          expect(request.endpoint.host).to eq('<%= test['resolvedHost'] %>')
+<% end %>
+<% if test.query_expect? -%>
```
mullermp, line 63: Should we move query matching to a matcher? Didn't we do this with the old implementation?

gems/smithy/lib/smithy/templates/client/protocol_spec.erb
```diff
@@ -0,0 +1,164 @@
...
+<% end -%>
+          response = { status_code: <%= test['code'] %> }
+<% if test['headers'] -%>
+          response[:headers] = <%= test['headers'] %>
+<% end -%>
+<% if test['body'] -%>
+          response[:body] = <%= test.stub_body %>
+<% end -%>
+
+          client = Client.new(client_options.merge(stub_response: response, validate_params: false))
```
mullermp, line 118: What is failing validation?

gems/smithy/lib/smithy/templates/client/protocol_spec.erb
```diff
@@ -0,0 +1,164 @@
...
+          expect(request.headers['<%= k %>']).to eq('<%= v %>')
+<% end -%>
+<% test['forbidHeaders']&.each do |k| -%>
+          expect(request.headers.key?('<%= k %>')).to be(false)
+<% end -%>
+<% test['requireHeaders']&.each do |k| -%>
+          expect(request.headers.key?('<%= k %>')).to be(true)
+<% end %>
+<% if test['body'] -%>
+          <%= test.body_expect %>
```
mullermp, line 93: I think we should define common matchers so code generation is more straightforward. It avoids complex cases in the view or having the view class build strings. For example, match query, match headers, match body.

gems/smithy/lib/smithy/views/client/endpoint_provider_spec.rb
```diff
@@ -87,50 +87,16 @@ def build_client_params(data)
 
           def build_operation_params(data, input)
             data.fetch('operationParams', {}).map do |k, v|
-              member_shape = @model['shapes'][input['members'][k]['target']]
-              Param.new(k.underscore, transform_operation_values(v, member_shape))
+              member_shape = Model.shape(@model, input['members'][k]['target'])
+              Param.new(k.underscore, ShapeToHash.transform_value(@model, v, member_shape))
```
mullermp, line 91: Should we also format these hashes?

gems/smithy/lib/smithy/views/client/protocol_spec.rb
```diff
@@ -0,0 +1,239 @@
...
+  module Views
+    module Client
+      # @api private
+      class ProtocolSpec < View
+        PROTOCOL_TEST_TRAITS = %w[smithy.test#httpRequestTests smithy.test#httpResponseTests].freeze
+
+        def initialize(plan)
+          @plan = plan
+          @model = plan.model
+          @all_operation_tests = Model::ServiceIndex
```
mullermp, line 13: Nit: operation tests

gems/smithy/lib/smithy/views/client/protocol_spec.rb
```diff
@@ -0,0 +1,239 @@
...
+
+          attr_reader :test_case
+
+          def [](key)
+            test_case[key]
+          end
+
+          def comments
+            if test_case['documentation']
+              test_case['documentation'].split("\n")
```
mullermp: Fetch()

gems/smithy/lib/smithy/views/client/shape_to_hash.rb
```diff
@@ -0,0 +1,76 @@
...
+            when 'Infinity' then CodegenFloat.new(Float::INFINITY)
+            when '-Infinity' then CodegenFloat.new(-Float::INFINITY)
+            when 'NaN' then CodegenFloat.new(Float::NAN)
+            else
+              value
+            end
+          end
+        end
+
+        class CodegenFloat
```
mullermp: You probably don't need this class anymore? The logic above can handle this easily..

Rakefile
alextwoods: No - the key is that its using the local smithy-build.json which has a transform that selects the cbor protocol test service.

gems/smithy-client/lib/smithy-client/cbor/value_matcher.rb
alextwoods: This is what we did in Hearth so I followed that pattern rather than trying to reinvent the wheel.

It is requirable at runtime, but not required anywhere within the gem, so its not default available for users and wouldn't affect their runtime memory usage.

Its a required *test* dependency for gems that are genreated with protocol tests.  Our options:
1. Create a Smithy-test gem that has these sort of depenencies.  Could move client helper there as well.
2. Codegen the matcher into the spec itself.

Thoughts or preferences?

gems/smithy-client/lib/smithy-client/cbor/value_matcher.rb
alextwoods: I was copying prior art in Hearth.  All of our various matchers followed this pattern.

gems/smithy-client/lib/smithy-client/cbor/value_matcher.rb
alextwoods: I can actually remove this now that I fixed the upstream issue with code generating hashses with infinite/nan float values.

gems/smithy-client/lib/smithy-client/protocols/rpc_v2.rb
alextwoods: Lets discuss on protocol implementation PR.

gems/smithy/lib/smithy/generators/client.rb
alextwoods: It feels a little weird to generate an empty protocol test file for most gems.   We could move the trait knowledge to some utility class.

gems/smithy/lib/smithy/templates/client/protocol_spec.erb
alextwoods: What dummy send handler - I couldn't find it? It also has the issue that it would need to be available as a test dependency for generated gems.  Code generating a small plugin as a stop gap until we have stup_responses here seems easier than trying to take more test dependencies.

gems/smithy/lib/smithy/templates/client/protocol_spec.erb
alextwoods: I don't think it did in Hearth?  Anyway, I can remove this if you prefer.  It was only in here as a reminder that when we siwtch to stub_resposnes to ensure that retrys are disabled (however that ends up happening).

gems/smithy/lib/smithy/templates/client/protocol_spec.erb
alextwoods: I could, but it made for a long messy looking line.

gems/smithy/lib/smithy/templates/client/protocol_spec.erb
alextwoods: Prior art from Hearth (and I think v3).

The protocol tests force and expect this value.

gems/smithy/lib/smithy/templates/client/protocol_spec.erb
alextwoods: We could move it to a matcher if we want.  I was trying to avoid too many matchers when simple generated code would do.

This is also prior art from Hearth.

gems/smithy/lib/smithy/templates/client/protocol_spec.erb
alextwoods: We make requests with no parameters (since this is only testing response), that will fail whenever there is a required parameter.

gems/smithy/lib/smithy/templates/client/protocol_spec.erb
alextwoods: Was following prior art from Hearth and wanted to avoid having too many matchers for simple code.

Open to changing it, but was following our prior art.

gems/smithy/lib/smithy/views/client/protocol_spec.rb
alextwoods: What do you mean?

gems/smithy/lib/smithy/views/client/protocol_spec.rb
alextwoods: The code got messier (IMO) with fetch:

```
test_case.fetch('documentation', nil)&.split("\n")
```

gems/smithy/lib/smithy/views/client/shape_to_hash.rb
alextwoods: It cannot.  We need some class with an Inspect method that returns `Float::Infinity`.  Float does not do this.  If we put a String directly into the hash then we end up with `:floatvalue => "Float::INFINITY"` (note the extra quotes). 

The only other option is to monkey patch Float class.  I prefer this.

gems/smithy-client/lib/smithy-client/protocols/rpc_v2.rb
jterapin: Yeah I need to update my PR on this - basically, we want to only pass in type when it comes to typed documents (since it will contain a discriminator id for us to look up in the TypeRegistry). Otherwise, we should just use the type associated with the output_shape.

gems/smithy/lib/smithy/templates/client/protocol_spec.erb
jterapin: I think Matt was referring to this: https://github.com/smithy-lang/smithy-ruby/blob/decaf/gems/smithy-client/spec/spec_helper.rb#L15

I think this is fine long as there's TODO

gems/smithy/lib/smithy/views/client/shape_to_hash.rb
```diff
@@ -0,0 +1,76 @@
...
+  module Views
+    module Client
+      # @api private
+      module ShapeToHash
+        class << self
+
+          def transform_value(model, value, shape)
+            return value unless shape && value
+
+            case shape['type']
```
jterapin, line 12: Do we need to cover enum and enumInt here?

gems/smithy/lib/smithy/generators/client.rb
mullermp: Is there not a default test we could have, like test the option works?

gems/smithy/spec/fixtures/protocol_tests/rpcv2_cbor/skipped_tests.smithy
```diff
@@ -0,0 +1,8 @@
+$version: "2.0"
+
+namespace smithy.protocoltests.rpcv2Cbor
+use smithy.ruby#skipTests
+
+apply OperationWithDefaults @skipTests([
+  { id: "RpcV2CborClientPopulatesDefaultValuesInInput", reason: "Defaults not Implemented yet.", type: "request" }
```
jterapin, line 7: Ohhhh nice, we just have a single test case that we need to resolve once defaults are implemented?

gems/smithy/lib/smithy/templates/client/protocol_spec.erb
mullermp: Ok whatever makes it easier to read.

gems/smithy/lib/smithy/templates/client/protocol_spec.erb
alextwoods: I could do it only with `client_options.merge!` and to me it was harder to read that way (requires more knowledge of how the lets work/are setup.  harder to jsut read the code as is).   If you prefer that style though, I'll change it. 

good idea on the endpoint method.

gems/smithy/lib/smithy/views/client/shape_to_hash.rb
```diff
@@ -0,0 +1,76 @@
...
+              value
+            end
+          end
+        end
+
+        class CodegenFloat
+          def initialize(value)
+            @value = value
+          end
+          def inspect
```
jterapin: newline in-between?

gems/smithy/lib/smithy/views/client/protocol_spec.rb
mullermp: With a default of an empty array? ..

alextwoods: We need to apply split on the string not an array

gems/smithy/lib/smithy/generators/client.rb
alextwoods: The `protocol` configuation option you mean?

That seems like a different test - a `protocol_plugin_spec`.

Rakefile
mullermp: Right but I think it should explicitly declare the model file and possibly the config file too?

gems/smithy-client/lib/smithy-client/cbor/value_matcher.rb
mullermp: I'm asking if we should have been defining methods in the match block or not. That's like defining methods within methods which is anti pattern. I'm not sure when you would.

gems/smithy/lib/smithy/generators/client.rb
mullermp: I don't think a utility class makes sense because it's just hiding away the same problem and only used once. I think a file with no tests is reasonable or we can keep this as is

gems/smithy/lib/smithy/templates/client/protocol_spec.erb
mullermp: A matcher could say we make a uuid that's "like" the one we expect. It's not a strong assertion though. This is otherwise always saying we use secure random as an implementation which is probably fine.

gems/smithy/lib/smithy/templates/client/protocol_spec.erb
mullermp: You can't do client_options = client_options.merge()?

gems/smithy/lib/smithy/templates/client/protocol_spec.erb
mullermp: Do the response tests not include input?

gems/smithy/lib/smithy/views/client/protocol_spec.rb
mullermp: Yeah, you can default fetch to an empty string so that split always returns an array.

gems/smithy/lib/smithy/views/client/protocol_spec.rb
mullermp: All_ prefix just seemed unnecessary. It's a collection of operation tests.

Rakefile
alextwoods: Cleaned it up with a shared task w/ args.

gems/smithy/lib/smithy/views/client/protocol_spec.rb
alextwoods: Oh, right.  Updated.

gems/smithy/lib/smithy/views/client/protocol_spec.rb
alextwoods: it conflicts with `operation_tests`.  `operations_tests` felt harder to read (the difference is this is a list of tests for all operations.  operation_tests is the tests for a single operation).

gems/smithy-client/lib/smithy-client/cbor/value_matcher.rb
alextwoods: Looking back - the reason we define the methods in the block is to avoid polluting any name spaces.  

I could define some other new module, like `Smithy::Matchers` or something like that and put these methods there, but having them defined inside the block limits their scope, which is exactly what we want here.

Rakefile
alextwoods: There is no model file - the model comes from the dependency in the smithy-build.json.  but updated to reference the config file.

gems/smithy/lib/smithy/templates/client/protocol_spec.erb
alextwoods: Ah I see, was looking in Smithy.  

Yeah, the same issue applies - we don't want to move that dummy send plugin to the lib directory in smithy-client so that it can be used as a test dependency in generated gems.

gems/smithy/lib/smithy/views/client/shape_to_hash.rb
alextwoods: I don't think we should need to - in the untransformed hash they will just be strings and numbers, which will translate fine.

gems/smithy/spec/fixtures/protocol_tests/rpcv2_cbor/skipped_tests.smithy
alextwoods: I'm not sure - I think there are a few test cases - this was just a demo of how to use the trait/skip mechanism for now :-)

gems/smithy/lib/smithy/views/client/protocol_spec.rb
mullermp: I didn't see an operation_tests in this class.

alextwoods: I was using it in the view:
```
<% all_operation_tests.each do |operation_tests| %>
```

Not sure of better naming for both of these.

mullermp: I guess you could name them protocol_tests which enumerates operation_tests, but whatever is fine. Is the operation separation relevant or important?

alextwoods: Hmm - yeah thats a good question.  

Tests themselves are modeled per operation and typically rspec breaks down to describe '#operation' behaviors.  But theres no fundamental reason we couldn't just generate a full list and skip going operation by operation.  
There might be some complexity in the IDs.  I'm not 100% sure, but I think IDs are unique within request tests and response tests.  See: https://github.com/smithy-lang/smithy/blob/main/smithy-protocol-test-traits/src/main/java/software/amazon/smithy/protocoltests/traits/UniqueProtocolTestCaseIdValidator.java

So I thikn it would be fine as long as we still had seperate `describe 'requests'` and resposne blocks.

mullermp: Are there protocol tests for certain operations like for S3?

jterapin: I recall seeing service-specific protocol tests somewhere - let me find it.

jterapin: https://github.com/smithy-lang/smithy/blob/main/smithy-aws-protocol-tests/model/restXml/services/s3.smithy

PR #271, opened 2025-02-11
Use matrix rubies in CI

mullermp:
Use a matrix for testing multiple Rubies

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,26 +10,41 @@ on:
       - decaf
 
 env:
-  ruby_version: 3.3
+  ruby_versions: (3.3 3.4)
 
 jobs:
   smithy-test:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -10,105 +10,125 @@ on:
...
+        run: bundle exec rake smithy-client:spec
 
-  smithy-test-rbs:
+  smithy-client-rbs:
+    needs: [smithy-client-test]
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby: [ 3.3, 3.4 ]
```
alextwoods: Should we run RBS on multiple ruby versions?  I think its probably sufficient to run it only on latest?

alextwoods (approved):

PR #274, opened 2025-02-12
Param converting

mullermp:
Does param converting on input to coerce into types similar to what we do in v3. This is prior to any validation.

### DIFF
--- a/gems/smithy-client/lib/smithy-client/param_converter.rb
+++ b/gems/smithy-client/lib/smithy-client/param_converter.rb
@@ -165,7 +165,7 @@ module Smithy
           converter
         end
 
-        def each_base_class(shape_class, &block)
+        def each_base_class(shape_class, &)
           shape_class.ancestors.each do |ancestor|
             yield(ancestor) if @converters.key?(ancestor)
           end
@@ -194,10 +194,10 @@ module Smithy
       end
 
       add(EnumShape, String)
-      add(EnumShape, Symbol) { |sym| sym.to_s }
+      add(EnumShape, Symbol) { |sym, _| sym.to_s }
 
       add(IntegerShape, Integer)
-      add(IntegerShape, Float) { |f| f.to_i }
+      add(IntegerShape, Float) { |f, _| f.to_i }
       add(IntegerShape, String) do |str|
         Integer(str)
       rescue ArgumentError
@@ -205,7 +205,7 @@ module Smithy
       end
 
       add(IntEnumShape, Integer)
-      add(IntEnumShape, Float) { |f| f.to_i }
+      add(IntEnumShape, Float) { |f, _| f.to_i }
       add(IntEnumShape, String) do |str|
         Integer(str)
       rescue ArgumentError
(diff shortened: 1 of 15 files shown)


gems/smithy-client/lib/smithy-client/param_converter.rb
```diff
@@ -0,0 +1,252 @@
...
+            next if v.nil?
+            next unless shape.member?(k)
+
+            values[k] = member(shape.member(k), v)
+          end
+        end
+        values
+      end
+
+      def union(shape, values)
```
mullermp, line 51: Not at all confident in this - but the intent is to convert all provided unions (maybe something later removes one) but validator will catch if multiple already.

gems/smithy-client/lib/smithy-client/param_converter.rb
```diff
@@ -0,0 +1,252 @@
...
+require 'time'
+require 'tempfile'
+
+module Smithy
+  module Client
+    # @api private
+    class ParamConverter
+      include Smithy::Client::Shapes
+
+      @mutex = Mutex.new
```
alextwoods, line 15: I assume this is all effectively copied from V3? It looks mostly identical.

gems/smithy-client/lib/smithy-client/param_converter.rb
alextwoods: I think it makes sense - kick it to validator to catch multiple.

gems/smithy-client/spec/smithy-client/param_converter_spec.rb
```diff
@@ -0,0 +1,374 @@
...
+
+require 'bigdecimal'
+require 'stringio'
+
+module Smithy
+  module Client
+    describe ParamConverter do
+      describe '#convert' do
+        it 'performs a deeply nested conversion of values' do
+          client_class = ClientHelper.sample_service
```
alextwoods, line 19: Should this be a let?

gems/smithy-client/lib/smithy-client/param_converter.rb
```diff
@@ -0,0 +1,252 @@
...
+        #
+        # @param [Class<Shapes::Shape>] shape_class
+        # @param [Class] value_class
+        # @param [#call] block An object that responds to `#call`
+        #    accepting a single argument. This function should perform
+        #    the value conversion if possible, returning the result.
+        #    If the conversion is not possible, the original value should
+        #    be returned.
+        # @return [void]
+        def add(shape_class, value_class, &block)
```
alextwoods, line 119: Should this be part of the public interface?

I guess I could see a user wanting to use custom data types and registering them this way, but in general it seems a little bit of an anti-pattern to support through global state.

alextwoods (approved):

gems/smithy-client/lib/smithy-client/param_converter.rb
mullermp: Yes mostly but I cleaned up code, added extra shape support, and extra conversions.

gems/smithy-client/spec/smithy-client/param_converter_spec.rb
mullermp: It can, but it's only used once in this test, so not too useful to lazy eval.

gems/smithy-client/lib/smithy-client/param_converter.rb
mullermp: The whole class is API private. This is public for registration within the file.

PR #275, opened 2025-02-13
Promote smithy model classes to their own gem

mullermp:
Moves shapes and other model related components to their own gem. This is to allow re-use across client, servers, and even in code generation.


mullermp:
We currently have Smithy::Model for internal generation utilities. We can consider different names for either. I think Smithy::Model for the shapes and structure/union runtime is probably preferred.

gems/smithy-model/lib/smithy-model/structure.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module Smithy
-  module Client
+  module Model
```
jterapin, line 4: When I think about generated shapes-only gems (which consists of shapes and types files and etc) having a single dependency on `smithy-gem`, I imagine keeping this structure/union files in the `smithy-model` gem makes sense.

I guess alternatively - we could code-generated `structure` and `union` files in the shapes-only gem, with the utility code being all the same except for generated namespaces? I'm not sure if that is useful at all for the customer.

gems/smithy-model/lib/smithy-model/shapes.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module Smithy
-  module Client
+  module Model
```
jterapin, line 4: Since this file also has `Prelude` module, wonder if we could use the prelude shapes here for `smithy` gem usage for prelude shapes - not sure if that is a good idea or not.

jterapin (approved):
I like the approach. If everyone is in agreement, we should move forward with this update.

gems/smithy-model/lib/smithy-model/shapes.rb
mullermp: Smithy gem depends on client anyway so it would get these, but we don't need the runtime shapes I don't think, just the json definitions of them?

jterapin: Yeah that's true - `smithy` gem just needs json definitions of these prelude. I guess I was wondering if we can streamline prelude representation in a single area instead of having a set in smithy ([code](https://github.com/smithy-lang/smithy-ruby/blob/decaf/gems/smithy/lib/smithy/model.rb#L13)) and smithy-client ([code](https://github.com/smithy-lang/smithy-ruby/blob/decaf/gems/smithy-client/lib/smithy-client/shapes.rb#L254))

Alternatively - maybe we could consider including prelude as part of generated shapes like we do in V3 again.

gems/smithy-model/lib/smithy-model/structure.rb
mullermp: Yeah, I think code generating structure and union base classes is probably not necessary. I think the baseline for what belongs in smithy-model is exactly what is necessary for schema/shapes/types gem generation.

gems/smithy-model/lib/smithy-model/shapes.rb
mullermp: While these are both "prelude definitions" I think they are entirely separate. One is for navigating the json model and the other for the runtime model, so I think separate is correct here. What do you mean about the v3 prelude? The model may define primitives but c2j always gives a complete model.

PR #276, opened 2025-02-17
Use ServiceShape as the container for operations and rename Shapes to Schema

mullermp:
This change does the following refactor:

* Service shape will be the container object for runtime shapes instead of a Schema object.
* shapes.rb has been renamed to schema.rb


jterapin (approved):
LGTM

PR #278, opened 2025-02-23
Stub Responses

mullermp:
Implements stub responses similar to v3.


gems/smithy-client/lib/smithy-client/codecs/cbor.rb
```diff
@@ -45,62 +46,71 @@ def format_blob(value)
...
+              data[member_shape.name] = format_data(member_shape.shape, value)
             end
           end
         end
 
-        def parse_data(value, shape, type = nil)
+        def format_union(shape, values)
+          data = {}
+          member_shape = shape.member_by_type(values.class)
+          data[member_shape.name] = format_data(member_shape.shape, values).value
```
jterapin: I had to make some modification here since my union-test (under `cbor_spec.rb`, testing the serde lifecycle of union shapes) was failing. This behavior is similar to what we had in caffeinated smithy-ruby.

Just returning the value of the union didn't felt right, since we would want to include what member name we are using within the union. 

Previous to my change, the format_union was defined like so:
```
member_shape = shape.member_by_type(values.class)
format_data(member_shape.shape, values).value
```

gems/smithy-client/lib/smithy-client/codecs/cbor.rb
```diff
@@ -45,62 +46,71 @@ def format_blob(value)
...
+              member_shape = shape.member(key)
+              data[member_shape.name] = format_data(member_shape.shape, value)
             end
           end
         end
 
-        def parse_data(value, shape, type = nil)
+        def format_union(shape, values)
+          data = {}
+          member_shape = shape.member_by_type(values.class)
```
jterapin: I _think_ I know why you wanted to revisit our implementation of union. When it comes to formatting union data for the request, customers are forced to use the subclasses of the union's type class - right? 

I noticed I couldn't just simply pass in a hash for my union-test in lieu of the Types instance for the value param in `serialize` method

gems/smithy-client/lib/smithy-client/codecs/cbor.rb
```diff
@@ -7,10 +7,11 @@ module Client
     module Codecs
       # Codec that serializes and deserializes in CBOR format.
       # TODO:
-      #   * Support (de)serializing union shapes once union is supported
       #   * Support handling of typed documents when it is supported
       #   * Update implementation to handle event streams
-      #   * Handle query_compatible trait
+      #   * Update (de)serializing document types
+      #   * Allow user to pass in their preferred type to deserialize
+      #     If it fails, resort to deserializing type on the shape.
```
jterapin: Added TODO before I forgets. 

This is specifically for shapes that has a type representation (structures and unions). 
I think Dowling visioned something like the below:
```
result = Smithy::Client::Codecs::CBOR.deserialize(BirdShape, data, type = Crow)
=> BirdService::Types::Crow(...) # if successful
# or 
=> BirdService::Types::Bird(...)  # if failed to deserialized based on given type
```

gems/smithy-client/lib/smithy-client/param_converter.rb
```diff
@@ -51,13 +51,16 @@ def structure(shape, values)
 
       def union(shape, values)
         values = c(shape, values)
-        if values.is_a?(Schema::Union) || values.is_a?(Hash)
+        if values.respond_to?(:each_pair)
```
jterapin: I'm thinking that is route were trying to cover the case where given `values` are hash than an instance of Union's subclasses? To resolve the `format_union`, I guess you could create an subclass instance here.

Alternatively, we could convert the given subclass instance with `.to_h` and just stick with hashes

gems/smithy-client/lib/smithy-client/param_validator.rb
```diff
@@ -53,33 +49,40 @@ def valid_structure?(shape, values, errors, context)
...
-            member_shape = shape.member(name)
-            shape(member_shape.shape, value, errors, context + "[#{name.inspect}]")
-          else
-            errors << "unexpected value at #{context}[#{name.inspect}]"
+            member(shape, name, value, errors, context)
           end
         end
       end
 
       def valid_union?(shape, values, errors, context)
```
jterapin: I think we could give the user a raise when the given object is not an instance of Union's subclasses? Something like [here](https://github.com/smithy-lang/smithy-ruby/blob/main/codegen/projections/rails_json/lib/rails_json/builders.rb#L871-L872)?

gems/smithy-client/lib/smithy-client/stubbing/data_applicator.rb
```diff
@@ -0,0 +1,46 @@
...
+module Smithy
+  module Client
+    module Stubbing
+      # @api private
+      class DataApplicator
+        include Smithy::Schema::Shapes
+
+        def initialize(rules)
+          @rules = rules
+        end
```
jterapin: `@rules` are technically shapes now?

gems/smithy-schema/lib/smithy-schema/shapes.rb
```diff
@@ -18,6 +18,54 @@ def initialize(options = {})
         attr_accessor :traits
       end
 
+      # Represents an aggregate shape that has members.
+      class Structure < Shape
```
jterapin: Nice~ that cleaned up nicely.

gems/smithy-schema/spec/smithy-schema/shapes_spec.rb
```diff
@@ -29,6 +29,64 @@ module Shapes
...
+        describe '#member' do
+          it 'returns the member' do
+            subject.add_member(:foo, 'foo', StringShape.new)
+            expect(subject.member(:foo)).to be_kind_of(MemberShape)
+          end
+        end
+
+        describe '#name_by_member_name?' do
+          it 'returns true if member by member name exists' do
+            subject.add_member(:foo, 'foo', StringShape.new)
```
jterapin: Could update the member name arg to be like `fOO` to ensure that we can reference it back.

gems/smithy-client/spec/smithy-client/codecs/cbor_spec.rb
```diff
@@ -36,18 +44,32 @@ module Codecs
...
         it 'serializes and deserializes a complex data' do
           typed = typed_struct.new(s: 'foo', l: %w[foo bar], m: { 'foo' => 'bar'.dup })
-          serialized = subject.serialize(typed, structure_shape)
-          expect(subject.deserialize(serialized, structure_shape)).to eq(typed)
+          serialized = subject.serialize(structure_shape, typed)
+          expect(subject.deserialize(structure_shape, serialized)).to eq(typed)
+        end
+
+        # TODO: Find better home that tests serde lifecycle for union types
+        it 'serializes and deserializes a union data' do
```
jterapin: I made this quick test during for union parsing implementation. 

May be already covered by other specs but I added this to ensure we are correctly serde union data. If it is already covered, could remove.  I think the biggest things that this test helped was SERDE the member names since it can be formatted differently on the model.

jterapin (commented):
Nice~ 😊 

Quick review but I will do more an in-depth once stub responses are fully covered.

gems/smithy-client/lib/smithy-client/codecs/cbor.rb
mullermp: Yes I think what you have is correct. If it isn't, we can change it later (out of scope for stub responses)

gems/smithy-client/lib/smithy-client/codecs/cbor.rb
mullermp: Yes. We should investigate this further.

gems/smithy-client/lib/smithy-client/param_converter.rb
mullermp: If it responds to each_pair, it is a struct or hash. The union does not respond to either of that, but it should probably.. - we could convert everything with to_h but that seems fragile.

gems/smithy-client/lib/smithy-client/param_validator.rb
mullermp: The param validator should already be doing that in valid_union?

gems/smithy-client/lib/smithy-client/stubbing/data_applicator.rb
mullermp: Yeah, well this is technically service shape or schema.. we can consider changing all of the rules lingo to schema or use rules.

gems/smithy-client/lib/smithy-client/plugins/stub_responses.rb
```diff
@@ -0,0 +1,93 @@
...
+        option(:stubs) { {} }
+        option(:stubs_mutex) { Mutex.new }
+        option(:api_requests) { [] }
+        option(:api_requests_mutex) { Mutex.new }
+
+        def add_handlers(handlers, config)
+          return unless config.stub_responses
+
+          handlers.add(APIRequestsHandler)
+          handlers.add(StubHandler, step: :send)
```
jterapin, line 28: Could we document what each handler does? Could be helpful  as a reminder and newcomers.

gems/smithy-client/lib/smithy-client/stubbing/data_applicator.rb
```diff
@@ -0,0 +1,46 @@
...
+
+        def apply(data, stub)
+          structure(@rules, data, stub)
+        end
+
+        private
+
+        def structure(shape, data, stub)
+          data.each do |key, value|
+            stub[key] = member(shape.member(key).shape, value)
```
jterapin, line 23: From what I understand here, `DataApplicator ` applies the given data to the stub object.

Should stub created out of given stub data, be mindful of member names? Like within the ruby context, we use `foo_bar` for the member but the member name is represented as `fooBar` in the model so we do some conversion behind the scenes when building/parsing.

I think the expectation here is that given data is `fooBar` for the member name instead of `foo_bar` - if that is the case, i think `shape.member(key).shape` should actually be `shape.name_by_member_name(key).shape`?

gems/smithy/lib/smithy/templates/client/protocol_plugin.erb
```diff
@@ -58,7 +58,7 @@ module <%= module_name %>
       class Error < Smithy::Client::Handler
         def call(context)
           output = @handler.call(context)
-          output.error = context.config.protocol.error(context)
+          output.error = context.config.protocol.error(context) unless output.error
```
jterapin, line 61: Don't we want to process the error if `output.error` is not `nil`? Or this temp until we do the protocol refactoring

gems/smithy-client/lib/smithy-client/stubbing/data_applicator.rb
jterapin: Yeah, might not be a bad idea to update to using `schema` for consistency

gems/smithy-client/lib/smithy-client/stubbing/data_applicator.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    module Stubbing
+      # @api private
+      class DataApplicator
```
jterapin, line 8: I think each Stubbing-related classes could use an one-linear of what it does. While it is obvious to us, might not be for others who will work on this repo soon (and honestly i could always be reminded)

gems/smithy-client/lib/smithy-client/stubs.rb
```diff
@@ -0,0 +1,213 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    # This module provides the ability to specify the data and/or errors to
+    # return when a client is using stubbed responses.
+    # @see Plugins::StubResponses
```
jterapin: `Plugins::StubResponses` is currently set to api-private so I think this doesn't link it.

jterapin: Screenshot 2025-03-04 at 9 37 54 AM

gems/smithy-client/lib/smithy-client/stubs.rb
```diff
@@ -0,0 +1,213 @@
...
+      # * **`:status_code`** - <Integer> - The HTTP status code
+      # * **`:headers`** - Hash<String, String> - A hash of HTTP header keys and values
+      # * **`:body`** - <String, IO> - The HTTP response body.
+      #
+      # ## Stubbing Multiple Responses
+      #
+      # Calling an operation multiple times will return similar responses.
+      # You can configure multiple stubs, and they will be returned in sequence.
+      #
+      #     client.stub_responses(:get_city, [
```
jterapin, line 90: I'm not sure why the code snippet here is not highlighted whereas others are... weird! 

Screenshot 2025-03-04 at 9 38 46 AM

gems/smithy-client/lib/smithy-client/stubs.rb
```diff
@@ -0,0 +1,213 @@
...
+        stub = convert_stub(operation_name, stub, context)
+        stub[:mutex] = Mutex.new
+        stub
+      end
+
+      private
+
+      def apply_stubs(operation_name, stubs)
+        @config.stubs_mutex.synchronize do
+          @config.stubs[operation_name.to_sym] = stubs
```
jterapin: I don't think we need to `to_sym` here since the expectation here is that we are already provided symbolized operation name based on the documentation? Jw

gems/smithy-client/spec/smithy-client/codecs/cbor_spec.rb
```diff
@@ -5,71 +5,103 @@
...
+            expect(subject.deserialize(shape, bytes).to_h).to eq(data)
+          end
         end
 
-        it 'deserializes returns an empty hash when given bytes are empty' do
-          expect(subject.deserialize(string_shape, '')).to be_empty
+        context 'unions' do
+          it 'serializes and deserializes union as a type' do
+            shape = service.operation(:operation).input
+            union = shape.member(:union).shape.member_type(:string).new('string')
```
jterapin, line 75: Could the union member name be more "complex"? i.e. `fooBar`,`:foo_bar`  vs `string`, `:string`
 
I want to make sure this works correctly.

same goes for SERDE test for hash case

gems/smithy-client/lib/smithy-client/protocols/rpc_v2.rb
```diff
@@ -36,26 +36,24 @@ def error(context)
...
+          resp.body = @codec.serialize(operation.output, data)
+          resp
+        end
+
+        def stub_error(error_code)
+          resp = HTTP::Response.new
+          resp.status_code = 400
+          resp.headers['Smithy-Protocol'] = 'rpc-v2-cbor'
+          resp.headers['Content-Type'] = 'application/cbor'
+          data = { '__type' => error_code, 'message' => 'stubbed-error-message' }
```
mullermp, line 53: Technically error code should be full shape id + error code, but in stub, it can just be error code, and the split('#').last logic still works. We could make this as real as possible but it requires parsing out the namespace. We may not even know the namespace of the error code anyway because it could be different than the service's.

jterapin: Yeah agreed, namespace could be anything in the real world

gems/smithy/lib/smithy/templates/client/client.erb
```diff
@@ -12,6 +12,8 @@ module <%= module_name %>
   class Client < Smithy::Client::Base
     self.service = Schema::SERVICE
 
+    include Smithy::Client::Stubs
```
jterapin, line 13: is it better practice to mix in modules first and then do `self.service...`?

gems/smithy/spec/interfaces/client/stub_responses_spec.rb
```diff
@@ -0,0 +1,147 @@
...
+        end
+
+        it 'can set stubbed data mixed with defaults' do
+          data = default_stub_data.merge(string: 'new string')
+          stub = subject.stub_data(:operation, { string: 'new string' })
+          expect(stub.to_h).to include(data)
+        end
+      end
+
+      describe '#stub_responses' do
```
jterapin, line 59: Can we add a test where we stub multiple responses?

gems/smithy-client/lib/smithy-client/plugins/stub_responses.rb
```diff
@@ -0,0 +1,93 @@
...
+            to return or errors to raise by calling {Stubs#stub_responses}.
+            @see Stubs
+          DOCS
+
+        option(:stubs) { {} }
+        option(:stubs_mutex) { Mutex.new }
+        option(:api_requests) { [] }
+        option(:api_requests_mutex) { Mutex.new }
+
+        def add_handlers(handlers, config)
```
jterapin, line 24: This is just for curiosity- do you prefer to have `add_handlers` on the top or bottom? Or does it really depend on context?

gems/smithy-client/lib/smithy-client/stubbing/data_applicator.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    module Stubbing
```
jterapin, line 5: Should we have a file, `stubbing.rb?` so we can mark that module api-private? Showing up on rendered docs.

gems/smithy/lib/smithy/templates/client/protocol_plugin.erb
mullermp: Error handling happens before parsing of data. If a stubbed error is set, then output.error will already have an object, and we don't want to parse the error again.

gems/smithy-client/lib/smithy-client/stubs.rb
mullermp: I'm not sure why either.. I can't get it to work.

gems/smithy/lib/smithy/templates/client/client.erb
mullermp: It should not matter because the module does not set service, but I've flipped them anyway.

gems/smithy-client/lib/smithy-client/plugins/stub_responses.rb
mullermp: I just follow the order that was defined in the Plugin class.

gems/smithy-client/lib/smithy-client/stubbing/data_applicator.rb
mullermp: I think we want to use the struct member names (the values that the customer would interact with). The customer should never know about the member names (location name) and that's an internal detail for us when parsing responses. I think the only case where member names are maybe useful is raw http response stubbing but I think we'd have a utility or a way to convert to that body (like with using the codecs).

gems/smithy/lib/smithy/templates/client/protocol_plugin.erb
jterapin: Gotcha, thanks for the clarity!

gems/smithy-client/lib/smithy-client/stubbing/data_applicator.rb
jterapin: Yeah that makes sense - that those are just internal details for us.

jterapin (approved):
🚢✨

PR #279, opened 2025-02-27
Refactor rake tasks and allow per file rspecs

mullermp:
Refactors rake tasks into their own files. Adds spec_helper requires to all spec files so that you can run rspec on each file individually.


tasks/smithy.rake
```diff
@@ -0,0 +1,118 @@
...
+
+  task 'spec:endpoint_tests' do
+    generated_spec_task('endpoint_tests')
+  end
+
+  task 'spec:protocol_tests' do
+    generated_spec_task('protocol_tests')
+  end
+
+  task 'spec' => %w[spec:unit spec:endpoint_tests spec:protocol_tests]
```
jterapin: Could we move this to line 14? Just wanted to make the grouped-up rake task similar to other files. Same goes for rbs bulk task here

jterapin (approved):
Looks good ✨

PR #281, opened 2025-03-03
Misc improvements

mullermp:
Misc clean ups to remove out of the stubs PR. Adds the following:

* Fix readme with smithy-schema
* Updates projections
* Union serde
* Param validation of union
* Codec error support
* RPCv2 error test cases
* Schema refactor to use a structure class for structure shapes

### DIFF
--- a/gems/smithy-client/lib/smithy-client/param_validator.rb
+++ b/gems/smithy-client/lib/smithy-client/param_validator.rb
@@ -15,11 +15,12 @@ module Smithy
       end
 
       # @param [Hash] params
+      # @param [String] context
       # @return [void]
       # @raise [ArgumentError] if the params are invalid
-      def validate!(params)
+      def validate!(params, context: 'params')
         errors = []
-        structure(@rules, params, errors, 'params')
+        structure(@rules, params, errors, context)
         raise ArgumentError, error_messages(errors) unless errors.empty?
       end
 
@@ -32,12 +33,7 @@ module Smithy
         values.each_pair do |name, value|
           next if value.nil?
 
-          if shape.member?(name)
-            member_shape = shape.member(name)
-            shape(member_shape.shape, value, errors, context + "[#{name.inspect}]")
-          else
-            errors << "unexpected value at #{context}[#{name.inspect}]"
-          end
+          member(shape, name, value, errors, context)
         end
       end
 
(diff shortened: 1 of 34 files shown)


gems/smithy-client/lib/smithy-client/param_validator.rb
```diff
@@ -53,33 +49,40 @@ def valid_structure?(shape, values, errors, context)
...
+        elsif values.respond_to?(:each_pair)
+          values.each_pair do |name, value|
+            next if value.nil?
 
-          if shape.member?(name)
-            member_shape = shape.member(name)
-            shape(member_shape.shape, value, errors, context + "[#{name.inspect}]")
-          else
-            errors << "unexpected value at #{context}[#{name.inspect}]"
+            member(shape, name, value, errors, context)
```
jterapin, line 59: Question - when the input is a hash (rather than union subclass), where do we do the conversion from input hash -> union subclass? Without this - will run into an error within `format_union` method since the expectation is for `values.class` will not be an union subclass.  

Does that the conversion responsibility fall somewhere else?

gems/smithy-client/lib/smithy-client/protocols/rpc_v2.rb
```diff
@@ -11,38 +11,85 @@ class RPCv2
         # @api private
         SHAPE_ID = 'smithy.protocols#rpcv2Cbor'
 
-        # @param options [Hash] Protocol options
-        # @option options [Boolean] :query_compatible (nil)
+        # @param options [Hash]
         def initialize(options = {})
-          @query_compatible = options[:query_compatible]
+          @codec = Codecs::CBOR.new(options)
```
jterapin, line 16: During the codec implementation, I wasn't completely set on whether codec should be a fresh instance every time or not but I think this is good for now. 

I want to say there's a reason why I always used a fresh instance of codec but I need to deep dive during typed document implementation....

gems/smithy-client/lib/smithy-client/protocols/rpc_v2.rb
```diff
@@ -11,38 +11,85 @@ class RPCv2
...
+
+          code = data.delete('__type').split('#').last
+          message = data['message']
+          data = parse_error_data(context, body, code)
+          [code, message, data]
+        end
+
+        def parse_error_data(context, body, code)
+          data = Schema::EmptyStructure.new
+          if (error_rules = context.operation.errors)
```
jterapin: These are actually error shapes, right?

gems/smithy-client/lib/smithy-client/protocols/rpc_v2.rb
```diff
@@ -53,7 +100,7 @@ def apply_accept_header(context)
           context.request.headers['Accept'] = 'application/cbor'
         end
 
-        def apple_content_type(context)
```
jterapin, line 56: Haha aww my bad 🍎

gems/smithy/lib/smithy/templates/client/schema.erb
```diff
@@ -15,11 +15,14 @@ module <%= module_name %>
 <% shape.members.each do |member| -%>
     <%= shape.name %>.<%= member.add_member_method(shape.type) %>
 <% end -%>
+<% if shape.type == 'UnionShape' -%>
+    <%= shape.name %>.add_member(:unknown, 'unknown', Prelude::Unit, Types::<%= shape.name %>::Unknown)
```
jterapin: I forgot to ask - I wasn't sure what the "shape" representation of Unknown member should be since I need to put something in there but went with `Prelude::Unit` since it is used to represented as `nil` in smithy models but open to alternatives

gems/smithy/spec/fixtures/protocol_tests/rpcv2_cbor/skipped_tests.smithy
```diff
@@ -1,15 +1,10 @@
...
 namespace smithy.protocoltests.rpcv2Cbor
+
 use smithy.ruby#skipTests
 
 apply OperationWithDefaults @skipTests([
   { id: "RpcV2CborClientPopulatesDefaultValuesInInput", reason: "Defaults not Implemented yet.", type: "request" }
   { id: "RpcV2CborClientPopulatesDefaultsValuesWhenMissingInResponse", reason: "Defaults not Implemented yet.", type: "response" }
 ])
-
-apply GreetingWithErrors @skipTests([
```
jterapin, line 11: Yay it passes!

jterapin (approved):

gems/smithy/lib/smithy/templates/client/schema.erb
mullermp: Good question. I think it may actually be a union shape?

mullermp: We should have tests for this later. It's ok for this to be "semi functional"

gems/smithy-client/lib/smithy-client/protocols/rpc_v2.rb
mullermp: Yes. I'll rename.

gems/smithy-client/lib/smithy-client/protocols/rpc_v2.rb
mullermp: I don't know, but it seems fine as an instance here. We can refactor later if we have to.

gems/smithy-client/lib/smithy-client/param_validator.rb
mullermp: I think it will need to handle both cases, unless we change union to be a struct-like thing.

jterapin: Yeah we plan to revisit anyways

PR #282, opened 2025-03-06
Implement Typed Documents and TypeRegistry

jterapin:
Description: Implementation for Typed Documents and TypeRegistry. Currently only supports JSON documents. 

It is highly likely that we have to revisit this implementation in the future since typed document/type registry still being evolved.


gems/smithy-schema/lib/smithy-schema/document.rb
```diff
@@ -0,0 +1,159 @@
...
+
+require 'json'
+
+module Smithy
+  module Schema
+    # TODO: need to address the following and more
+    #  * documentation
+    #  * some json considerations like timestamp, jsonName trait etc
+    #  * handle union stuffs
+    class Document
```
jterapin: There are about 3~ possible given data types:
* Case 1: some Ruby object (e.g, `{foo:"bar"}`)
  * We will set the data as it is given.
  * If we are given a schema, we will attempt to return a runtime shape.
* Case 2: an instance of runtime shape (e.g, `BirdServices::Types::Bird.new(...)`)
  *  Expectation is that we are given a schema here as well (we need this information to properly serialize as a document data). 
  * **Open Question:** Should we have a schema attribute on Document class? 
 * Case 3: Parsed json (e.g, a json response that went through `JSON.parse(...)`
   * Expectation here is that we get a parsed json object. We will set the data as given but extract and set the discriminator attribute.

gems/smithy-schema/lib/smithy-schema/document.rb
```diff
@@ -0,0 +1,104 @@
...
+
+        case data
+        when Smithy::Schema::Structure
+          schema = opts[:schema]
+          if schema.nil? || !valid_schema?(schema)
+            raise ArgumentError, "Unable to create a document with given schema: #{schema}"
+          end
+
+          opts = opts.except(:schema)
+          # case 1 - extract data from runtime shape, schema is required to know to properly extract
```
jterapin: Note: I'll remove it before merging.

gems/smithy-schema/lib/smithy-schema/type_registry.rb
```diff
@@ -0,0 +1,94 @@
...
+      # @param [Class] type
+      # @return [Shapes::Shape, nil]
+      def schema_by_type(type)
+        @schema_by_types[type]
+      end
+
+      # Deserializes a document into a typed shape from registry.
+      # @param [Document] document
+      # @return [Shapes::Structure] typed shape
+      def convert_as_typed(document)
```
jterapin: We _could_ start off by not including this functionality if we still feel that it's in a weird place. 
Just few extra lines when working with typed documents:
```ruby
# assume that this document has a discriminator 
document = Document.new(....)

# find schema via TypeRegistry
schema = BirdService::Schema:::TypeRegistry.schema_by_id(document.discriminator)

# convert from document -> typed shape
document.as_typed(schema)
=> BirdServices::Types::Bird.....
```

I'm cool with either decision, since we meet the required functionalities.

gems/smithy-schema/spec/smithy-schema/document_spec.rb
```diff
@@ -0,0 +1,241 @@
+# frozen_string_literal: true
+
+require_relative '../spec_helper'
+
+module Smithy
+  module Schema
+    describe Document do
+      let(:simple_runtime) do
```
jterapin: I think I could really use the SchemaHelper to help the test set up look cleaner here but this means that `smithy-schema` will need to declare dependency on `smithy` gem (see [here](https://github.com/smithy-lang/smithy-ruby/blob/decaf/gems/smithy-schema/spec/support/schema_helper.rb#L128)). Not sure if we want that.

gems/smithy-schema/lib/smithy-schema/document.rb
```diff
@@ -0,0 +1,104 @@
...
+require_relative 'document_utils'
+
+module Smithy
+  module Schema
+    # A Smithy document type, representing typed or untyped data from Smithy data model.
+    # ## Document types
+    # Document types are protocol-agnostic view of untyped data. They could be combined
+    # with a schema to serialize its contents.
+    #
+    # Smithy-Ruby currently only support JSON documents.
```
jterapin: Not sure if we could use the same class to support XML documents in the future. We could potentially create some extension points from here and change DocumentUtils into XML/JSON specific document builders/parsers. Future problem if there's a use case.

gems/smithy-schema/lib/smithy-schema/document.rb
```diff
@@ -0,0 +1,104 @@
+# frozen_string_literal: true
+
+require_relative 'document_utils'
+
+module Smithy
+  module Schema
+    # A Smithy document type, representing typed or untyped data from Smithy data model.
+    # ## Document types
```
jterapin: I could add more details (e.g. code examples) but I might punt that to the Smithy-Ruby Wiki instead.

gems/smithy-schema/lib/smithy-schema/document.rb
```diff
@@ -0,0 +1,104 @@
...
+      # @param  [Object] data  document data
+      # @param [Hash] options
+      # @option options [Smithy::Schema::Structure] :schema schema to reference when setting
+      #  document data. Only applicable when data param is a type of {Shapes::StructureShape}.
+      # @option options [Boolean] :use_timestamp_format Whether to use the `timestampFormat`
+      #  trait or ignore it when creating a {Document} with given schema. The `timestampFormat`
+      #  trait is ignored by default.
+      # @option options [Boolean] :use_json_name Whether to use the `jsonName` trait or ignore
+      #  it when creating a {Document} with given schema. The `jsonName` trait is ignored
+      #  by default.
```
jterapin: I was pretty torn on whether I want to have these options around. These options only applies when we create a document with a given schema to decide whether we want to honor specific traits in the `@data`

My idea is that once `@data` is set - we can simply do `JSON.dump(document.data)` to load the data for a request - in lieu of having to iterate over the document data to adhere to traits during SERDE process.

gems/smithy-schema/lib/smithy-schema/document_utils.rb
```diff
@@ -0,0 +1,231 @@
+# frozen_string_literal: true
+
+require 'base64'
+require 'time'
+
+module Smithy
+  module Schema
+    # @api private
+    # Document Utilities to help (de)construct data to/from Smithy document
+    module DocumentUtils
```
jterapin: I could separate them into different modules/classes like Builder, Parsers (maybe call it something else) down the road if this file is bloated.

gems/smithy-schema/spec/smithy-schema/document_spec.rb
jterapin: Also - I want to expand on these tests to cover all cases but holding off until I get a clear direction of what we want to do.

projections/shapes/lib/shapes/schema.rb
```diff
@@ -57,28 +53,29 @@ module Schema
...
         operation.name = "Operation"
         operation.input = OperationInputOutput
         operation.output = OperationInputOutput
-        operation.traits = {"smithy.ruby.tests#shape" => {}}
+        operation.traits = {"smithy.ruby.tests#shape"=>{}}
       end)
     end
+
+    TYPE_REGISTRY = Smithy::Schema::TypeRegistry.new
+    TYPE_REGISTRY.register(OperationInputOutput, Structure, Union)
```
jterapin: I THINK we only care about shapes that has a type representation so I only include those specific ones.

projections/weather/lib/weather/schema.rb
```diff
@@ -93,5 +83,8 @@ module Schema
         operation[:paginator] = Paginators::ListCities.new
       end)
     end
+
+    TYPE_REGISTRY = Smithy::Schema::TypeRegistry.new
+    TYPE_REGISTRY.register(CityCoordinates, CitySummary, GetCityInput, GetCityOutput, GetCurrentTimeOutput, GetForecastInput, GetForecastOutput, ListCitiesInput, ListCitiesOutput, NoSuchResource)
```
mullermp: I think a separate register method doesn't make too much sense. I'd just pass these things into new().

gems/smithy-schema/lib/smithy-schema/type_registry.rb
```diff
@@ -0,0 +1,94 @@
...
+      end
+
+      # @api private
+      attr_reader :schema_by_types
+
+      # @return [Hash<String, Shapes::Shape>]
+      attr_accessor :registry
+
+      # @param [Array<Shapes::Shape>] shapes
+      def register(*shapes)
```
mullermp: This really seems like a glorified map - don't we know these at code generation time? Rather than create this at runtime, we can pass a static map to initialize of id => constant. We have a lot of big APIs.

gems/smithy-schema/lib/smithy-schema/type_registry.rb
```diff
@@ -0,0 +1,94 @@
...
+          when Shapes::UnionShape
+            s.member_types.values { |v| @schema_by_types[v] = s }
+          end
+        end
+      end
+
+      # Returns true if this type registry contains specific shape id.
+      # @param [String] id
+      # @return [Boolean]
+      def schema_by_id?(id)
```
mullermp: We should consider making this a hashy interface. Perhaps including Enumerable, implementing `key?`, `[]=` and `[]`

gems/smithy-schema/lib/smithy-schema/type_registry.rb
```diff
@@ -0,0 +1,94 @@
...
+      # @return [Shapes::Shape, nil]
+      def schema_by_id(id)
+        @registry[id]
+      end
+
+      # Returns true if this type registry contains a schema associated
+      #  with the given typed shape.
+      # @param [Class] type
+      # @return [Boolean]
+      def schema_by_type?(type)
```
mullermp: Where is schema_by_type used or useful? Isn't this also a shape not a schema?

gems/smithy-client/lib/smithy-client/base.rb
```diff
@@ -161,6 +161,16 @@ def service=(service)
           define_operation_methods
         end
 
+        # @return [Schema::TypeRegistry]
+        def type_registry
```
mullermp: What's the intention of placing this on base and why is it needed here? This will add a method to clients. I think we should bias to make this feature for schema gems only for now.

gems/smithy-schema/lib/smithy-schema/document.rb
mullermp: I think in-code examples/documentation are preferred but it doesn't have to be done this moment.

gems/smithy-schema/lib/smithy-schema/document.rb
mullermp: If you are referring to the use_x options, I think they're pretty useless, and can always be added later. I'd start with them removed.

gems/smithy-schema/lib/smithy-schema/document.rb
mullermp: Documents should be JSON - so no symbolized keys, e.g. `{"foo" => "bar"}`. Case 1 will need to understand this.

I think in general we should abstract away schema/shapes/typing as much as possible.

This is very much a "typed language" feature. Ruby is not that kind of language, so I think we just want maybe a simple way to take json data and make a few structs from it kind of like we do today when we return client output.

gems/smithy-schema/lib/smithy-schema/document.rb
```diff
@@ -0,0 +1,104 @@
...
+
+      # @return [Object] data
+      attr_reader :data
+
+      # @return [String] discriminator
+      attr_reader :discriminator
+
+      # @param [Object] key
+      # @return [Object]
+      def [](key)
```
mullermp: Why? - I think this class should delegate to the data (perhaps even use Delegator). It should just add some utility to the raw document value.

gems/smithy-schema/lib/smithy-schema/type_registry.rb
mullermp: I would start with the minimum and add more later if we need it. I'm still skeptical of this feature being necessary.

gems/smithy-schema/lib/smithy-schema/type_registry.rb
```diff
@@ -0,0 +1,94 @@
...
+          msg = "Unable to find schema with #{document.discriminator} in Registry"
+          raise ArgumentError, msg
+        end
+      end
+
+      class << self
+        # Composes multiple type registries together.
+        # @param [Array<TypeRegistry>]
+        # @return [TypeRegistry]
+        def compose(*type_registries)
```
mullermp: I think implementing `concat` is idiomatic, and it would be an instance method like `registry.concat(other)`

gems/smithy-schema/spec/smithy-schema/document_spec.rb
mullermp: SchemaHelper should exist in smithy-schema already. In spec files we don't need dependencies, you can do relative pathing.

gems/smithy/spec/support/examples/schema_examples.rb
```diff
@@ -390,4 +390,23 @@ def expect_generated_shape(subject, shape_class, shape_hash)
...
     end
   end
+
+  context 'type registry' do
+    subject { ShapeService::Schema::TYPE_REGISTRY }
+
+    let(:typed_shapes) do
+      fixture['shapes'].select do |_k, v|
+        %w[union structure].include?(v['type']) &&
+          !v['traits']&.include?('smithy.api#trait')
```
mullermp: It would have any union/structure wouldn't it?

gems/smithy-schema/lib/smithy-schema/document_utils.rb
mullermp: Probably, it does two different things. It can also be a class that initializes with the schema and build/parse when given data and options.

gems/smithy-schema/lib/smithy-schema/document.rb
```diff
@@ -0,0 +1,104 @@
...
+        return unless @data.is_a?(Hash) && @data.key?(key)
+
+        @data[key]
+      end
+
+      # @param [Shapes::Shape] schema
```
mullermp: We've been referring to schema as the entire API (the file is called schema, defines a service shape, and other shapes). Should that be true for sub-shapes inside the schema? This is probably just a shape.

gems/smithy-schema/lib/smithy-schema/document.rb
```diff
@@ -0,0 +1,104 @@
...
+
+          schema.id
+        end
+      end
+
+      def set_data(data, opts)
+        return if data.nil?
+
+        case data
+        when Smithy::Schema::Structure
```
mullermp: Isn't this already "done"? If it's a structure then it's already typed.

mullermp (commented):
I put some comments to start with, didn't dig into any testing code.

gems/smithy-client/lib/smithy-client/base.rb
alextwoods: I think all gems should have the type register available - though I'm not sure it belongs as a method on client class - maybe as a method on `Schema` module, eg: `Schema.type_registery`

gems/smithy-schema/lib/smithy-schema/document.rb
alextwoods: Yea - I'd agree examples and more detail would be better as in code documentation.

gems/smithy-schema/lib/smithy-schema/document.rb
alextwoods: To me case 1 and case 3 generally seem the same? (assuming that  case 1 is limited to ruby objects of supported classes only like hash/list/number/boolean/string).  

The issue of symbol vs string keys is a tricky one - parsed json should generally have string keys, but arbitrary Ruby hashes we get are likely to have symbol keys.  

Maybe its just a concern of as_typed to be able to handle both? (ie, when trying to convert a document to a specific runtime shape, we would need to consider both string and symbol keys?)

gems/smithy-schema/lib/smithy-schema/document.rb
```diff
@@ -0,0 +1,104 @@
...
+          schema = opts[:schema]
+          if schema.nil? || !valid_schema?(schema)
```
alextwoods: It does still seem weird to me that the run time type doesn't know its own schema

gems/smithy-schema/lib/smithy-schema/document_utils.rb
```diff
@@ -0,0 +1,231 @@
...
+        # Use the timestampFormat trait of the member, if present.
+        # Use the timestampFormat trait of the shape, if present.
+        # If none of the above applies, use epoch-seconds as default
+        def timestamp_format(schema)
+          if schema.traits['smithy.api#timestampFormat']
+            schema.traits['smithy.api#timestampFormat']
+          elsif schema.shape.traits['smithy.api#timestampFormat']
+            schema.shape.traits['smithy.api#timestampFormat']
+          else
+            'epoch-seconds'
```
alextwoods: Should the default be configurable in some way?

gems/smithy-schema/lib/smithy-schema/document_utils.rb
```diff
@@ -0,0 +1,231 @@
...
+          end
+        end
+
+        def time(data, trait = nil)
```
alextwoods: Probably a generic util and not document specific?

gems/smithy-schema/lib/smithy-schema/document_utils.rb
```diff
@@ -0,0 +1,231 @@
...
+          end
+        end
+
+        def apply_union_member(key, value, shape, type)
+          member_name = shape.name_by_member_name(key)
+          type = shape.member_type(member_name) if type.nil?
+          type.new(apply(value, shape.member(member_name)))
+        end
+
+        def extract_blob(data)
```
alextwoods: I know we're only supporting json right now, but blobs would be one of the things that would differ with other formats - eg: in a cbor document, we wouldn't base64 encode these.... I'm not sure how best to represent that behavior though and maybe its fine to leave in generic DocumentUtils for now since its api private and we could refactor it into JsonDocumentUtils, CborDocumentUtils, ect later if needed?

gems/smithy-schema/lib/smithy-schema/type_registry.rb
```diff
@@ -0,0 +1,94 @@
...
+    # A registry that contains a map of Smithy shape ID to its schema.
+    # Also includes a way to find schema based on its shape type representation
+    class TypeRegistry
+      def initialize
+        @registry = {}
+        @schema_by_types = {}
+      end
+
+      # @api private
+      attr_reader :schema_by_types
```
alextwoods: Is there a reason we need an attr_reader for this that is api private?

gems/smithy-schema/lib/smithy-schema/type_registry.rb
```diff
@@ -0,0 +1,94 @@
...
+
+      # @return [Hash<String, Shapes::Shape>]
+      attr_accessor :registry
```
alextwoods: Should this be private?

gems/smithy-schema/lib/smithy-schema/type_registry.rb
alextwoods: Yeah - I think I'd lean towards not including this to start as part of the type registry since its relatively easy to do with Document.as_typed + a schema lookup.

gems/smithy-schema/lib/smithy-schema/type_registry.rb
```diff
@@ -0,0 +1,94 @@
...
+
+      class << self
+        # Composes multiple type registries together.
+        # @param [Array<TypeRegistry>]
+        # @return [TypeRegistry]
+        def compose(*type_registries)
+          raise ArgumentError, 'Expected an array of TypeRegistries' unless type_registries.all?(self)
+
+          new_type_registry = new
+          new_type_registry.registry =
```
alextwoods: Does this add to the @schema_by_type?  I think would be potentially better to use the public APIs (`register` or`[]=`) - but I guess that requires implementing say enumerable so you could enumerate everything from each and add to the new.

alextwoods (commented):
Nice - looking good overall!

gems/smithy-client/lib/smithy-client/base.rb
jterapin: We need access to TypeRegistry during SERDE process - when it comes to deserializing typed documents. I haven't quite figured out where would be a good "home" for this.

I'm thinking of few options...
1 - Update `Smithy::Client::HandlerContext` to have access to the type registry (when deserializing typed documents - we can find the associated shape) - like: `context.type_registry`
2 - Keep the reference on the client so we don't need to add another attribute on the `HandlerContext` - we could do: `context.client.type_registry` to access during SERDE process
3 - Should `ServiceShape` have an attribute called `type_registry` so we could do: `client.service.type_registry` for easily access?
4 - Just omit the reference for now until we need it. If you need it for whatever reason, you gotta use this constant: `BirdService::Schema.TYPE_REGISTRY` 

_If we decide to omit this now, when do we need it?_ 
Any JSON-esque protocol since they support `JSON` docs

What do y'all think?

gems/smithy-schema/lib/smithy-schema/document.rb
jterapin: I will add them now - since information is very fresh in my head.

gems/smithy-schema/lib/smithy-schema/document_utils.rb
jterapin: Yeah that makes sense. I imagine once we cross to that point and the specification for XML Documents comes out - we have flexibility to change up things here.

gems/smithy-schema/lib/smithy-schema/document.rb
jterapin: Yeah the symbol vs string keys is the part I struggled with for each specific case. I could just pull away from symbolized keys much as I'm able too in the context of storing the document data. 

Currently, `as_typed` checks for both representation of member keys - I think it would be best to leave that functionality in there to ensure that the document is deserialized correctly....

gems/smithy-schema/lib/smithy-schema/document.rb
jterapin: I noticed that there's test cases now, I might see what they ended up doing to determine whether i want to keep these options around.

gems/smithy-schema/lib/smithy-schema/document.rb
jterapin: I have a question about this comment but we could discuss it offline.

gems/smithy-schema/lib/smithy-schema/document.rb
jterapin: Yeah shapes will be more appropriate, I'll update.

gems/smithy-schema/lib/smithy-schema/document.rb
jterapin: Can you expand on that? what do you mean by being "done" - I still want to re-format the typed shape into a document format.

gems/smithy-schema/lib/smithy-schema/document.rb
jterapin: I think it would be useful to have that information at hand but we had some previous discussion on whether this should be a thing or not - based on context of typed shapes being PORO . I don't have a strong opinion but I think starting out without is fine until we have a use case where we absolutely need them to.

gems/smithy-schema/lib/smithy-schema/document_utils.rb
jterapin: We could consider it - I was thinking w/o it for now and let schema traits drive the format.

gems/smithy-schema/lib/smithy-schema/document_utils.rb
jterapin: Ah, yeah. I could create an utility like TimeHelper that we had in the previous iteration of Smithy-Ruby (but in this context, it will live inside smithy-schema gem)

gems/smithy-schema/lib/smithy-schema/type_registry.rb
jterapin: I was thinking that we should hide away the actual data storage here but let customers use `schema_by_id` or `schema_by_type` to access type registry details.

gems/smithy-schema/lib/smithy-schema/type_registry.rb
jterapin: Yeah I could make it private to be aligned with the `:schema_by_types`

gems/smithy-schema/lib/smithy-schema/type_registry.rb
jterapin: Yeah I could pass it at initialization if given.  I'm thinking Dowling still want the ability to "register" schemas that is not bounded to service model.

gems/smithy-schema/lib/smithy-schema/type_registry.rb
jterapin: Yeah we could do that for `:schema_by_id` stuff - but still need to keep around `:schema_by_type` methods

gems/smithy-schema/lib/smithy-schema/type_registry.rb
jterapin: Yeah I'll update the method names to be aligned with the shapes references.

These method will be used in case #2 mentioned in this comment: https://github.com/smithy-lang/smithy-ruby/pull/282/files#r2027627236

When a customer decides to pass a runtime shape for the document param - we need a way to find its shape-id so we can load the discriminator for the service to find it in their type registry

This is a proposed solution instead of having runtime shapes know its own schema.

gems/smithy-schema/lib/smithy-schema/type_registry.rb
jterapin: Cool, I'll remove it for now.

gems/smithy-schema/lib/smithy-schema/type_registry.rb
jterapin: Gotcha, I will use that.

gems/smithy-schema/lib/smithy-schema/type_registry.rb
jterapin: Good callout, I have forgotten to resolve the latter mapping.  I'll take a look to see what options make sense here.

gems/smithy-schema/spec/smithy-schema/document_spec.rb
jterapin: Sounds good, I will make use of relative pathing - that should make the tests cleaner (easier to update over time and etc).

jterapin (commented):
Responding back to comments first before I start revising. Thanks!

gems/smithy-schema/lib/smithy-schema/document.rb
jterapin: I think we talked this over offline.

gems/smithy-schema/lib/smithy-schema/document.rb
```diff
@@ -0,0 +1,51 @@
...
+    #
+    # @example Basic usage with a document
+    #   data = Document::Data.new({ name: "document" })
+    #   data.data  # => { "name" => "example" }
+    #
+    # @example Using with a shape
+    #   shape = Smithy::Schema::StructureShape.new
+    #   data = Document::Data.new({ "name" => "example" }, shape: shape)
+    #
+    module Document
```
mullermp: I was expecting document to be a class and have it be the delegator. What was the intention of making another data subclass?

gems/smithy-schema/lib/smithy-schema/document.rb
```diff
@@ -0,0 +1,51 @@
...
+          super(@data)
+          @discriminator = options[:discriminator] || nil
+        end
+
+        # Returns the discriminator value for the document
+        #
+        # @return [String. nil] discriminator
+        attr_reader :discriminator
+
+        def data
```
mullermp: We possibly don't need this method. It should already delegate to the value. I think in union it's not used either.

gems/smithy-schema/lib/smithy-schema/document.rb
```diff
@@ -0,0 +1,51 @@
...
+      # document-specific functionality.
+      class Data < ::SimpleDelegator
+        # @param [Object] data  document data that is in JSON-friendly format
+        # @param [Hash] options
+        # @option options [String] :discriminator This value is used to identify a specific
+        # shape. This is equivalent of a Smithy shape ID.
+        def initialize(data, options = {})
+          @data = data
+          super(@data)
+          @discriminator = options[:discriminator] || nil
```
mullermp: It already defaults to nil if not provided.

gems/smithy-schema/lib/smithy-schema/type_registry.rb
```diff
@@ -0,0 +1,120 @@
...
+    # @example Combining multiple registries
+    #  TypeRegistry.concat(registry1, registry2)
+    #  # => #<Smithy::Schema::TypeRegistry...>
+    class TypeRegistry
+      include Enumerable
+
+      # @param  [Hash<String, Shapes::StructureShape>] registry
+      def initialize(registry = {})
+        @registry = registry
+        @shapes_by_type = register_shape_types(registry.values)
```
mullermp: Is there a way to populate this from the code generated side? If we must iterate shapes, we may as well backtrack and populate both maps in one pass. That at least reduces generated code. Otherwise is shapes_by_type even necessary?

gems/smithy-schema/lib/smithy-schema/type_registry.rb
```diff
@@ -0,0 +1,120 @@
...
+      end
+
+      # @api private
+      # @return [Hash<String, Shapes::StructureShape>]
+      attr_accessor :registry
```
mullermp: Both of these accessors shouldn't exist. Our public methods should hide this detail.

gems/smithy-schema/lib/smithy-schema/type_registry.rb
```diff
@@ -0,0 +1,120 @@
...
+      # @return [Shapes::StructureShape, nil]
+      def [](id)
+        @registry[id]
+      end
+
+      # @param [String] id
+      # @param [Shapes::StructureShape] shape
+      def []=(id, shape)
+        msg = 'Expected a StructureShape that has a type representation'
+        raise ArgumentError, msg unless shape.is_a?(Shapes::StructureShape) && shape.type
```
mullermp: All structure shapes should have a type now (at least in my PR)

gems/smithy-schema/lib/smithy-schema/type_registry.rb
```diff
@@ -0,0 +1,120 @@
...
+
+        @registry[id] = shape
+        @shapes_by_type[shape.type] = shape
+      end
+
+      # Returns true if the registry contains specific shape id.
```
mullermp: Nit: non-return or param documentation for simple Boolean methods and getters/setters is too verbose and is already obvious.

gems/smithy-schema/lib/smithy-schema/type_registry.rb
```diff
@@ -0,0 +1,120 @@
...
+
+          h[s.type] = s
+        end
+      end
+
+      class << self
+        # Combines multiple registries and returns a new registry.
+        # @param [Array<TypeRegistry>]
+        # @return [TypeRegistry]
+        def concat(*type_registries)
```
mullermp: I've mentioned this already a few times but this is not idiomatic . Concat is an instance method not a class method. https://ruby-doc.org/core-2.6.7/Array.html#method-i-concat

I also don't think we need to validate that they are registries (any enumerable will work) and I don't think we need to access the registry method either because we implemented each() which I think would work.

gems/smithy-schema/spec/support/schema_helper.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require_relative '../../../smithy/lib/smithy'
```
mullermp: This should not be necessary. The spec helper should already pull this in. What issue were you seeing?

projections/shapes/lib/shapes/schema.rb
```diff
@@ -75,5 +75,7 @@ module Schema
         operation.traits = {"smithy.ruby.tests#shape" => {}}
       end)
     end
+
+    TYPE_REGISTRY = Smithy::Schema::TypeRegistry.new('smithy.ruby.tests#OperationInputOutput' => OperationInputOutput, 'smithy.ruby.tests#Structure' => Structure)
```
mullermp: We should consider a class method on the schema instead of a constant. There could be a shape that conflicts with this.

gems/smithy-schema/lib/smithy-schema/document/serializer.rb
```diff
@@ -0,0 +1,283 @@
...
+              h[k.to_s] = serialize_untyped(v)
+            end
+          when Array
+            values.map { |d| serialize_untyped(d) }
+          else
+            values
+          end
+        end
+
+        def shape(ref, values, opts = {}) # rubocop:disable Metrics/CyclomaticComplexity
```
mullermp: Rather than passing options around in method calls, this is a class, so it seems better to set instance variables for the options. You can see my latest PR where I add one to a codec.

gems/smithy-schema/lib/smithy-schema/document/serializer.rb
```diff
@@ -0,0 +1,283 @@
...
+        #  JSON response with type discriminator.
+        # @return [Data] document data
+        #
+        # @example Ruby objects as input
+        #   # create serializer with an existing type registry
+        #   serializer = Smithy::Schema::Document::Serializer(type_registry)
+        #
+        #   # ruby objects as input
+        #   serializer.create_document("some document")
+        #   # => {"foo" => "bar"}
```
mullermp: This example is not really helpful. How did you get from a string to a hash here?

gems/smithy-schema/lib/smithy-schema/document/serializer.rb
```diff
@@ -0,0 +1,283 @@
+# frozen_string_literal: true
+
+require 'base64'
+require 'time'
+
+module Smithy
+  module Schema
+    module Document
+      # Serializes data into a document data.
+      class Serializer
```
mullermp: Should this really be a public API? I would think these classes should be private behind some methods in Document.

gems/smithy-schema/lib/smithy-schema/document/serializer.rb
```diff
@@ -0,0 +1,283 @@
...
+        #   parsed_json = {
+        #     "__type" => "smithy.ruby.tests#Structure",
+        #     "string" => "hello"
+        #   }
+        #
+        #   document = serializer.create_document(parsed_json)
+        #   # => an instance of Smithy::Schema::Document::Data
+        #   document.discriminator
+        #   # => "smithy.ruby.tests#Structure"
+        def create_document(data)
```
mullermp: This feels out of place - this seems like it would be Document.new (as a class method or as initialize).

gems/smithy-schema/lib/smithy-schema/document/serializer.rb
```diff
@@ -0,0 +1,283 @@
...
+        # Serializes a document data with optional formatting.
+        # @param [Data] document The document to serialize
+        # @param [Hash] opts serialization options
+        # @option opts [Boolean] :use_timestamp_format Whether to use the
+        #  `timestampFormat` trait or ignore it. The `timestampFormat` trait
+        #   is ignored by default.
+        # @option opts [Boolean] :use_json_name Whether to use `jsonName` trait
+        #   or just member name. The `jsonName` trait is ignored by default.
+        # @return [Hash] Serialized document data
+        def serialize_document(document, opts = {})
```
mullermp: This also feels out of place, perhaps serialized and deserialized methods (or somewhat similar naming like typed/raw) on Document should invoke these paths.

gems/smithy-schema/lib/smithy-schema/document/serializer.rb
```diff
@@ -0,0 +1,283 @@
...
+            if values.is_a?(Smithy::Schema::Structure)
+              @type_registry.shape_by_type(values.class)
+            else
+              @type_registry[values['__type']]
+            end
+          format_document_data(shape, values, opts)
+        end
+
+        def typed_document?(values)
+          (values.is_a?(Smithy::Schema::Structure) && @type_registry.shape_by_type(values.class)) ||
```
mullermp: Isn't this already checked? And wouldn't this always be true if it was a structure, because it would already be registered?

gems/smithy-schema/lib/smithy-schema/document/serializer.rb
```diff
@@ -0,0 +1,283 @@
...
+          when Numeric then Time.at(value)
+          else Time.parse(value)
+          end
+        end
+
+        def resolve_member_ref(ref, name)
+          ref.shape.member(name) || find_member_ref_by_names(ref, name)
+        end
+
+        def find_member_ref_by_names(ref, name)
```
mullermp: This seems inefficient. Similar to what we do in codecs, for structure and union, you will want to iterate the shape members and not the values, then you can check json name that way. You're doing a loop for every member, so it's n^2 performance.

gems/smithy-schema/lib/smithy-schema/document/serializer.rb
```diff
@@ -0,0 +1,283 @@
...
+          ref.shape.members.values.find do |r|
+            r.traits['smithy.api#jsonName'] == name || r.member_name == name
+          end
+        end
+
+        def resolve_member_name(member_ref, opts)
```
mullermp: Check out the location_name approach in my PR - we should use similar terms. You can easily handle this with || optionality.

gems/smithy-schema/lib/smithy-schema/document/deserializer.rb
```diff
@@ -0,0 +1,182 @@
...
+            begin
+              fractional_time = Time.parse(value).to_f
+              Time.at(fractional_time).utc
+            rescue ArgumentError
+              raise "unhandled timestamp format `#{value}'"
+            end
+          end
+        end
+
+        def union(ref, values, target = nil) # rubocop:disable Metrics/AbcSize
```
mullermp: This one is wrong I think - just found this out today. Check out what is in my PR.

gems/smithy-schema/spec/smithy-schema/document/serializer_spec.rb
```diff
@@ -0,0 +1,208 @@
...
+require_relative '../../spec_helper'
+require_relative '../../support/schema_helper'
+
+module Smithy
+  module Schema
+    module Document
+      describe Serializer do
+        let(:shapes) do
+          shapes = SchemaHelper.sample_shapes
+          shapes['smithy.ruby.tests#Structure']['members']['timestampDateTime'] = {
```
mullermp: I would prefer if you move these definitions closer to the test (in the actual tests where they are needed) - it's easier to manage tests that way if they are discrete.

mullermp (commented):
Nice. Looking better. Still have a bunch of comments though .. In general I think we can still simplify and also be less aggressive on validation and instead be more permissive where possible.

alextwoods (commented):
Nice - its generally looking good.  
I think the functionality from the Document::Data class could be moved into `Document` as a class (unless theres some reason I'm missing).  I also understand why the Document serializer and deserializer exist separately and require a type registery - but I think I would lean towards the public interface for serializing/deserializing documents living on the top level class - it could still require a type registry to be provided and could use these classes under the hood to implement it (and they could then be api private).

mullermp:
That's effectively what I was also saying but I agree.

gems/smithy-schema/lib/smithy-schema/document.rb
jterapin: From what I recall - I believe I wanted to the Document (De)serializers to be nested under the Document namespace.

gems/smithy-schema/lib/smithy-schema/document/serializer.rb
jterapin: Yeah... Ideally, I would like that but for typed documents to work properly (deserializing nested typed documents) - it needs to reference a specific `type_registry` to (de)serialize properly.  

This is prob why Dowling wanted TypeRegistry to handle these responsibilities.

gems/smithy-schema/lib/smithy-schema/document/serializer.rb
jterapin: Yeah this is wrong, I'll fix.

gems/smithy-schema/lib/smithy-schema/document/serializer.rb
jterapin: I could create a class method under Document that references this file to handle (under the hood).

gems/smithy-schema/lib/smithy-schema/document/serializer.rb
jterapin: Yeah I can make class methods at the Document level

gems/smithy-schema/spec/smithy-schema/document/serializer_spec.rb
jterapin: That's fair - I'll move it.

gems/smithy-schema/spec/support/schema_helper.rb
jterapin: Ohh, I see why. Smithy-Schema's `spec_helper` never required `smithy`...I'll update.

gems/smithy-schema/lib/smithy-schema/document/serializer.rb
jterapin: Yeah, that makes sense. I could try that approach to see how that works. 
This would mean that this specific instance will be unique to that document only - which follows other codec patterns.

gems/smithy-schema/lib/smithy-schema/document/serializer.rb
jterapin: Yeah, I'll refactor - now that unions are not included in type registry.

gems/smithy-schema/lib/smithy-schema/type_registry.rb
jterapin: Yeah, we can populate this from code generated side or build at runtime - depends on what we want. I'm not sure which will be preferred when given context is like EC2 for example. 

From what I gathered, we are expected customers to use the type registry so I felt that hiding this detail is better. I guess at that point, you could say we might as well just accept an array of shapes so we can build the `registry` and `shapes_by_type` (previous approach before this iteration)

Any preference?

The `shapes_by_type` is necessary to find shape schema based on runtime shape class. Since runtime shape does not know its own shape schema.

gems/smithy-schema/lib/smithy-schema/type_registry.rb
jterapin: They _should_ be. If a customer wants to manually add shapes for whatever reason, we should make sure there is a runtime representation of the shape. I don't think it hurts to have this check.

gems/smithy-schema/lib/smithy-schema/type_registry.rb
jterapin: I'm dumb. I completely misunderstood your feedback here - thought you meant the naming of it: https://github.com/smithy-lang/smithy-ruby/pull/282#discussion_r2044962268

Anyway, hope the changes make sense.

gems/smithy-schema/lib/smithy-schema/type_registry.rb
jterapin: Yup, that makes sense. Hope the changes I made makes sense.

projections/shapes/lib/shapes/schema.rb
jterapin: Good idea, I did something like below:

``` ruby
# some_service/schema.rb

module SomeService 
  module Schema

  ...

  class << self
   
    def type_registry; end
  end
end


```

gems/smithy-schema/lib/smithy-schema/document.rb
```diff
@@ -0,0 +1,172 @@
...
+      # @param [Hash] opts Formatting options
+      # @option opts [Boolean] :timestamp_format Whether to use the `timestampFormat`
+      #  trait or ignore it. The `timestampFormat` trait is ignored by default.
+      # @option opts [Boolean] :json_name Whether to use `jsonName` trait
+      #   or just member name. The `jsonName` trait is ignored by default.
+      def serialize_contents(type_registry, opts = {})
+        validate_document(type_registry)
+
+        opts[:type_registry] = type_registry
+        opts[:json] = true
```
jterapin, line 56: I couldn't think of anything better with naming but I use this flag to determine whether I need to encode/decode blobs (created document instance - its blob data should be encoded)

gems/smithy-schema/lib/smithy-schema/document_utils/deserializer.rb
```diff
@@ -0,0 +1,124 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Schema
+    module DocumentUtils
+      # Deserializes document data into runtime shape.
+      # @api private
```
jterapin, line 7: Starting these out in `@api private` for now

gems/smithy-schema/lib/smithy-schema/document_utils/serializer.rb
```diff
@@ -0,0 +1,178 @@
...
+          values.each.with_object({}) do |(key, value), data|
+            next if value.nil?
+
+            data[key.to_s] = shape(ref.shape.value, value)
+          end
+        end
+
+        def structure(ref, values)
+          ref.shape.members.each_with_object({}) do |(member_name, member_ref), data|
+            value = resolve_value(member_name, member_ref, values.to_h)
```
jterapin, line 114: Since we are iterating over members...  just checking a `member_name` within values is not sufficient - given that input values could come from different contexts

gems/smithy-schema/lib/smithy-schema/document_utils/serializer.rb
```diff
@@ -0,0 +1,178 @@
...
+        end
+
+        def union(ref, values)
+          data = {}
+          if values.is_a?(Smithy::Schema::Union)
+            _name, member_ref = ref.shape.member_by_type(values.class)
+            data[location_name(member_ref)] = shape(member_ref, values)
+          else
+            key, value = values.first
+            if (member_ref = resolve_member_ref(ref, key))
```
jterapin, line 142: Same as I have mentioned above.

gems/smithy-schema/lib/smithy-schema/document.rb
jterapin: Reverted back to Document class approach with some class methods.

gems/smithy-schema/lib/smithy-schema/type_registry.rb
jterapin: Ok - for now, I decided to just accept an array of shapes for now (just thinking about EC2 schema file, oof)  We can do a switchroo later anyways - it's not firm.

gems/smithy-schema/lib/smithy-schema/document/serializer.rb
jterapin: This PR now reflects this suggested approach.

gems/smithy-schema/lib/smithy-schema/document/serializer.rb
jterapin: I THINK for now - we can start this out as private until we see that typed documents/registry takes some time to settle down.

gems/smithy-schema/lib/smithy-schema/document/serializer.rb
jterapin: Updated my approach.

gems/smithy-schema/lib/smithy-schema/document/serializer.rb
jterapin: Yeah - I streamlined closely to how a general codec serializer works. The updated approach does that but - we still need to do some checks since the given value key - could be a jsonName....symbolized key or memberName reflected on the modeled shape.

gems/smithy-schema/lib/smithy-schema/document/serializer.rb
jterapin: Yup, brought that over.

gems/smithy-schema/lib/smithy-schema/document/deserializer.rb
jterapin: yeah - saw that so updated my pr to reflect that

gems/smithy/spec/support/examples/schema_examples.rb
```diff
@@ -385,4 +385,23 @@ def expect_generated_shape(subject, shape_class, shape_hash)
...
+    end
+
+    it 'generates a type registry' do
+      expect(subject).to be_a(Smithy::Schema::TypeRegistry)
+    end
+
+    # TODO: failing due to synthetic shape changes
+    # it 'contains a registry of typed shapes' do
+    #   expect(subject.keys).to match_array(typed_shapes.keys)
+    # end
```
jterapin: Commented out failing test due to synthetic input/output shape id changes

projections/weather/lib/weather/schema.rb
```diff
@@ -83,5 +83,11 @@ module Schema
         operation[:paginator] = Paginators::ListCities.new
       end)
     end
+
+    class << self
+      def type_registry
+        Smithy::Schema::TypeRegistry.new([CityCoordinates, CitySummary, GetCityInput, GetCityOutput, GetCurrentTimeOutput, GetForecastInput, GetForecastOutput, ListCitiesInput, ListCitiesOutput, NoSuchResource])
```
jterapin: Could go back to doing map, etc but this will increase file size for services like ec2

jterapin (commented):
I tried to look over every comment that I wrote (since I like to write it as I go to remember where I left off) so if something doesn't make sense, let me know! Thanks!

mullermp (approved):

PR #283, opened 2025-03-07
Refactor protocols as plugins

mullermp:
This approach refactors the protocol implementations to be plugins. The major advantages of this approach are that we can remove ~400 generated plugins (one per service), control how and when protocol serde occurs, and support multi-protocol cases with simple string configuration values.

Adding protocols via welds is done with plugins as normal - the for? check can check for presence of traits.

The codec code should also be cleaned up - to be a module and split out build/parse logic.


gems/smithy-client/lib/smithy-client/plugins/content_length.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    module Plugins
+      # @api private
+      class ContentLength < Plugin
+        # @api private
+        class Handler < Client::Handler
```
jterapin, line 9: I know it's obvious here but I prefer that we always name our handlers - than leaving it generic

gems/smithy-client/lib/smithy-client/plugins/content_length.rb
```diff
@@ -0,0 +1,26 @@
...
+          # https://github.com/ruby/net-http/blob/master/lib/net/http/requests.rb
+          METHODS_WITH_BODY = Set.new(%w[POST PUT PATCH PROPFIND PROPPATCH MKCOL LOCK UNLOCK])
+
+          def call(context)
+            body = context.request.body
+            if body.respond_to?(:size) && METHODS_WITH_BODY.include?(context.request.http_method)
```
jterapin, line 15: This might be something that we need address when we get to eventstreams but how do we ensure that we do not set `Content-Length` header when the request is streaming?  Since `Content-Length` handler is always built it into the plugin list

From [Smithy documentation](https://smithy.io/2.0/additional-specs/protocols/smithy-rpc-v2.html#requests):
> Event streaming requests MUST NOT specify a content length (instead using Transfer-Encoding: chunked on HTTP/1.1).

gems/smithy-client/lib/smithy-client/rpc_v2_cbor/build_handler.rb
```diff
@@ -0,0 +1,60 @@
...
+          build_url(context)
+        end
+
+        def build_body(context)
+          Codecs::CBOR.new.serialize(context.operation.input, context.params)
+        end
+
+        def apply_content_type_header(context)
+          return if context.operation.input == Schema::Shapes::Prelude::Unit
+          return if event_stream?(context.operation.input)
```
jterapin: Can you add a TODO to handle event stream cases when it is implemented so that we won't forget?

gems/smithy-client/lib/smithy-client/rpc_v2_cbor/build_handler.rb
```diff
@@ -0,0 +1,60 @@
...
+        def event_stream?(shape)
+          shape.members.each_value do |member_shape|
+            return true if event_stream_shape?(member_shape.shape)
+          end
+          false
+        end
+
+        def event_stream_shape?(shape)
+          shape.traits.include?('smithy.api#streaming') && shape.is_a?(Shapes::Union)
+        end
```
jterapin, line 60: I imagine these methods will be frequently used across protocols - maybe we could consider a utility module of somekind?

jterapin (commented):
With the refactoring, there's few things that we need to address in the PR:
* Document how customer could switch protocol on the client if they choose to do so
  * We should also provide a test demonstrating this as well since this is one of the key objectives of Serialization and Schema Decoupling. I understand that we don't have any other protocol plugin for smithy-ruby but maybe use a fake protocol or something.
* Document on how does the customer switch back to using the modeled protocol default? I think we should document that adding a new protocol plugin will override the former and if they choose to go back to using the default, they will need to pass in 'smithy-rpc-v2-cbor'  for the protocol config to use it on the client. 
* Documentation on how one will implementing a custom protocol plugin, what does it require at minimum for the plugin to be sufficient to add and just work on the client?

Also - what are our plans of bringing AWS-ism into RPCv2CBOR? I'm guessing there will be a weld that overrides the Smithy-Client's RPCv2CBOR plugin with the one from the AWS-Core? We also may want to add TODO somewhere about resolving AWS-related query trait somewhere so this is not admiss

gems/smithy-client/lib/smithy-client/plugins/content_length.rb
mullermp: The convention I'm following is to use Handler for one handler, because it's obvious, or concrete names when there's more than one.

gems/smithy-client/lib/smithy-client/plugins/content_length.rb
mullermp: When we implement event streams, we can conditionally add this plugin, or skip it for a request with event stream. For now I think it's fine.

gems/smithy-client/lib/smithy-client/rpc_v2_cbor/build_handler.rb
mullermp: We will not forget. The protocol tests will fail.

gems/smithy-client/lib/smithy-client/rpc_v2_cbor/build_handler.rb
mullermp: Let's see how it pans out for aws-sdk-core, but I don't think we need anything right now.

jterapin (approved):
Approving with expectation that tests are being added, client documentation to expand on how to go about switching protocol plugin and then, eventually - we will do wiki/dev guide on making your own "custom" protocol

PR #284, opened 2025-03-07
Smithy CBOR gem

mullermp:
Moves the CBOR utilities into their own cbor gem where the codec can be defined.


gems/smithy-cbor/spec/spec_helper.rb
```diff
@@ -0,0 +1,111 @@
...
+  add_filter '/spec/'
+  add_filter 'gems/smithy/'
+  add_filter 'gems/smithy-client/'
+  add_filter 'gems/smithy-schema/'
+end
+
+require 'smithy-cbor'
+
+require 'smithy'
+require_relative '../../smithy-client/spec/support/client_helper'
```
mullermp: Later I think we can make a schema_helper and rely on it from smithy-schema to test the codecs.

jterapin: That sounds more appropriate, could add a TODO here so we won't forget?

gems/smithy-cbor/lib/smithy-cbor/codec.rb
```diff
@@ -0,0 +1,34 @@
...
+    #   * Update (de)serializing document types
+    #   * Allow user to pass in their preferred type to deserialize
+    #     If it fails, resort to deserializing type on the shape.
+    module Codec
+      include Schema::Shapes
+
+      # @param [Shape] shape
+      # @param [Object] data
+      # @return [String, nil] the encoded bytes in CBOR format
+      def self.serialize(shape, data)
```
jterapin, line 21: Generally, we want these codecs to take `options = {}` as a param for both (de)serialize methods. I don't see a reason to have options param for _generic_ cbor cases but we could just have anyway to be consistent with other codecs (like generic json codec will for sure have it. I'm not entirely sure if that is the case for Xml/Query). 

I don't have an option on this so feel free to do what makes sense.

In the context of cbor + aws, we do reference query compatible trait when deserializing. We talked about this previously - we could solve this by using welds and etc. But when I think about smithy-cbor being an entirely separate gem that services can just pull in for usage - maybe better to do query compatible trait stuff here after all to keep smithy-cbor as a standalone

jterapin (approved):
Nice, i like the flattening

gems/smithy-cbor/lib/smithy-cbor/codec.rb
mullermp: We can add options backwards compatibly to the method signature when we need to, so I don't think we should add it until requirements dictate it. The important thing is that we can do that when we have to.

As for the query compatible case, I still think we can keep that separate. To support it, we would overwrite the behavior of the module (create an AWS CBOR module in core and extend the methods) and make a new plugin that uses that module in the handlers. Worst case we can add query compatible to generic, but it's not a property of white label SDKs anyway (because query protocol is AWS).

PR #285, opened 2025-03-09
Retry errors

mullermp:
Adds the retry errors plugin with two retry strategies: standard and adaptive. Allows configuration as a string with flat options or a custom class. Passes the tests in the design.

Code is based off a hybrid of v3 and hearth v4.

### DIFF
--- a/gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
+++ b/gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
@@ -15,7 +15,7 @@ module Smithy
           doc_type: 'String, Class',
           docstring: <<~DOCS)
             The retry strategy to use when retrying errors. This can be one of the following:
-            * `standard` - A standardized set of retry rules across the AWS SDKs. This includes support
+            * `standard` - A standardized retry strategy used by the AWS SDKs. This includes support
               for retry quotas, which limit the number of unsuccessful retries a client can make.
             * `adaptive` - An experimental retry strategy that includes all the functionality of the
               `standard` strategy along with automatic client side throttling. This is a provisional
@@ -38,6 +38,7 @@ module Smithy
         option(
           :retry_backoff,
           default: Retry::EXPONENTIAL_BACKOFF,
+          doc_default: 'Smithy::Client::Retry::EXPONENTIAL_BACKOFF',
           doc_type: 'lambda',
           docstring: <<~DOCS)
             A callable object that calculates a backoff delay for a retry attempt. The callable
@@ -50,9 +51,9 @@ module Smithy
           default: true,
           doc_type: 'Boolean',
           docstring: <<~DOCS)
-            When true, the request will sleep until there is sufficient client side capacity
-            to retry the request. When false, the request will raise `RetryCapacityNotAvailableError`
-            and will not retry instead of sleeping.
+            When true, the request will sleep until there is sufficient client side capacity to retry
+            the request. When false, the request will raise a `CapacityNotAvailableError` and will
+            not retry instead of sleeping.
           DOCS
 
         def after_initialize(client)
--- a/gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
+++ b/gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
@@ -1,0 +1,84 @@
+# frozen_string_literal: true
+
+require 'pathname'
+
+require_relative '../retry/handler'
+
+module Smithy
+  module Client
+    module Plugins
+      # @api private
+      class RetryErrors < Plugin
+        option(
+          :retry_strategy,
+          default: 'standard',
+          doc_type: 'String, Class',
+          docstring: <<~DOCS)
+            The retry strategy to use when retrying errors. This can be one of the following:
+            * `standard` - A standardized set of retry rules across the AWS SDKs. This includes support
+              for retry quotas, which limit the number of unsuccessful retries a client can make.
+            * `adaptive` - An experimental retry strategy that includes all the functionality of the
+              `standard` strategy along with automatic client side throttling. This is a provisional
+              strategy that may change behavior in the future.
+
+            In addition to these canned strategies, a custom retry strategy can be provided by passing
+            a class that implements the following methods:
+            * `acquire_initial_retry_token(token_scope)`
+            * `refresh_retry_token(retry_token, error_info)`
+            * `record_success(retry_token)`
+          DOCS
+
+        option(
+          :retry_max_attempts,
+          default: 3,
+          doc_type: Integer,
+          docstring: <<~DOCS)
+            The maximum number attempts that will be made for a single request, including
+            the initial attempt. Used in the `standard` and `adaptive` retry strategies.
+          DOCS
+
+        option(
+          :retry_backoff,
+          default: Retry::EXPONENTIAL_BACKOFF,
+          doc_type: 'lambda',
+          docstring: <<~DOCS)
+            A callable object that calculates a backoff delay for a retry attempt. The callable
+            should accept a single argument, `attempts`, that represents the number of attempts
+            that have been made. Used in the `standard` and `adaptive` retry strategies.
+          DOCS
+
+        option(
+          :adaptive_retry_wait_to_fill,
+          default: true,
+          doc_type: 'Boolean',
+          docstring: <<~DOCS)
+            When true, the request will sleep until there is sufficient client side capacity
+            to retry the request. When false, the request will raise `RetryCapacityNotAvailableError`
+            and will not retry instead of sleeping.
+          DOCS
+
+        def after_initialize(client)
+          config = client.config
+          config.retry_strategy =
+            case config.retry_strategy
+            when 'standard'
+              Retry::Standard.new(
+                max_attempts: config.retry_max_attempts,
+                backoff: config.retry_backoff
+              )
+            when 'adaptive'
+              Retry::Adaptive.new(
+                max_attempts: config.retry_max_attempts,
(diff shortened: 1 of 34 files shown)


gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
```diff
@@ -0,0 +1,82 @@
...
+
+require 'pathname'
+
+require_relative '../retry/handler'
+
+module Smithy
+  module Client
+    module Plugins
+      # @api private
+      class RetryErrors < Plugin
```
jterapin, line 11: I'm not sure if `RetryErrors` is a good name for this plugin, since it reads to me like a plugin that handles errors that comes out of retries. Feels more like `RetryMode` or `RetryStrategy` plugin?

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
```diff
@@ -0,0 +1,82 @@
...
+        option(
+          :retry_strategy,
+          default: 'standard',
+          doc_type: 'String, Class',
```
jterapin, line 15: What are your thoughts on allowing `Symbol` usage too? Or is that too funky?

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
```diff
@@ -0,0 +1,82 @@
...
+          docstring: <<~DOCS)
+            The retry strategy to use when retrying errors. This can be one of the following:
+            * `standard` - A standardized set of retry rules across the AWS SDKs. This includes support
```
jterapin: Remove the `AWS SDKs` since this plugin is intended for generic. Looking at [Hearth docs](https://github.com/smithy-lang/smithy-ruby/blob/main/hearth/lib/hearth/retry/standard.rb#L5)- we could say:
> A standard retry strategy for retrying requests.

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
```diff
@@ -0,0 +1,82 @@
...
+            that have been made. Used in the `standard` and `adaptive` retry strategies.
+          DOCS
+
+        option(
+          :adaptive_retry_wait_to_fill,
+          default: true,
+          doc_type: 'Boolean',
+          docstring: <<~DOCS)
+            When true, the request will sleep until there is sufficient client side capacity
+            to retry the request. When false, the request will raise `RetryCapacityNotAvailableError`
```
jterapin: I believe the actual raise will be `CapacityNotAvailableError` based error outlined under `retry.rb`

gems/smithy-client/lib/smithy-client/retry.rb
```diff
@@ -0,0 +1,39 @@
...
+require_relative 'retry/standard'
+
+module Smithy
+  module Client
+    module Retry
+      # The maximum backoff delay for retrying requests.
+      MAX_BACKOFF = 20
+
+      # Raised when the adaptive retry strategy is unable to acquire capacity.
+      class CapacityNotAvailableError < RuntimeError; end
```
jterapin: This error class is only intended for Adaptive Retries so on the inital thought,  it makes more sense to be under Adaptive but I could see that you may want to flatten it here

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
```diff
@@ -0,0 +1,82 @@
...
+
+require_relative '../retry/handler'
+
+module Smithy
+  module Client
+    module Plugins
+      # @api private
+      class RetryErrors < Plugin
+        option(
+          :retry_strategy,
```
jterapin, line 13: I was looking over the SEP - seems like retry configuration names like `retry_mode` and `max_attempts` are used. While the client configuration naming does not need to be consistent, I figured it doesn't to hurt to "match" here? Just a thought. Nothing major.

gems/smithy-client/lib/smithy-client/retry/adaptive.rb
```diff
@@ -0,0 +1,66 @@
...
+
+        # @return [Integer]
+        attr_reader :max_attempts
+
+        # @return [Boolean]
+        attr_reader :wait_to_fill
+
+        def acquire_initial_retry_token(_token_scope = nil)
+          @client_rate_limiter.token_bucket_acquire(1, wait_to_fill: @wait_to_fill)
+          Token.new(retry_count: 0)
```
jterapin: There is a default set to `retry_count` so I don't think we need to explicitly pass here.

gems/smithy-client/lib/smithy-client/retry/adaptive.rb
```diff
@@ -0,0 +1,66 @@
...
+        # @return [#call]
+        attr_reader :backoff
+
+        # @return [Integer]
+        attr_reader :max_attempts
+
+        # @return [Boolean]
+        attr_reader :wait_to_fill
+
+        def acquire_initial_retry_token(_token_scope = nil)
```
jterapin, line 35: Is there a reason why we keep `_token_scope` around? I was looking at its usage for both Standard and Adaptive Retry and it remains as an unused param. 

The only I could think of if there is a customer that wants to implement their own retry plugin...

gems/smithy-client/lib/smithy-client/retry/standard.rb
```diff
@@ -0,0 +1,52 @@
...
+
+        # @return [#call]
+        attr_reader :backoff
+
+        # @return [Integer]
+        attr_reader :max_attempts
+
+        def acquire_initial_retry_token(_token_scope = nil)
+          Token.new(retry_count: 0)
+        end
```
jterapin: Ditto from my comments on Adaptive

projections/shapes/lib/shapes/client.rb
```diff
@@ -102,6 +108,24 @@ class Client < Smithy::Client::Base
     # @option options [Boolean] :raise_response_errors (true)
     #  When `true`, response errors are raised. When `false`, the error is placed on the
     #  output in the {Smithy::Client::Output#error error accessor}.
+    # @option options [lambda] :retry_backoff (#<Proc:0x000000011dffeb00 /Users/mamuller/workplace/smithy-ruby/gems/smithy-client/lib/smithy-client/retry.rb:18 (lambda)>)
```
jterapin: Huh, it references your alias/workspace here - that don't look right.

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
mullermp: I copied from v3. It retries errors :). Can think on other names.

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
mullermp: I liked the idea of symbols but it falls apart when config names have dashes, like with smithy-rpc-v2-cbor.

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
mullermp: I put this intentionally but we can maybe remove it. What I was trying to convey is that this method of retrying was developed by AWS and is now generic usage.

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
mullermp: Will fix

gems/smithy-client/lib/smithy-client/retry.rb
mullermp: I think there was a reason I put it here but I can't remember. I'll investigate.

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
mullermp: The SEP was written before SRA. This is the new terminology.

gems/smithy-client/lib/smithy-client/retry/adaptive.rb
mullermp: Will fix

gems/smithy-client/lib/smithy-client/retry/adaptive.rb
mullermp: Yeah, I'm not sure what to do with it. This is a classic "over engineering" example of the SRA where we are predicting future requirements. The idea is that retrying can apply to "scopes" such as only in one AWS partition. I think we can add this later with a suitable default and not break existing usage. Especially also since we don't have a base class anymore, this should be fine. We can probably remove it.

projections/shapes/lib/shapes/client.rb
mullermp: Weird, I'll fix.

gems/smithy-client/lib/smithy-client/retry.rb
mullermp: Oh, this is because it is raised inside the client rate limiter. I don't think we want the error defined under that class.

gems/smithy-client/lib/smithy-client/retry/adaptive.rb
mullermp: On second thought, it breaks others if we add it later. So that could be intentional if we did need to add it.

jterapin: Gotcha, good to know the context - incase we are wondering in the future.

jterapin (approved):
LGTM

PR #286, opened 2025-03-11
Revert back to protocol class but with a generic plugin

mullermp:
Uses a generic plugin instead with protocol classes, instead of code generated plugins.


gems/smithy-client/lib/smithy-client.rb
```diff
@@ -36,6 +36,9 @@
 require_relative 'smithy-client/net_http/connection_pool'
 require_relative 'smithy-client/net_http/handler'
 
+# temp - should be another gem?
+require_relative 'smithy-client/rpc_v2_cbor/protocol'
```
jterapin, line 40: Maybe? If we want to provide protocol-specific tooling for smithy-server users

gems/smithy-client/lib/smithy-client/plugins/protocols.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    module Plugins
+      # @api private
+      class Protocols < Plugin
```
jterapin: Should this be just `Protocol` instead of plural? Given that a single protocol is used at a time.

gems/smithy/lib/smithy/views/client/client.rb
```diff
@@ -71,11 +75,18 @@ def option_tag(option)
...
+          default = option_default(option)
           tag << " (#{default})" if default
           tag.string
         end
 
+        def option_default(option)
+          default = option.doc_default || option.default
+          return default unless option.name == :protocol
+
+          default.gsub('<DEFAULT_PROTOCOL>', protocols.keys.first || 'nil')
```
jterapin, line 87: This is pretty cool solution

gems/smithy/lib/smithy/welds/rpc_v2_cbor.rb
```diff
@@ -1,23 +1,8 @@
...
-    #  The priority order is as follows (within AWS context):
-    #  - Smithy RPCv2 CBOR
-    #  - AWS JSON 1.0
-    #  - AWS JSON 1.1
-    #  - REST JSON
-    #  - REST XML
-    #  - AWS/Query
-    #  - EC2/Query
-    #  Possible solution: priority system similar to the handler registration
+    # Adds the RPCv2 CBOR protocol to the client if the service has the trait.
```
jterapin, line 5: Another TODO is to resolve AWS-ism for RPCv2CBOR protocol

gems/smithy-cbor/lib/smithy-cbor/codec.rb
```diff
@@ -12,22 +12,27 @@ module CBOR
     #   * Update (de)serializing document types
     #   * Allow user to pass in their preferred type to deserialize
     #     If it fails, resort to deserializing type on the shape.
-    module Codec
+    class Codec
```
jterapin, line 13: Since you are already touching this file - can you remove "support handling of typed documents" and "Update (de)serializing document types" TODOs - RPCv2CBOR does not support Document shapes when I was looking into document types. Thanks!

gems/smithy-client/lib/smithy-client/plugins/protocols.rb
```diff
@@ -0,0 +1,59 @@
...
+        def before_initialize(client_class, options)
+          protocol = options[:protocol]
+
+          case protocol
+          when nil
+            protocol_class = client_class.protocols.values.first
+            options[:protocol] = protocol_class.new if protocol_class
+          when String
+            protocol_class = client_class.protocols[protocol]
+            raise ArgumentError, "Unknown protocol: #{protocol}" unless protocol_class
```
jterapin: Good call on adding this - we want to only support what protocols are supported by the model/server. I like that the Client file will contain these bring these details over.

jterapin (approved):
I think this is a good solution toward the generic protocol plugin approach. 

We are able do the following customer experience:
* Change their default protocol by passing in a string:
```
SomeService::Client.new(protocol: 'railsjson')
```
* Change the default by passing in a class with preferred options:
```
json_protocol = Smithy::Client::Protocols::RailsJson.new(json_name: false)
SomeService::Client.new(protocol: json_protocol)
```
* Be alerted when unsupported protocol is used:
```
SomeService::Client.new(protocol: unsupported_protocol)
# => Raise Error
```

Other additional benefits:
* Client configuration is not overloaded with protocol-related config options.
* Documented default protocol in the generated docs
* No need to generate more files on the codegen's side
* Consistent Builder/Parser handlers

PR #287, opened 2025-03-17
Identity auth

mullermp:
Not at all working yet


gems/smithy-client/lib/smithy-client/identities/http_login.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    module Identities
+      # Identity class for bearer token authentication.
```
jterapin: Update comment here since it's used for HttpBasic and HttpDigest

gems/smithy-client/lib/smithy-client/auth_scheme.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    # Base class for all AuthScheme classes.
+    class AuthScheme
```
jterapin, line 6: Rather than a class - could we have a map within a module? If we need to reference auth schemes for resolution. I'm trying to think of ways to simplify things.

gems/smithy-client/lib/smithy-client/identity_provider.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    # Basic Identity resolver that uses a proc to resolve an Identity.
+    class IdentityProvider
```
jterapin: I'm guessing we didn't want to use the word "resolver" again here so opt to using provider?

jterapin (commented):
I think we also need to cover:
https://smithy.io/2.0/spec/authentication-traits.html#httpapikeyauth-trait

gems/smithy-client/lib/smithy-client/auth_scheme.rb
mullermp: I think a class is preferred right now, or for it to be API private if possible..

gems/smithy-client/lib/smithy-client/identities/anonymous.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    module Identities
+      class Anonymous < Identity; end
```
jterapin, line 7: Minor - Documentation to follow suit with other identities.

gems/smithy-client/lib/smithy-client/plugins/sign_requests.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    module Plugins
+      # @api private
+      class SignRequests < Plugin
+        option(:anonymous_auth_scheme) do |_config|
+          Smithy::Client::AuthSchemes::Anonymous.new
+        end
```
jterapin: This config is already defined under AnonymousAuth Plugin - any reason why we have to duplicate?

gems/smithy-client/lib/smithy-client/signer.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    # Base class for all Signer classes.
+    class Signer
```
jterapin, line 6: Curious if this should this live under Signers module with other signers? Same goes for other auth classes that has a base that inherits from

gems/smithy-client/sig/smithy-client/identity_provider.rbs
```diff
@@ -0,0 +1,8 @@
+module Smithy
+  module Client
+    class IdentityProvider
```
jterapin: I don't think we need this rbs file anymore? I see that we have an identity provider interface defined

gems/smithy-client/spec/smithy-client/plugins/http_bearer_auth_spec.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+require_relative '../../spec_helper'
+
+require 'smithy-client/plugins/http_bearer_auth'
+
+module Smithy
+  module Client
+    module Plugins
+      describe HttpBearerAuth do
```
jterapin, line 10: Probably should cover HttpDigestAuth for our sanity? I think it's very similar to this but uses a different plugin module

gems/smithy/lib/smithy/views/client/auth_resolver.rb
```diff
@@ -0,0 +1,127 @@
+# frozen_string_literal: true
+
+require 'stringio'
```
jterapin: I didn't see any usages of `StringIO` below

gems/smithy/spec/interfaces/client/http_digest_auth_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+        expect(HttpDigestAuth::Client.plugins).to include(Smithy::Client::Plugins::HttpDigestAuth)
+      end
+
+      it 'adds the http digest auth scheme' do
+        expect(client.config.auth_schemes).to include('smithy.api#httpDigestAuth')
+      end
+
+      pending 'resolves http digest auth'
+
+      pending 'signs the request'
```
jterapin, line 22: TODO?

gems/smithy/spec/interfaces/welds/auth_schemes_spec.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+# require_relative '../../spec_helper'
+#
+# describe 'Welds: Auth Schemes' do
```
jterapin: Another TODO?

jterapin (approved):
Looks good - do we want to cover custom auth cases or is that a stretch to do? I guess it is highly unlikely that a customer will implement their own custom auth.

Was thinking of covering tests like this from previous smithy-ruby: https://github.com/smithy-lang/smithy-ruby/blob/main/codegen/projections/white_label/spec/auth_spec.rb#L78

gems/smithy-client/lib/smithy-client/plugins/sign_requests.rb
mullermp: I was trying to find a way to get rid of that plugin. I'm sure we can stuff it under sign and special case it in generation.

gems/smithy-client/lib/smithy-client/signer.rb
mullermp: In V4 old we had "signers::base" and "identities:base" but I'm not sure if we care to have that here. I think it makes sense to have it this way, similar to plugins and handlers.

gems/smithy-client/sig/smithy-client/identity_provider.rbs
mullermp: No we don't, good catch

gems/smithy-client/spec/smithy-client/plugins/http_bearer_auth_spec.rb
mullermp: Yes. It's similar to Basic. I forgot it.

gems/smithy/spec/interfaces/client/http_digest_auth_spec.rb
mullermp: Pending operates this way (it prints)

gems/smithy/spec/interfaces/welds/auth_schemes_spec.rb
mullermp: I meant to do this

PR #289, opened 2025-04-04
Pagination

mullermp:
Support the paginated trait by using methods on Output for iteration.


gems/smithy-client/lib/smithy-client/pageable_output.rb
```diff
@@ -0,0 +1,124 @@
...
+    # `each_item`:
+    #
+    #     weather = Weather::Client.new
+    #     weather.list_cities.each_item do |item|
+    #       puts item.name
+    #     end
+    #
+    # # Handling Paged Output Manually
+    # To handle paging yourself, use the output's `next_page?` method to verify
+    # there are more pages to retrieve, or use the last_page? method to verify
```
jterapin:
```suggestion
    # there are more pages to retrieve, or use the `last_page?` method to verify
```

gems/smithy-client/lib/smithy-client/plugins/pageable_output.rb
```diff
@@ -0,0 +1,37 @@
...
+  module Client
+    module Plugins
+      # @api private
+      class PageableOutput < Plugin
+        # @api private
+        class Handler < Client::Handler
+          def call(context)
+            output = @handler.call(context)
+            output.extend(Client::PageableOutput)
+            output.paginator = context.operation[:paginator] || NullPaginator.new
```
jterapin, line 13: Does operation shape has paginator?

gems/smithy-schema/lib/smithy-schema/shapes.rb
```diff
@@ -9,13 +9,25 @@ class Shape
         def initialize(options = {})
           @id = options[:id]
           @traits = options[:traits] || {}
+          @metadata = {}
```
jterapin, line 12: We previously used this attribute in the `ShapeRef` class from V3. 

I like that we have this around for flexibility but hard to keep track of all metadata usages over time (actually, is there a way to see all metadata usages in V3? So far, I just search for `ref[:` to see if there's usage but that's making an assumption that `ref` variable is used). 

Any ideas to elevate this?

gems/smithy/lib/smithy/views/client/paginators.rb
```diff
@@ -0,0 +1,113 @@
...
+            code << ''
+            code << 'tokens = Hash.new { |h, k| h[k] = {} }'
+            prev_token_setter = input_getter(@input_token, 'tokens')
+            code << "#{prev_token_setter} = prev_token"
+            code << 'tokens'
+            code
+          end
+
+          def items_code
+            return "raise NotImplementedError, 'item iteration is not implemented for this operation'" unless items?
```
jterapin: Nit: Line is so long, maybe we update make this two lines?

gems/smithy-client/lib/smithy-client/pageable_output.rb
```diff
@@ -0,0 +1,124 @@
...
+
+      # Yields the current and each following item to the given block.
+      # @yieldparam [Object] item
+      # @return [Enumerable, nil] Returns a new Enumerable if no block is given.
+      def each_item(&)
+        output = self
+        @paginator.items(output.data).each(&)
+        until output.last_page?
+          output = output.next_page
+          @paginator.items(output.data).each(&)
```
jterapin, line 104: n00b question: Why is `@paginator.items(output.data).each(&)` done twice here?

gems/smithy-client/lib/smithy-client/plugins/pageable_output.rb
jterapin: I see, you added a metadata to the `Shapes`

gems/smithy-schema/lib/smithy-schema/shapes.rb
```diff
@@ -9,13 +9,25 @@ class Shape
...
 
         # @return [Hash<String, Object>]
         attr_accessor :traits
+
+        # @return [Hash<Object, Object>]
+        def [](key)
+          @metadata[key]
+        end
+
+        # @param [Object] key
```
jterapin: Why Object? Maybe we should lean either symbol or string as keys for consistency? Thoughts?

gems/smithy/lib/smithy/views/client/paginators.rb
```diff
@@ -0,0 +1,113 @@
...
+        def module_name
+          @plan.module_name
+        end
+
+        def paginators
+          Model::ServiceIndex
+            .new(@model)
+            .operations_for(@plan.service)
+            .map do |id, operation|
+              operation_trait = operation.fetch('traits', {}).fetch('smithy.api#paginated', {})
```
jterapin: `.fetch('traits', {}).fetch('smithy.api#paginated', {})`  - could make a helper method? since i see it used both here and for `@service_trait`

gems/smithy/lib/smithy/views/client/paginators.rb
```diff
@@ -0,0 +1,113 @@
...
+          end
+
+          def items?
+            !!@items
+          end
+
+          # Builds a getter using the output shape and a path.
+          # This is used to get the value of the output token or items.
+          def output_getter(path)
+            getter = StringIO.new
```
jterapin: n00b question - This is the first time I seen `StringIO` used like this (although, it's heavily used in `endpoint_provider_class`). What are some pro/cons of doing it this way?

gems/smithy/lib/smithy/views/client/paginators.rb
```diff
@@ -0,0 +1,113 @@
...
+            getter << 'data'
+            shape = @output
+            path.split('.').each do |member|
+              target = shape['members'][member]['target']
+              shape = Model.shape(@model, target)
```
jterapin: Where is `shape` used after this assignment? Same goes for `input_getter` method below

gems/smithy/lib/smithy/views/client/schema.rb
```diff
@@ -249,23 +257,51 @@ def initialize(options = {})
...
+            @traits.key?('smithy.api#paginated')
+          end
+
+          def paginator
+            "Paginators::#{@name}.new"
+          end
+
+          def traits
+            @traits.except(*OMITTED_TRAITS)
+          end
```
jterapin: Tripped me up a bit since i noticed this (and `ServiceShape` below) wasn't using a similar pattern other shapes with `@traits` omitting traits from the getgo but I see why now... (`paginated?`)

jterapin: Could create `traits_view` or `traits_code` for View's usage and keep the `@trait` consistent between other view shapes here

projections/shapes/lib/shapes/paginators.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+# This is generated code!
+
+module ShapeService
+  # @api private
+  module Paginators
```
jterapin, line 7: If the service does not have any paginator traits on their service/operations - i think we should omit these pagination-related files, right?

projections/weather/lib/weather/paginators.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+# This is generated code!
+
+module Weather
+  # @api private
+  module Paginators
```
jterapin, line 7: Gonna need `.rbs` version of this file

jterapin (commented):
Lookin' good. I like that you push the `filter_traits` to the view shapes. That really streamlines the "View" side of things.

gems/smithy-client/lib/smithy-client/plugins/pageable_output.rb
mullermp: Yeah. I added it as metadata like in v3. Paginators aren't part of the schema (like a schema gem) but it's client context.

gems/smithy-schema/lib/smithy-schema/shapes.rb
mullermp: We have this in v3 on shape too, not just shaperef. Inspect on shape will print out metadata. What specifically is the concern you have about finding usage of it? I think it's a dumping ground for anything - any plugin/handler could add context to a shape like from a custom runtime trait or something.

gems/smithy/lib/smithy/views/client/paginators.rb
mullermp: Rubocop forces this because it's less than 120 characters (117)

gems/smithy-client/lib/smithy-client/pageable_output.rb
mullermp: We do it once to yield the current items that were returned with this service response, then we do it on subsequent pages. Just like above where we yield self and then yield output again.

gems/smithy-schema/lib/smithy-schema/shapes.rb
mullermp: We could scope it to symbol/string keys if we want but I just made it opaque because we can dump anything on it. You can use any object as a key in a hash in ruby.

projections/weather/lib/weather/paginators.rb
mullermp: They are API private - I don't think we need one for paginator classes - no interaction will be done here, but we can probably add rbs for next_page etc.

projections/shapes/lib/shapes/paginators.rb
mullermp: Eh. We could omit them and conditionally add the pagination if we want to. Right now for protocol tests we generate an empty file too. We can do a separate change for both of these and support autoloading after this.

gems/smithy/lib/smithy/views/client/schema.rb
mullermp: Here I just made the view classes responsible for how they want to render traits with the traits method, but I pass all the traits into the view class so it can decide what to do with them. I pushed the paginated check to the initializer. Let me know if that is less or more helpful.

gems/smithy/lib/smithy/views/client/paginators.rb
mullermp: Ah - I think this is a relic of an alternative I was working on. I think ideally member name like 'fooBar' on the model can be different than what we generate the struct member name as 'foo_bar'. I just make a massive assumption that it's underscored. I think it's fine for the 99% but maybe re-evaluate this later.

gems/smithy/lib/smithy/views/client/paginators.rb
mullermp: We can't use a string literal because of frozen string literals. But we can use `String` instead of `StringIO` - it may actually be faster than StringIO. Either is necessary for string building. StringIO is necessary if we need to seek.

jterapin (approved):
🚢✨

PR #290, opened 2025-04-09
Improve testing of client features

mullermp:
Extra protocol plugin tests and pushes stub response tests to client where it can use a default protocol, and pushes auth tests to the client.


gems/smithy-client/spec/smithy-client/plugins/http_digest_auth_spec.rb
```diff
@@ -77,6 +80,10 @@ module Plugins
           provider = client.config.http_login_provider
           expect(provider).to be_nil
         end
+
+        context 'signing' do
+          it 'signs in the header'
+        end
```
jterapin, line 86: Missing part of a test here?

jterapin (approved):
Just need to fix the rbs failures + some partially eaten test that I commented on.

gems/smithy-client/spec/smithy-client/plugins/http_digest_auth_spec.rb
mullermp: It's pending implementation

PR #291, opened 2025-04-24
Improvements to make V4 work

mullermp:
These are changes necessary to make v4 work correctly (and pass rest-json protocols).


gems/smithy-client/lib/smithy-client/plugins/host_prefix.rb
```diff
@@ -0,0 +1,69 @@
...
+    module Plugins
+      # @api private
+      class HostPrefix < Plugin
+        option(
+          :disable_host_prefix_injection,
+          default: false,
+          doc_type: 'Boolean',
+          docstring: 'When true, the SDK will not prepend the modeled host prefix to the endpoint.'
+        ) do |_config|
+          value = ENV['DISABLE_HOST_PREFIX_INJECTION'] || 'false'
```
jterapin, line 14: Wondering if we have a set plan for accommodating `AWS_DISABLE_HOST_PREFIX_INJECTION` version of this ENV variable.

gems/smithy-client/lib/smithy-client/plugins/host_prefix.rb
```diff
@@ -0,0 +1,69 @@
...
+          def label_value(input, label, params)
+            name = nil
+            input.members.each do |member_name, member_shape|
+              next unless member_shape.traits.include?('smithy.api#hostLabel')
+              next unless member_shape.name == label
+
+              name = member_name
+            end
+            raise ArgumentError, "#{label} is not a valid host label" if name.nil?
+            raise ArgumentError, "params[#{name}] must not be nil or blank" if params[name].nil? || params[name].empty?
```
jterapin, line 61: Nice, helpful error messages - when comparing to what we had previously

gems/smithy-client/lib/smithy-client/plugins/request_compression.rb
```diff
@@ -0,0 +1,202 @@
...
+
+        option(
+          :disable_request_compression,
+          default: false,
+          doc_type: 'Boolean',
+          docstring: 'When true, the request body will not be compressed for supported operations.'
+        ) do |_config|
+          value = ENV['DISABLE_REQUEST_COMPRESSION'] || 'false'
+          Util.str_to_bool(value)
+        end
```
jterapin, line 23: Ditto on our plan on accommodating AWS variation of these ENV variables + profile setting.

gems/smithy-client/spec/smithy-client/plugins/request_compression_spec.rb
```diff
@@ -0,0 +1,141 @@
...
+        end
+
+        it 'does not add the handler if :disable_request_compression is true' do
+          client = client_class.new(disable_request_compression: true)
+          expect(client.handlers).not_to include(RequestCompression::Handler)
+        end
+
+        it 'validates :disable_request_compression option as a boolean' do
+          client_class.new(disable_request_compression: true)
+          client_class.new(disable_request_compression: false)
```
jterapin, line 45: Do we need these?

gems/smithy-client/spec/smithy-client/plugins/request_compression_spec.rb
```diff
@@ -0,0 +1,141 @@
...
+        it 'allows :disable_request_compression to be set via ENV' do
+          ENV['DISABLE_REQUEST_COMPRESSION'] = 'true'
+          client = client_class.new
+          expect(client.config.disable_request_compression).to be(true)
+        end
+
+        it 'validates :request_min_compression_size_bytes option as an integer within a range' do
+          msg = /must be a non-negative integer value between `0` and `10,485,760` bytes inclusive/
+          client_class.new(request_min_compression_size_bytes: 0)
+          client_class.new(request_min_compression_size_bytes: 10_485_760)
```
jterapin, line 59: Ditto here

gems/smithy-client/spec/smithy-client/plugins/request_compression_spec.rb
```diff
@@ -0,0 +1,141 @@
...
+            client = client_class.new(stub_responses: true)
+            output = client.operation(string: large_body)
+            compressed_body = Zlib::GzipReader.new(output.context.request.body)
+            expect(compressed_body.read).to include(large_body) # cbor
+          end
+
+          it 'uses the first supported encoding found' do
+            shapes['smithy.ruby.tests#Operation']['traits'] = {
+              'smithy.api#requestCompression' => { 'encodings' => %w[custom gzip] }
+            }
```
jterapin, line 98: I like this pattern where we just edit the shapes to what we need for our tests.

gems/smithy-client/spec/smithy-client/plugins/request_compression_spec.rb
```diff
@@ -0,0 +1,141 @@
...
+          end
+
+          it 'compresses the body using gzip and sets the content-encoding header' do
+            output = client.operation(string: large_body)
+            expect(output.context.request.headers['Content-Encoding']).to eq('gzip')
+          end
+
+          it 'preserves the compressed body' do
+            client = client_class.new(stub_responses: true)
```
jterapin: Could use the same client from line 24? since its stubbed is already set to true

gems/smithy-schema/lib/smithy-schema/shapes.rb
```diff
@@ -34,6 +34,7 @@ def []=(key, value)
       class Structure < Shape
         def initialize(options = {})
           super
+          @type = options[:type] # TODO: send() in Base
```
jterapin: I forgot to bring this up while implementing typed documents - I don't think that `Structure` should have `type` defined - might be best that these are defined on the actual `StructureShape` and `UnionShape` themselves.

Currently, this class is inherited by `IntEnumShape` and `EnumShape` - if we do checks like `if shape.type...` - we need to ensure that the shape is not `IntEnum` or `Enum` shape

projections/weather/lib/weather/schema.rb
```diff
@@ -51,45 +51,45 @@ module Schema
...
 
     SERVICE = ServiceShape.new do |service|
-      service.id = 'example.weather#Weather'
-      service.name = 'Weather'
-      service.version = '2006-03-01'
-      service.traits = {}
+      service.id = "example.weather#Weather"
+      service.name = "Weather"
+      service.version = "2006-03-01"
+      service.traits = {"smithy.api#paginated" => {"inputToken" => "nextToken", "outputToken" => "nextToken", "pageSize" => "pageSize"}}
```
jterapin, line 64: This is supposed to be omitted right? Noticed that the `smithy.api#paginated` trait is generated for the operation as well

gems/smithy-cbor/lib/smithy-cbor/deserializer.rb
```diff
@@ -77,6 +78,7 @@ def structure(shape, values, target = nil)
       end
 
       def union(shape, values, target = nil)
+        # TODO: delete target instead of checking key?
         key, value = values.flatten
         return nil if key.nil? || key == ' __target'
```
jterapin, line 83: I think this is my bad - I meant to put `__type` instead of `__target`

gems/smithy-client/spec/smithy-client/plugins/checksum_required_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+          client_class.clear_plugins
+          client_class.add_plugin(sample_client::Plugins::Auth)
+          client_class.add_plugin(sample_client::Plugins::Endpoint)
+          client_class.add_plugin(AnonymousAuth)
+          client_class.add_plugin(ChecksumRequired)
+          client_class.add_plugin(Protocol)
+          client_class.add_plugin(SignRequests)
+          client_class.add_plugin(StubResponses)
+          client_class
+        end
```
jterapin, line 25: What do you think about a method ClientHelper that adds in the "default" plugins or is it just better to do this manually?

gems/smithy-client/lib/smithy-client/plugins/host_prefix.rb
mullermp: I think so, by overwriting the option in a different plugin.

gems/smithy-client/spec/smithy-client/plugins/request_compression_spec.rb
mullermp: Maybe? I just check both valid values.

gems/smithy-client/spec/smithy-client/plugins/request_compression_spec.rb
mullermp: We check the boundary of possible values. Seems fine.

gems/smithy-client/spec/smithy-client/plugins/request_compression_spec.rb
mullermp: Yes same

gems/smithy-schema/lib/smithy-schema/shapes.rb
mullermp: If we're not actually making typed enums I wonder if we should even have those shapes anyway. We also don't validate a value in the enum members for forwards compatibility. We could make these both string and integers.

gems/smithy-client/spec/smithy-client/plugins/checksum_required_spec.rb
mullermp: I'm adding a minimal set that the spec needs. It's different per test. We could alternatively not clear all plugins but it doesn't isolate testing behavior.

gems/smithy-cbor/lib/smithy-cbor/deserializer.rb
mullermp: Ok, we need to revisit this anyway, and I don't see any tests related to it.

projections/weather/lib/weather/schema.rb
mullermp: I had removed it from omission. I'm not sure. I guess it makes sense to remove it again. Let me ask about model purity in the channel again.

gems/smithy-schema/lib/smithy-schema/shapes.rb
jterapin: As for typed enums, I think we decided against that but I don't recall why.

I think we still need to to have these enum shapes as a representation of smithy model data type.

mullermp: Added a todo to fix this. I think we can re-evaluate enums with defaults.

PR #292, opened 2025-04-24
Waiters

richardwang1124:
*Description of changes:*
Support the `#waitable` trait with waiters. Customers can use code generated waiters through `client.wait_until(:waiter_name, params, options)`.


gems/smithy-client/lib/smithy-client/waiters/poller.rb
```diff
@@ -0,0 +1,117 @@
...
+                .gsub(/([a-z\d])([A-Z])/,'\1_\2')
+                .tr("-", "_")
+                .downcase
+        end
+
+        def underscore_jmespath(expression)
+          expression
+            .gsub(' or ', '||')
+            .gsub(/(?<![`'])\b\w+\b(?![`'])/) { |str| underscore(str) }
+        end
```
richardwang1124: I have these methods here temporarily just to get the runtime trait approach to work but I feel like these would be better living in a util module or something instead. Also the specific logic may need to be modified to account for irregular inflections like in V3.

gems/smithy-client/lib/smithy-client/waiters/poller.rb
```diff
@@ -0,0 +1,117 @@
...
+
+        def call(client, params)
+          @input = params
+          begin
+            resp = client.send(@operation_name, params)
+          rescue StandardError => e
+            error = e
+          end
+          resp_or_error = resp || error
```
richardwang1124: I should be able to assume that either one or the other will be guaranteed to be non nil?

gems/smithy/lib/smithy/generators/client.rb
```diff
@@ -50,6 +50,7 @@ def source_files
           e.yield "lib/#{@gem_name}/endpoint_parameters.rb", Views::Client::EndpointParameters.new(@plan).render
           e.yield "lib/#{@gem_name}/endpoint_provider.rb", Views::Client::EndpointProvider.new(@plan).render
           e.yield "lib/#{@gem_name}/paginators.rb", Views::Client::Paginators.new(@plan).render
+          e.yield "lib/#{@gem_name}/waiters.rb", Views::Client::Waiters.new(@plan).render
```
richardwang1124: Does the order matter for when we render waiters?

gems/smithy/lib/smithy/views/client/module.rb
```diff
@@ -43,7 +43,7 @@ def relative_requires
           return %i[customizations types schema] if @plan.type == :schema
 
           # paginators must come before schemas
-          %w[types paginators schema auth_parameters auth_resolver client customizations errors endpoint_parameters
+          %w[types paginators waiters schema auth_parameters auth_resolver client customizations errors endpoint_parameters
```
richardwang1124: Does the requiring order matter for waiters?

gems/smithy/spec/interfaces/client/waiters_spec.rb
```diff
@@ -0,0 +1,933 @@
...
+                {
+                  'state' => 'success',
+                  'matcher' => {
+                    'success' => true
+                  }
+                }
+              ]
+            }
+          }
+          client.config.service.operations[:delete_widget].traits['smithy.waiters#waitable'].merge!(custom_waiter)
```
richardwang1124: Should there be a method or something to make modifying traits easier for the customer? Also since the customer is modifying the trait at runtime, no Smithy validations are ran. How can we ensure that the traits they add are valid?

gems/smithy/spec/interfaces/client/waiters_spec.rb
```diff
@@ -0,0 +1,933 @@
...
+            min_delay: 3,
+            max_delay: 4
+          }
+          output = {}
+          2.times do
+            allow(client).to receive(:get_widget).and_return(output)
+          end
+          expect_any_instance_of(waiter).to receive(:delay).and_wrap_original do |m, *args|
+            delay = m.call(*args)
+            expect(delay).to equal(5)
```
richardwang1124: This test case is an example of the question I had in my doc regarding waiter retries. Because the remaining time (5) minus the delay (value between min delay of 3 and max delay of 4) is less than the min delay, the final delay value is set to be equal to the remaining time, which is 5. However, this exceeds the max delay. Is this expected behavior?

gems/smithy-client/lib/smithy-client/waiters/errors.rb
```diff
@@ -0,0 +1,54 @@
...
+  module Client
+    module Waiters
+      module Errors
+
+        # Raised when a waiter detects a condition where the waiter can never
+        # succeed.
+        class WaiterFailed < StandardError; end
+
+        class FailureStateError < WaiterFailed
+          MSG = "stopped waiting, encountered a failure state"
```
mullermp: No need to make these constants.

gems/smithy-client/lib/smithy-client/waiters/poller.rb
mullermp: Yes, but this should be output_or_error I think. (Operations return an output but in v3 we called that a response. I'm not sure if I like that and maybe we can change it but let's continue to be consistent). Raise response errors should maybe be disabled instead of having this begin rescue block. We can think about that.

gems/smithy-client/lib/smithy-client/waiters/poller.rb
```diff
@@ -0,0 +1,117 @@
...
+
+          actual = JMESPath.search(underscore_jmespath(path_matcher['path']), resp)
+          is_equal?(actual, path_matcher['expected'], path_matcher['comparator'])
+        end
+
+        def matches_inputOutput?(path_matcher, resp, error)
+          return false unless error.nil? && @input
+
+          data = {
+            input: @input, ### Where do we get this?
```
mullermp: What do you mean? It looks to be set on the initial call. Is this what we do in v3?

gems/smithy-client/lib/smithy-client/waiters/poller.rb
```diff
@@ -0,0 +1,117 @@
...
+        def is_equal?(actual, expected, comparator)
+          case comparator
+          when 'stringEquals'
+            return actual == expected
+          when 'booleanEquals'
+            return actual.to_s == expected
+          when 'allStringEquals'
+            return false if actual.nil? || actual.empty?
+
+            actual.each do |value|
```
mullermp: You can probably use .all? for code like this.

gems/smithy-client/lib/smithy-client/waiters/poller.rb
mullermp: Yes, that's fine for now. Leave these here and we can think about which approach to take. A util module would make sense if we decide that way.

gems/smithy-client/lib/smithy-client/waiters/waiter.rb
```diff
@@ -0,0 +1,69 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    module Waiters
+      class Waiter
+        def initialize(options = {})
+          unless options[:max_wait_time].is_a?(Integer)
+            raise ArgumentError, 'Waiter must be initialized with `:max_wait_time`'
```
mullermp: Nit: change this to something like

:max_wait_time must be an integer

gems/smithy-client/lib/smithy-client/waiters/waiter.rb
```diff
@@ -0,0 +1,69 @@
...
+          end
+
+          @max_wait_time = options[:max_wait_time]
+          @remaining_time = @max_wait_time
+          @min_delay = options[:min_delay]
+          @max_delay = options[:max_delay]
+          if @max_delay < 1
```
mullermp: A clean way to do this is to make a method called max_delay(options) that validates and returns the value, and you can directly assign to instance variable.

gems/smithy/lib/smithy/generators/client.rb
mullermp: No, we are doing alphabetical.

gems/smithy/lib/smithy/templates/client/client.erb
```diff
@@ -36,6 +36,21 @@ module <%= module_name %>
     end
 
 <% end -%>
+<% if waiters? -%>
```
mullermp: I don't think we should bother checking if any exist or not. I think we always want the wait until method on the client so that all of our clients are duck typed and respond to it. Even if that means there are no available waiters, that's fine. We may end up supporting custom waiters either by trait or some way to inject some, so we want the method to exist. We could even consider putting this on base but I think the documentation story of that is not great.

gems/smithy/lib/smithy/templates/client/client.erb
```diff
@@ -51,6 +66,72 @@ module <%= module_name %>
...
+    end
+
+    def waiters
+<% waiters.each do |line| -%>
+    <%= line %>
+<% end -%>
+    end
+<% end -%>
+
+    # runtime approach
```
mullermp: If we go with this approach, all of this logic should be in smithy client, not generated here.

gems/smithy/lib/smithy/views/client/client.rb
```diff
@@ -60,6 +60,22 @@ def protocols
           @protocols ||= @plan.welds.map(&:protocols).reduce({}, :merge)
         end
 
+        def waiters
+          waiters = Views::Client::Waiters.new(@plan).waiters
+          code = ['  {']
```
mullermp: I've been calling this lines for consistency. Also I don't think you should indent here in the code - can indenting be in the view?

gems/smithy/lib/smithy/views/client/module.rb
mullermp: Unless it depends on another else code generated, no. It can probably go alphabetical. If it fails then you'll know.

gems/smithy/lib/smithy/views/client/waiters.rb
```diff
@@ -0,0 +1,70 @@
...
+        class Waiter
+          def initialize(operation, name, waiter)
+            @operation_name = operation
+            @name = name
+            @documentation = waiter['documentation']
+            @acceptors = formatted_acceptors(waiter['acceptors'])
+            @min_delay = waiter['minDelay'] || 2
+            @max_delay = waiter['maxDelay'] || 120
+            @deprecated = waiter['deprecated']
+            @tags = waiter['tags']
```
mullermp: I think we can ignore tags entirely. I don't see a use for them even in documentation.

gems/smithy/lib/smithy/views/client/waiters.rb
```diff
@@ -0,0 +1,70 @@
...
+          Model::ServiceIndex
+            .new(@model)
+            .operations_for(@plan.service)
+            .map do |operation_id, operation|
+              waiters_from_trait = waitable_trait(operation)
+              next if waiters_from_trait.empty?
+
+              operation_name = Model::Shape.name(operation_id).underscore
+
+              waiters_from_trait.map do |waiter_name, waiter|
```
mullermp: This might be cleaner if it's not entirely chained together. Perhaps get the set of operations, and iterate them, then iterate waiters, and add to an array.

gems/smithy/spec/fixtures/wait_service/model.smithy
```diff
@@ -0,0 +1,386 @@
...
+
+namespace smithy.ruby.tests
+
+use smithy.waiters#waitable
+
+// A service which has a GET operation with waiters defined upon it.
+// The acceptor in each waiter serves as subject for unit testing,
+// to ensure that the logic in code-generated acceptors works as
+// expected.
+service WaitService {
```
mullermp: You'll want to verify this covers all the possible edge cases.

model/weather.smithy
```diff
@@ -35,6 +36,30 @@ resource Forecast {
 string CityId
 
 @readonly
+@waitable(
```
mullermp: I'm assuming you'll revert the weather changes right?

gems/smithy/spec/interfaces/client/waiters_spec.rb
```diff
@@ -0,0 +1,933 @@
+# frozen_string_literal: true
+
+require_relative '../../spec_helper'
+
+describe 'Client: Waiters' do
```
mullermp, line 5: I think this test file should simply test that the wait_until method exists and works, but not necessarily dive into the behavior of the waiters. I think that's best tested for a spec file in smithy client since we are technically testing the abstract poller/waiter classes. You would have to break up the test model to be something like how other plugins are tested there - with each test set adding their own waiter acceptor as needed.

mullermp (commented):
Good start!

gems/smithy-client/lib/smithy-client/waiters/poller.rb
richardwang1124: I think this was a comment I left earlier for myself to figure out where to get the input since in your Smithy-Ruby Java waiters PR you used middleware and I wasn't sure what that was/if that was necessary for Decaf. Also in V3 there's no input output matcher, only a path matcher.

gems/smithy-client/lib/smithy-client/waiters/waiter.rb
richardwang1124: Does that message imply that `:max_wait_time` must be defined? I wanted to convey that it's a required parameter.

gems/smithy/lib/smithy/templates/client/client.erb
richardwang1124: Should this logic go in the abstract Waiter class? It feels a little weird putting it there since this logic is more like at the client level and not at a specific waiter level.

gems/smithy-client/lib/smithy-client/waiters/waiter.rb
mullermp: Yeah I think so. nil is not an integer (if not provided). You can make it something like:
```
expected :max_wait_time to be an integer, got: #{value}
```

gems/smithy/lib/smithy/templates/client/client.erb
mullermp: Yeah, if we took the runtime approach, we would probably have a Waiters.new(operations) class that made a bunch of Waiter, or something like that. Some core logic would do that unpacking of operation traits.

model/weather.smithy
richardwang1124: Yep once I'm done testing and making edits I'll revert these changes.

gems/smithy/lib/smithy/views/client/client.rb
richardwang1124: I think you used code for paginators which is why I also used code here. I can change these to lines though if that's what's consistent?

gems/smithy-client/lib/smithy-client/waiters/errors.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    module Waiters
+      module Errors
```
mullermp: Not sure yet but I think we should stuff these under smithy client errors.

gems/smithy-client/lib/smithy-client/waiters/errors.rb
```diff
@@ -0,0 +1,44 @@
...
+
+        # Raised when a waiter detects a condition where the waiter can never
+        # succeed.
+        class WaiterFailed < StandardError; end
+
+        class FailureStateError < WaiterFailed
+          def initialize(error)
+            msg = "stopped waiting, encountered a failure state: %s"
```
mullermp: Use string interpolation with #{}

gems/smithy-client/lib/smithy-client/waiters/poller.rb
```diff
@@ -0,0 +1,94 @@
...
+      class Poller
+        def initialize(options = {})
+          @operation_name = options[:operation_name]
+          @acceptors = options[:acceptors]
+        end
+
+        def call(client, params)
+          @input = params
+          begin
+            output = client.send(@operation_name, params)
```
mullermp: We should consider supporting raise response errors as a request option. Then you don't need to rescue here.

gems/smithy-client/lib/smithy-client/waiters/poller.rb
```diff
@@ -0,0 +1,94 @@
...
+        end
+
+        def matches_output?(path_matcher, output, error)
+          return false unless error.nil?
+
+          actual = JMESPath.search(path_matcher['path'], output)
+          is_equal?(actual, path_matcher['expected'], path_matcher['comparator'])
+        end
+
+        def matches_inputOutput?(path_matcher, output, error)
```
mullermp: Does RoboCop not complain? I assume you want inline disables for these.

gems/smithy-client/lib/smithy-client/waiters/poller.rb
```diff
@@ -0,0 +1,94 @@
...
+            !output.nil?
+          else
+            !error.nil?
+          end
+        end
+
+        def matches_errorType?(path_matcher, output, error)
+          return false unless output.nil?
+
+          err = path_matcher.split('#').last.split('#').first
```
mullermp: Is this an absolute shape id? We should consider preprocessing this.

gems/smithy-client/lib/smithy-client/waiters/poller.rb
```diff
@@ -0,0 +1,94 @@
...
+          error.class.to_s.include?(err)
```
mullermp: Wouldn't they be equal here?

gems/smithy-client/lib/smithy-client/waiters/waiter.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    module Waiters
+      class Waiter
+        def initialize(options = {})
+          unless options[:max_wait_time].is_a?(Integer)
```
mullermp: You can wrap this in a method like the others?

gems/smithy/lib/smithy/templates/client/client.erb
```diff
@@ -36,6 +36,11 @@ module <%= module_name %>
     end
 
 <% end -%>
+    def wait_until(waiter_name, params = {}, options = {})
```
mullermp, line 101: We will need documentation.

gems/smithy/lib/smithy/templates/client/client.erb
```diff
@@ -52,6 +57,25 @@ module <%= module_name %>
...
+      waiter_class = waiters[waiter_name]
+      if waiter_class
+        waiter_class.new(options.merge(client: self))
+      else
+        raise Smithy::Client::Waiters::Errors::NoSuchWaiterError.new(waiter_name, waiters.keys)
+      end
+    end
+
+    def waiters
+<% if waiters? -%>
```
mullermp: Rather than check this, you should just have the view handle the empty case I think.

gems/smithy/lib/smithy/templates/client/waiters.erb
```diff
@@ -0,0 +1,40 @@
...
+
+  # @api private
+  module Waiters
+<% waiters.each do |waiter| -%>
+<% if waiter.deprecated -%>
+    # @deprecated
+<% end -%>
+    # @api private
+    #
+    # <%= waiter.documentation %>
```
mullermp: You'll need to loop over this. Check out other examples in the package. The documentation may have newlines in it. I believe the other methods are called docstrings and split on newline. Just make it consistent.

gems/smithy/lib/smithy/views/client/client.rb
```diff
@@ -60,6 +60,22 @@ def protocols
           @protocols ||= @plan.welds.map(&:protocols).reduce({}, :merge)
         end
 
+        def waiters
+          waiters = Views::Client::Waiters.new(@plan).waiters
+          lines = ['{']
+          waiters.each_with_index do |waiter, i|
+            line = '  ' + waiter.name.underscore + ': Waiters::' + waiter.name + ','
```
mullermp: Use interpolation. Do you have RoboCop set up locally? I'm surprised it didn't complain here.

gems/smithy/lib/smithy/views/client/client.rb
```diff
@@ -60,6 +60,22 @@ def protocols
...
+            line.chomp!(',') if i == waiters.length - 1
```
mullermp: Check out the operation examples code. You shouldn't need to keep track of index I think. You can just run the loop and always chomp comma.

gems/smithy/lib/smithy/views/client/waiters.rb
```diff
@@ -0,0 +1,73 @@
...
+        def module_name
+          @plan.module_name
+        end
+
+        def waiters
+          waiters = []
+          operations = Model::ServiceIndex.new(@model).operations_for(@plan.service)
+          operations.each do |operation_id, operation|
+            waiters_from_trait = waitable_trait(operation)
+            unless waiters_from_trait.empty?
```
mullermp: Use a guard clause. Next if it's empty.

gems/smithy/spec/interfaces/client/waiters_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+  let(:input) { { string_property: 'input_string' } }
+  let(:client) { Wait_Service::Client.new(stub_responses: true) }
+  let(:unexpected_error) { Smithy::Client::Waiters::Errors::UnexpectedError }
+  let(:no_such_waiter_error) { Smithy::Client::Waiters::Errors::NoSuchWaiterError }
+
+  ['generated client gem', 'generated client from source code'].each do |context|
+    context context do
+      include_context context, 'Wait_Service'
```
mullermp: Nit: WaiterService

gems/smithy/spec/interfaces/client/waiters_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+      it 'returns when successful' do
+        output = {}
+        expect(client).to receive(:get_operation).and_return(output)
+        expect {
```
mullermp: Do .. end for multiline blocks
Or you can put it on one line and start the next line with .to

gems/smithy/spec/interfaces/client/waiters_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+        }.to raise_error(unexpected_error)
+      end
+
+      it 'raises an error for nonexistent waiters' do
+        expect {
+          client.wait_until(:nonexistent_waiter, input, max_wait_time: 60)
+        }.to raise_error(no_such_waiter_error)
+      end
+
+      it 'does not allow custom waiters' do
```
mullermp: Interesting. While true, I guess this documents that? I don't think it's needed but you can keep it if you want. I did not add one for paginators.

gems/smithy/spec/interfaces/client/waiters_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      end
+
+      it 'raises an error when unsuccessful' do
+        expect(client).to receive(:get_operation).and_raise(StandardError)
+        expect {
+          client.wait_until(:success_matcher, input, max_wait_time: 60)
+        }.to raise_error(unexpected_error)
+      end
+
+      it 'raises an error for nonexistent waiters' do
```
mullermp, line 24: I think this test file really just needs 2 or 3 tests. One for trying to get a non existent waiter (failure), a "good case" that doesn't raise an error, and one that verifies the return type. Is there any requirement on what waiters should return? What happens in v3? I think we possibly want to always return nil or like kernel.sleep we should return the time we waited.

gems/smithy/spec/interfaces/client/waiters_spec.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+require_relative '../../spec_helper'
+
+describe 'Client: Waiters' do
+  let(:input) { { string_property: 'input_string' } }
+  let(:client) { Wait_Service::Client.new(stub_responses: true) }
+  let(:unexpected_error) { Smithy::Client::Waiters::Errors::UnexpectedError }
```
mullermp: Nit: unless you're using these in more than one place, it's more readable to stick into the test itself.

gems/smithy-client/spec/smithy-client/waiters_spec.rb
```diff
@@ -0,0 +1,1661 @@
+# frozen_string_literal: true
+
+require_relative '../spec_helper'
+
+module Smithy
+  module Client
+    module Waiters
+      describe Waiters do
+        let(:shapes) do
```
mullermp: I was hoping this would be slimmed down. We don't need to define all of the possibilities up front and we certainly don't need documentation traits etc. This model would be difficult to maintain.

I think we would want context blocks and make small services that have each concrete example.

Let's talk offline.

gems/smithy-client/lib/smithy-client/waiters/errors.rb
jterapin:
```suggestion
            msg = "stopped waiting, encountered a failure state: #{error}"
```

gems/smithy-client/lib/smithy-client/waiters/poller.rb
```diff
@@ -0,0 +1,94 @@
...
+          end
+          if error
+            'error'
+          else
+            'retry'
+          end
+        end
+
+        def acceptor_matches?(matcher, output, error)
+          matcher_type = matcher.keys[0]
```
jterapin: I think `matcher.keys.first` will work here.

gems/smithy-client/lib/smithy-client/waiters/waiter.rb
```diff
@@ -0,0 +1,70 @@
...
+          delay
+        end
+
+        def poll(client, params)
+          attempts = 0
+          loop do
+            output_or_error, status = @poller.call(client, params)
+            attempts += 1
+
+            case status.to_sym
```
jterapin: Can we just ensure that status is symbol always?

gems/smithy-client/spec/smithy-client/waiters_spec.rb
jterapin: These shapes is a lot to scroll through - might be better to put these in a separate json file?

jterapin (commented):
Nice! Great work so far - mini-review since you are working through changes and etc.

gems/smithy-client/lib/smithy-client/waiters/errors.rb
richardwang1124: Are all errors currently defined there? I see an error for paginators in there but I also notice that errors are in other places too, like networking errors in the `networking_error.rb` file or `CapacityNotAvailableError` in `client_rate_limiter.rb`.

gems/smithy-client/lib/smithy-client/waiters/poller.rb
richardwang1124: How would I do that? I see in V3 we remove the RaiseResponseErrors handler and I tried that here but an error is still raised.

gems/smithy-client/lib/smithy-client/waiters/poller.rb
richardwang1124: This can be an absolute or relative shape id I think.

gems/smithy-client/lib/smithy-client/waiters/poller.rb
richardwang1124: At least when I was testing the actual error raised will include namespaces too (e.g. actual error is `SampleClient1::Errors::MyError` vs expected error `MyError` or `smithy.ruby.tests#MyError`).

gems/smithy/spec/interfaces/client/waiters_spec.rb
richardwang1124: I actually had this same question earlier but I assumed the return type would either be the response or there would be an error raised based on the V3 behavior I saw. However, I don't think the return type is mentioned in the SEP or in the Smithy docs. Do any customers expect a waiter to return something? I would feel that most are using waiters to watch for the state of a resource and not really the response of an operation.

gems/smithy/spec/interfaces/client/waiters_spec.rb
richardwang1124: Depending on whether or not we want to support registering new custom waiters this test case may be modified or removed.

gems/smithy-client/lib/smithy-client/waiters/poller.rb
mullermp: That should not be the case. How did you try to that?

gems/smithy-client/lib/smithy-client/waiters/errors.rb
mullermp: Hmm yeah. We can think on that and re-organize. A few are spread around. You can at minimum probably flatten these and not have a separate Errors module within waiters.

gems/smithy-client/lib/smithy-client/waiters/poller.rb
mullermp: What would happen if we preprocess this to always be a relative shape id?

gems/smithy-client/lib/smithy-client/waiters/poller.rb
mullermp: I think this is the same preprocessing problem right?

gems/smithy/spec/interfaces/client/waiters_spec.rb
mullermp: Right. So I think the return type of wait_until should either raise (if waiter doesn't exist) or return nil after waiting.

gems/smithy-client/lib/smithy-client/waiters/poller.rb
richardwang1124: That would make comparing errors a little more consistent here I think since it would be guaranteed to be relative shape ids. Let me do that.

gems/smithy-client/lib/smithy-client/waiters/poller.rb
richardwang1124: Hm preprocessing would ensure the expected error would always be `MyError` instead of `smithy.ruby.tests#MyError` but the actual error returned from the request would still be `SampleClient1::Errors::MyError`. Will it be guaranteed that the last part of the actual error after the `::` delimiter will be equal to the expected error if it's supposed to match?

gems/smithy/spec/interfaces/client/waiters_spec.rb
richardwang1124: Interesting, I see in the [V3 documentation](https://github.com/aws/aws-sdk-ruby/blob/version-3/build_tools/aws-sdk-code-generator/templates/client_class.mustache#L199) that the return type for `wait_until` is either `true` when successful or an error is raised, but when I test with `awsrepl` I see that a successful `wait_until` actually just returns the response itself. Should we return the response, `nil`, or `true` when successful?

gems/smithy/lib/smithy/views/client/client.rb
richardwang1124: I need the commas at the end of each line since I'm making a hash. I could leave the comma on the last line, but I thought it looked cleaner without it which is why I chomp the comma for the last line.

gems/smithy-client/lib/smithy-client/waiters/poller.rb
richardwang1124: Hm well I probably did it wrong previously. I'm trying to call something like `client.remove_plugin(Smithy::Client::Plugins::RaiseResponseErrors)` but I'm getting a NoMethodError since I think `remove_plugin` should be called on the client class?

gems/smithy-client/lib/smithy-client/waiters.rb
```diff
@@ -0,0 +1,4 @@
+# frozen_string_literal: true
+
+require_relative 'waiters/poller'
```
mullermp: I would just place these in smithy-client.rb. This file is not big enough to justify loading.

gems/smithy-client/lib/smithy-client/errors.rb
```diff
@@ -3,6 +3,44 @@
 module Smithy
   module Client
     module Errors
+      # Raised when a waiter detects a condition where the waiter can never
+      # succeed.
+      class WaiterFailed < StandardError; end
```
mullermp: I'm not sure where the conversation went, but I think we want these in Smithy::Client::Waiters (but not have an Errors module inside there - just define the errors classes in the waiter namespace in the file where they are raised).

gems/smithy-client/lib/smithy-client/waiters/poller.rb
```diff
@@ -0,0 +1,100 @@
...
+          [output, status.to_sym]
+        end
+
+        private
+
+        def evaluate_acceptors(output)
+          @acceptors.each do |acceptor|
+            return acceptor['state'] if acceptor_matches?(acceptor['matcher'], output)
+          end
+          if output.error.nil?
```
mullermp: guard clause

gems/smithy-client/lib/smithy-client/waiters/poller.rb
```diff
@@ -0,0 +1,100 @@
...
+        end
+
+        def matches_output?(path_matcher, output)
+          return false if output.data.nil?
+
+          actual = JMESPath.search(path_matcher['path'], output.data)
+          equal?(actual, path_matcher['expected'], path_matcher['comparator'])
+        end
+
+        def matches_inputOutput?(path_matcher, output) # rubocop:disable Naming/MethodName
```
mullermp: For these comments, you can disable once above this method and enable once again after the last method.

gems/smithy-client/lib/smithy-client/waiters/poller.rb
```diff
@@ -0,0 +1,100 @@
...
+          data = {
+            input: @input,
+            output: output.data
+          }
+          actual = JMESPath.search(path_matcher['path'], data)
+          equal?(actual, path_matcher['expected'], path_matcher['comparator'])
+        end
+
+        def matches_success?(path_matcher, output)
+          if path_matcher == true
```
mullermp: guard clause

gems/smithy-client/lib/smithy-client/waiters/waiter.rb
```diff
@@ -0,0 +1,73 @@
...
+
+        def poll(client, params)
+          attempts = 0
+          loop do
+            output, status = @poller.call(client, params)
+            attempts += 1
+
+            case status
+            when :success then return
+            when :failure then raise Smithy::Client::Errors::FailureStateError, output.error
```
mullermp: You don't have to use the fully qualified error name here - you can start with Errors:: or since you should move the error definitions here IMO, you can use FailureStateError directly.

gems/smithy/lib/smithy/templates/client/client.erb
```diff
@@ -36,6 +36,73 @@ module <%= module_name %>
...
+    #       max_wait_time: 25,
+    #       min_delay: 2,
+    #       max_delay: 10
+    #     })
+    #
+    # ## Handling Errors
+    #
+    # When a waiter is unsuccessful, it will raise an error.
+    # All the failure errors extend from
+    # {Smithy::Client::Waiters::Errors:WaiterFailed}.
```
mullermp: Missing colon here and below, but don't forget to update this.

gems/smithy/lib/smithy/templates/client/client.erb
```diff
@@ -36,6 +36,73 @@ module <%= module_name %>
...
+    # ## Configuration
+    #
+    # You must configure the maximum amount of time in seconds a
+    # waiter should wait for. You may also configure the minimum
+    # and maximum amount of time in seconds to delay between
+    # retries. You can pass these configuration as the final
+    # arguments hash.
+    #
+    #     # poll for a maximum of 25 seconds
+    #     client.wait_until(waiter_name, params, {
```
mullermp: I'd make up an example here, and make waiter_name a symbol as the documentation suggests. I'd also define params. I've been using Weather examples (see pagination examples, etc). You can probably just make up a fake waiter here, I'd say something like `:forecast_exists` and maybe params you can do something like `{ forecast_id: '1' }`. You can also use it in your error examples below.

gems/smithy/lib/smithy/templates/client/client.erb
```diff
@@ -36,6 +36,73 @@ module <%= module_name %>
...
+    # All the failure errors extend from
+    # {Smithy::Client::Waiters::Errors:WaiterFailed}.
+    #
+    #     begin
+    #       client.wait_until(...)
+    #     rescue Smithy::Client::Errors:WaiterFailed
+    #       # resource did not enter the desired state in time
+    #     end
+    #
+    # @raise [Errors::FailureStateError] Raised when the waiter terminates
```
mullermp: nit - raises and returns go after params.

gems/smithy/lib/smithy/templates/client/client.erb
```diff
@@ -36,6 +36,73 @@ module <%= module_name %>
...
+    #
+    # @return [nil] Returns `nil` if the waiter was successful.
+    # @param [Symbol] waiter_name
+    # @param [Hash] params ({})
+    # @param [Hash] options ({})
+    # @option options [Integer] :max_wait_time
+    # @option options [Integer] :min_delay
+    # @option options [Integer] :max_delay
+    def wait_until(waiter_name, params = {}, options = {})
+      w = waiter(waiter_name, options)
```
mullermp: No need to make w a variable imo - waste of allocation.

gems/smithy/lib/smithy/templates/client/client.erb
```diff
@@ -52,6 +119,21 @@ module <%= module_name %>
       Smithy::Client::Input.new(handlers: handlers, context: context)
     end
 
+    def waiter(waiter_name, options = {})
+      waiter_class = waiters[waiter_name]
+      if waiter_class
```
mullermp: guard clause

gems/smithy/lib/smithy/templates/client/client.erb
```diff
@@ -52,6 +119,21 @@ module <%= module_name %>
       Smithy::Client::Input.new(handlers: handlers, context: context)
     end
 
+    def waiter(waiter_name, options = {})
```
mullermp: This method could maybe live in client base?

gems/smithy/lib/smithy/templates/client/waiters.erb
```diff
@@ -0,0 +1,42 @@
...
+# This is generated code!
+
+require 'smithy-client/waiters'
+
+module <%= module_name %>
+
+  # @api private
+  module Waiters
+<% waiters.each do |waiter| -%>
+<% if waiter.deprecated -%>
```
mullermp: boolean methods should end in `?` by convention. Does deprecated have a since property?

gems/smithy/lib/smithy/views/client/client.rb
mullermp: I understand that, but you don't need to keep track of the index.  Just do something like this after your loop:
```
          lines.last.chomp!(',') if lines.last.end_with?(',')
```

look at operation_examples.rb

gems/smithy/lib/smithy/views/client/waiters.rb
```diff
@@ -0,0 +1,79 @@
...
+              quote_strings: true,
+              indent: '            '
+            ).format(acceptors: acceptors)
+          end
+
+          def preprocess_acceptor(acceptor)
+            if (matcher = acceptor['matcher']['output'] || acceptor['matcher']['inputOutput'])
+              matcher['path'] = Util::Underscore.underscore_jmespath(matcher['path'])
+            elsif (error_type = acceptor['matcher']['errorType'])
+              acceptor['matcher']['errorType'] = error_type.split('#').last.split('#').first
```
mullermp: What is this trying to find and split by? Look at `Model::Shape` - I think you can use this to extract the name.

What I was trying to say before, is you can actually get the error module at runtime with client.class.errors_module, then check if the error type is a class defined in errors_module, or you can compare strings as you've done.

gems/smithy/lib/smithy/views/client/waiters.rb
```diff
@@ -0,0 +1,79 @@
...
+        def waitable_trait(operation)
+          operation.fetch('traits', {}).fetch('smithy.waiters#waitable', {})
+        end
+
+        # @api private
+        class Waiter
+          def initialize(operation, name, waiter)
+            @operation_name = operation
+            @name = name
+            @documentation = waiter['documentation']
```
mullermp: You can do `waiter.fetch('documentation', '')` then your docstrings method can just always split on "\n"

gems/smithy/lib/smithy/views/client/waiters.rb
```diff
@@ -0,0 +1,79 @@
...
+
+        private
+
+        def waitable_trait(operation)
+          operation.fetch('traits', {}).fetch('smithy.waiters#waitable', {})
+        end
+
+        # @api private
+        class Waiter
+          def initialize(operation, name, waiter)
```
mullermp: nit - operation_name

gems/smithy/lib/smithy/views/client/waiters.rb
```diff
@@ -0,0 +1,79 @@
...
+          end
+
+          def formatted_acceptors(acceptors)
+            acceptors.each { |acceptor| preprocess_acceptor(acceptor) }
+
+            Util::HashFormatter.new(
+              wrap: false,
+              inline: false,
+              quote_strings: true,
+              indent: '            '
```
mullermp: This is very context dependent. I would actually just not indent here, call .split("\n"), then render acceptors in place just like you do docstrings. Then it handles any context.

gems/smithy/spec/fixtures/waiter_service/model.json
```diff
@@ -0,0 +1,367 @@
+{
+    "smithy": "2.0",
+    "shapes": {
+        "smithy.protocols#StringList": {
```
mullermp: Let's figure out how to strip these. If you'd like to tackle this, we need to smithy build instead of smithy ast, then copy out the built json from a build dir rather than copy stdout.

gems/smithy/spec/interfaces/client/waiters_spec.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+require_relative '../../spec_helper'
+
+describe 'Client: Waiters' do
+  let(:input) { { string_property: 'input_string' } }
+  let(:client) { WaiterService::Client.new(stub_responses: true) }
```
mullermp: nit - define these instead `context context` - they are otherwise invalid at this scope

gems/smithy/spec/interfaces/client/waiters_spec.rb
mullermp: I'd say just remove it - it's cruft that can break and relies on modifying api private information.

gems/smithy/spec/interfaces/client/waiters_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+  let(:no_such_waiter_error) { Smithy::Client::Errors::NoSuchWaiterError }
```
mullermp: I'd eliminate this - it's used once - just put it in the test. Same with input honestly. It's used twice but it's simple enough that you don't need to scope it to a variable. I think in general, having simple tests not rely too much on variables is good practice, because they can be modified as one unit. I use them when they are complex and used many times, or for setup.

gems/smithy-client/spec/smithy-client/waiters_spec.rb
```diff
@@ -0,0 +1,1364 @@
...
+                  'target' => 'smithy.api#String'
+                }
+              },
+              'traits' => {
+                'smithy.api#error' => 'client'
+              }
+            }
+          }
+        end
+        let(:sample_client) { ClientHelper.sample_client(shapes: shapes) }
```
mullermp: nit: add a few spacings between all of this, it's kind of an eye sore..

gems/smithy-client/spec/smithy-client/waiters_spec.rb
```diff
@@ -0,0 +1,1364 @@
...
+          client_class.add_plugin(Smithy::Client::Plugins::StubResponses)
+          client_class
+        end
+        let(:client) { client_class.new(stub_responses: true) }
+        let(:input) { { string_property: 'input_string' } }
+        let(:waiter) { Waiter }
+        let(:poller) { Poller }
+        let(:my_error) { sample_client::Errors::MyError.new(nil, nil) }
+        let(:unexpected_error) { Smithy::Client::Errors::UnexpectedError }
+        let(:failure_state_error) { Smithy::Client::Errors::FailureStateError }
```
mullermp: I think a lot of these are just not necessary - use them in place. They are already constants, so we are assigning constants to variables that are also used as constants. You also don't need the fully qualified error namespace - you are already scoped inside Smithy::Client so just FailureStateError works (After you move it), and that's short enough. They also make the test more readable.

gems/smithy-client/spec/smithy-client/waiters_spec.rb
```diff
@@ -0,0 +1,1364 @@
...
+                  {
+                    'state' => 'success',
+                    'matcher' => {
+                      'success' => false
+                    }
+                  }
+                ]
+              }
+            }
+            client
```
mullermp: You shouldn't need to call `client` here

gems/smithy-client/spec/smithy-client/waiters_spec.rb
```diff
@@ -0,0 +1,1364 @@
...
+          end
+
+          describe '#poll' do
+            it 'delays when status is retry' do
+              output = Smithy::Client::Output.new(data: { string_property: 'expected' })
+              expect_any_instance_of(poller).to receive(:call).and_return([{}, :retry], [output, :success])
```
mullermp: I don't think we should be mocking tests like this - we should be testing behavior. This is very implementation specific. I would not even bother testing anything against Poller as an implementation. If we want to test that delays happen, then we should use stub_responses to stub out maybe two failure calls and a success call, then expect waiter to receive delay twice. Then you can just call client.wait_until, you don't need to expect that it equals nil, because we already tested the return value in the smithy fixture tests, and it didn't raise anything, so we don't really care what it returns here, just that it does what the test describes.

gems/smithy-client/spec/smithy-client/waiters_spec.rb
```diff
@@ -0,0 +1,1364 @@
...
+              expect_any_instance_of(waiter).to receive(:delay).and_return(0)
+              expect(client.wait_until(:success_true_matcher, input, max_wait_time: 60)).to eq(nil)
+            end
+
+            it 'returns nil when status is success' do
```
mullermp: We already tested these right?

gems/smithy-client/spec/smithy-client/waiters_spec.rb
```diff
@@ -0,0 +1,1364 @@
...
+              }
+              expect do
+                client.wait_until(:success_true_matcher, input, options)
+              end.to raise_error(ArgumentError,
+                                 '`:min_delay` must be greater than 0 and less than or equal to `:max_delay`')
+            end
+          end
+        end
+
+        describe 'poller' do
```
mullermp: describe Waiters::Poller

gems/smithy-client/spec/smithy-client/waiters_spec.rb
```diff
@@ -0,0 +1,1364 @@
...
+          describe 'success matcher' do
```
mullermp: Use describe when you are describing a class or method. Use context when you are testing a specific set of things (this is context because you're testing success cases)

gems/smithy-client/spec/smithy-client/waiters_spec.rb
```diff
@@ -0,0 +1,1364 @@
...
+            end
+
+            it 'retries and succeeds when matched' do
+              client.stub_responses(
+                :get_widget,
+                {},
+                {},
+                StandardError
+              )
+              expect_any_instance_of(waiter).to receive(:delay).twice.and_return(0)
```
mullermp: I think it's probably worthwhile to just globally mock `allow(Kernel).to receive(:sleep)` in this file, then you don't need to do this everywhere. You can still test the "delay specific timeperiod" cases because that method sits above Kernel.sleep

gems/smithy-client/spec/smithy-client/waiters_spec.rb
```diff
@@ -0,0 +1,1364 @@
...
+        let(:input) { { string_property: 'input_string' } }
+        let(:waiter) { Waiter }
+        let(:poller) { Poller }
+        let(:my_error) { sample_client::Errors::MyError.new(nil, nil) }
+        let(:unexpected_error) { Smithy::Client::Errors::UnexpectedError }
+        let(:failure_state_error) { Smithy::Client::Errors::FailureStateError }
+        let(:max_wait_time_exceeded_error) { Smithy::Client::Errors::MaxWaitTimeExceededError }
+        let(:no_such_waiter_error) { Smithy::Client::Errors::NoSuchWaiterError }
+
+        describe 'waiter' do
```
mullermp: describe Waiter

gems/smithy-client/spec/smithy-client/waiters_spec.rb
```diff
@@ -0,0 +1,1364 @@
...
+          before(:each) do
+            shapes['smithy.ruby.tests#GetWidget']['traits']['smithy.waiters#waitable'] = {
```
mullermp: What I was trying to say in slack is that you can actually define in GetWidget

```
'traits' => { 'smithy.waiters#waitable' => matchers }
```

and then in contexts, instead of before(:each) blocks, you can just do

let(:matchers) { .. }

gems/smithy-client/spec/smithy-client/waiters_spec.rb
```diff
@@ -0,0 +1,1364 @@
...
+            end
+          end
+        end
+
+        describe 'poller' do
+          describe 'success matcher' do
+            before(:each) do
+              shapes['smithy.ruby.tests#GetWidget']['traits']['smithy.waiters#waitable'] = {
+                'SuccessTrueMatcher' => {
+                  'acceptors' => [
```
mullermp: In general, long, simple line JSONs should be shortened. You can put state, matcher, and success on one line

gems/smithy-client/spec/smithy-client/waiters_spec.rb
```diff
@@ -0,0 +1,1364 @@
...
+
+              it 'fails when output property is nil' do
+                client.stub_responses(:get_widget, {})
+                expect do
+                  client.wait_until(:output_string_array_any_property_matcher, input, max_wait_time: 0)
+                end.to raise_error(max_wait_time_exceeded_error)
+              end
+            end
+
+            context 'flatten' do
```
mullermp: Is flatten a real matcher? Let's use the terminology from https://smithy.io/2.0/additional-specs/waiters.html

gems/smithy-client/spec/smithy-client/waiters_spec.rb
```diff
@@ -0,0 +1,1364 @@
...
+                    ]
+                  }
+                }
+              end
+
+              it 'succeeds when matched' do
+                output = {
+                  children: [
+                    {
+                      grandchildren: [
```
mullermp: Condense please

gems/smithy-client/spec/smithy-client/waiters_spec.rb
```diff
@@ -0,0 +1,1364 @@
...
+                expect do
+                  client.wait_until(:contains_field_matcher, input, max_wait_time: 0)
+                end.to raise_error(max_wait_time_exceeded_error)
+              end
+            end
+
+            context 'and inequality' do
+              before(:each) do
+                shapes['smithy.ruby.tests#GetWidget']['traits']['smithy.waiters#waitable'] = {
+                  'AndInequalityMatcher' => {
```
mullermp: This is a fundamental difference I think between kotlin's approach and ruby - I don't know if we care to test things like flatten/and inequality, etc - These are really testing JMESPath implementations. I don't think Kotlin has a runtime JMESPath implementation like we do. I think we care to test the basic waiter matcher types (like booleanEquals) and we can rely on the fact that JMESPath handles "&&" expressions without having to test it here. So I think you can cut a good chunk of the tests in this file.

mullermp (commented):
Nice! Getting better. Sorry for the barrage of comments.

gems/smithy/lib/smithy/templates/client/waiters.erb
richardwang1124: This is just the `deprecated` member, not a method. Is it worth creating a method for?

I also don't see anything about a since property in the Smithy docs.

gems/smithy/spec/fixtures/waiter_service/model.json
richardwang1124: Maybe we can create a separate PR to address this?

gems/smithy-client/lib/smithy-client/base.rb
```diff
@@ -41,6 +41,13 @@ def inspect
         "#<#{self.class.name || 'Smithy::Client::Base'}>"
       end
 
+      def waiter(waiter_name, options = {})
```
mullermp: I believe this method should be private. You can also shorten to `Waiters::NoSuchWaiterError.new`. I would also invert this guard clause to raise unless waiter_class.

gems/smithy-client/lib/smithy-client/waiters/poller.rb
```diff
@@ -0,0 +1,94 @@
...
+        def initialize(options = {})
+          @operation_name = options[:operation_name]
+          @acceptors = options[:acceptors]
+        end
+
+        def call(client, params)
+          @input = params
+          # TODO: make build_input public and update this line
+          input = client.send(:build_input, @operation_name, params)
+          input.handlers.remove(Smithy::Client::Plugins::RaiseResponseErrors::Handler)
```
mullermp: Please shorten namespaces throughout. If you are under module Smithy and module Client, you do not need to reference it again.

gems/smithy/lib/smithy/templates/client/waiters.erb
mullermp: Up to you - you can always have a method `deprecated?` instead of attribute accessor deprecated - since you never actually need to retrieve any nested value of deprecated, it's simply a boolean. That's more conventional.

gems/smithy/lib/smithy/views/client/client.rb
```diff
@@ -60,6 +60,20 @@ def protocols
           @protocols ||= @plan.welds.map(&:protocols).reduce({}, :merge)
         end
 
+        def waiters
+          waiters = Views::Client::Waiters.new(@plan).waiters
+          return ['{}'] if waiters.empty?
+
+          lines = ['{']
+          waiters.each_with_index do |waiter, _i|
```
mullermp: .each

lines << " ... "

gems/smithy/spec/fixtures/waiter_service/model.smithy
```diff
@@ -0,0 +1,38 @@
+$version: "2"
```
mullermp: nit - name of folder `waiters`

gems/smithy/spec/interfaces/client/waiters_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+      it 'raises waiter failed error when unsuccessful' do
+        client.stub_responses(:get_operation, StandardError)
+        expect do
+          client.wait_until(:success_matcher, { string_property: 'input_string' }, max_wait_time: 60)
+        end.to raise_error(Smithy::Client::Waiters::WaiterFailed)
+      end
+
+      it 'raises an error for nonexistent waiters' do
+        expect do
+          client.wait_until(:nonexistent_waiter, { string_property: 'input_string' }, max_wait_time: 60)
```
mullermp: client.wait_until(:nonexistent_waiter) is sufficient

gems/smithy/spec/interfaces/client/waiters_spec.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+require_relative '../../spec_helper'
+
+describe 'Client: Waiters' do
+  ['generated client gem', 'generated client from source code'].each do |context|
+    context context do
+      include_context context, 'WaiterService'
+
+      let(:input) { { string_property: 'input_string' } }
```
mullermp: Not used?

gems/smithy/spec/interfaces/client/waiters_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+      it 'returns nil when successful' do
+        input = { string_property: 'input_string' }
+        client.stub_responses(:get_operation, { string_property: 'success' })
+        expect(client.wait_until(:success_matcher, input, max_wait_time: 60)).to be(nil)
+      end
+
+      it 'raises waiter failed error when unsuccessful' do
+        client.stub_responses(:get_operation, StandardError)
+        expect do
+          client.wait_until(:success_matcher, { string_property: 'input_string' }, max_wait_time: 60)
```
mullermp: Success matchers don't care about input do they? You can remove stringProperty from the model and just do `client.wait_until(:success_matcher)` assuming you've stubbed the response.

gems/smithy/spec/interfaces/client/waiters_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+    context context do
+      include_context context, 'WaiterService'
+
+      let(:input) { { string_property: 'input_string' } }
+      let(:client) { WaiterService::Client.new(stub_responses: true) }
+      let(:no_such_waiter_error) { Smithy::Client::Waiters::NoSuchWaiterError }
+
+      it 'returns nil when successful' do
+        input = { string_property: 'input_string' }
+        client.stub_responses(:get_operation, { string_property: 'success' })
```
mullermp: By default, stub responses will return a successful response. You can just skip stub responses here and use `client.wait_until(:success_matcher, {}, max_wait_time: 60)`.

gems/smithy-client/spec/smithy-client/waiters_spec.rb
```diff
@@ -0,0 +1,708 @@
...
+                  'acceptors' => [{ 'state' => 'success', 'matcher' => { 'errorType' => 'MyError' } }]
+                },
+                'AbsoluteErrorTypeMatcher' => {
+                  'acceptors' => [{ 'state' => 'success', 'matcher' => { 'errorType' => 'smithy.ruby.tests#MyError' } }]
+                }
+              }
+            end
+
+            it 'succeeds when error matches for relative shape name' do
+              client.stub_responses(:get_widget, my_error)
```
mullermp: You can just stub 'MyError' and not need this variable.

gems/smithy-client/spec/smithy-client/waiters_spec.rb
```diff
@@ -0,0 +1,708 @@
...
+            },
+            'smithy.ruby.tests#BooleanArray' => {
+              'type' => 'list',
+              'member' => { 'target' => 'smithy.api#Boolean' }
+            },
+            'smithy.ruby.tests#StringArray' => {
+              'type' => 'list',
+              'member' => { 'target' => 'smithy.api#String' }
+            },
+            'smithy.ruby.tests#MyError' => {
```
mullermp: nit - Error

gems/smithy-client/spec/smithy-client/waiters_spec.rb
```diff
@@ -0,0 +1,708 @@
...
+
+          describe '#poll' do
+            it 'delays when status is retry' do
+              client.stub_responses(
+                :get_widget,
+                { string_property: 'retry' },
+                { string_property: 'retry' },
+                { string_property: 'success' }
+              )
+              expect_any_instance_of(Waiter).to receive(:delay).exactly(2).times.and_call_original
```
mullermp: Unless you're trying to measure delay seconds, no need to and_call_original

gems/smithy-client/spec/smithy-client/waiters_spec.rb
```diff
@@ -0,0 +1,708 @@
...
+              options = {
+                max_wait_time: 60,
+                min_delay: min_delay,
+                max_delay: max_delay
+              }
+              output = Smithy::Client::Output.new(data: {})
+              expect_any_instance_of(Poller).to receive(:call).and_return(
+                [output, :retry], [output, :retry], [output, :retry], [output, :success]
+              )
+              expect_any_instance_of(Waiter).to receive(:delay).with(1).and_wrap_original do |m, *args|
```
mullermp: You can probably make this more readable by making a loop `1.upto(3) do |attempt|` to make it clear what the delay arguments are.

gems/smithy-client/spec/smithy-client/waiters_spec.rb
```diff
@@ -0,0 +1,708 @@
...
+            it 'generates a random delay between min_delay and max_delay' do
+              min_delay = 5
+              max_delay = 60
+              options = {
+                max_wait_time: 60,
+                min_delay: min_delay,
+                max_delay: max_delay
+              }
+              output = Smithy::Client::Output.new(data: {})
+              expect_any_instance_of(Poller).to receive(:call).and_return(
```
mullermp: Use stub responses instead of mocking poller in these tests.

gems/smithy-client/spec/smithy-client/waiters_spec.rb
```diff
@@ -0,0 +1,708 @@
...
+                delay = m.call(*args)
+                expect(delay.between?(min_delay, max_delay)).to be true
+                0
+              end
+              client.wait_until(:output_string_property_matcher, input, options)
+            end
+
+            it 'sets the delay to remaining time for the last attempt' do
+              remaining_time = 40
+              min_delay = 25
```
mullermp: These aren't used anywhere else right? not really necessary to define/use.

gems/smithy-client/spec/smithy-client/waiters_spec.rb
```diff
@@ -0,0 +1,708 @@
...
+                max_wait_time: remaining_time,
+                min_delay: min_delay,
+                max_delay: max_delay
+              }
+              output = Smithy::Client::Output.new(data: {})
+              expect_any_instance_of(Poller).to receive(:call).and_return([output, :retry], [output, :success])
+              expect_any_instance_of(Waiter).to receive(:delay).and_wrap_original do |m, *args|
+                delay = m.call(*args)
+                expect(delay).to eq(remaining_time)
+                0
```
mullermp: return delay - you already mock out sleeping right?

gems/smithy-client/spec/smithy-client/waiters_spec.rb
```diff
@@ -0,0 +1,708 @@
...
+              client.stub_responses(:get_widget, {})
+              expect do
+                client.wait_until(:output_string_property_matcher, input, max_wait_time: 0)
+              end.to raise_error(MaxWaitTimeExceededError)
+            end
+
+            it 'raises an error when max_wait_time is not provided' do
+              expect do
+                client.wait_until(:output_string_property_matcher, input)
+              end.to raise_error(ArgumentError, 'expected `:max_wait_time` to be an integer, got: ')
```
mullermp: This should say NilClass

gems/smithy-client/lib/smithy-client/waiters/waiter.rb
```diff
@@ -0,0 +1,111 @@
...
+        end
+
+        def wait(client, params)
+          poll(client, params)
+        end
+
+        private
+
+        def max_wait_time(time)
+          raise ArgumentError, "expected `:max_wait_time` to be an integer, got: #{time}" unless time.is_a?(Integer)
```
mullermp: Integer

time.class

gems/smithy-client/spec/smithy-client/waiters_spec.rb
```diff
@@ -0,0 +1,708 @@
...
+            end
+
+            it 'retries and succeeds when matched' do
+              client.stub_responses(
+                :get_widget,
+                {},
+                {},
+                StandardError
+              )
+              expect_any_instance_of(Waiter).to receive(:delay).twice.and_call_original
```
mullermp: Should not need this in any of the below tests

gems/smithy-client/spec/smithy-client/waiters_spec.rb
```diff
@@ -0,0 +1,708 @@
+# frozen_string_literal: true
+
+require_relative '../spec_helper'
+
+module Smithy
+  module Client
+    module Waiters
```
mullermp: I just noticed this, but should not be under module Waiters if you are describing Waiters

gems/smithy-client/spec/smithy-client/waiters_spec.rb
```diff
@@ -0,0 +1,708 @@
...
+
+        let(:client) { client_class.new(stub_responses: true) }
+
+        let(:input) { { string_property: 'input_string' } }
+
+        let(:my_error) { sample_client::Errors::MyError.new(nil, nil) }
+
+        before(:each) { allow_any_instance_of(Waiter).to receive(:sleep) }
+
+        describe 'Waiter' do
```
mullermp: `Waiter` as a class not string

gems/smithy-client/spec/smithy-client/waiters_spec.rb
```diff
@@ -0,0 +1,708 @@
...
+                client.stub_responses(:get_widget, { boolean_property: true })
+                expect do
+                  client.wait_until(:output_boolean_property_matcher, input, max_wait_time: 0)
+                end.to raise_error(MaxWaitTimeExceededError)
+              end
+
+              it 'fails when output property is nil' do
+                client.stub_responses(:get_widget, {})
+                expect do
+                  client.wait_until(:output_boolean_property_matcher, input, max_wait_time: 0)
```
mullermp: Might be useful to DRY this up a little. You could either, extract options into a let() variable, always set max_wait_time to 60, and just pass it into all these for convenience.

Or make a method like:
```
def wait(waiter_name)
  client.wait_until(waiter_name, { string_property: 'input' }, { max_wait_time: 60 })
end

expect { wait(:output_boolean_property_matcher) }
  .to raise_error(MaxWaitTimeExceededError)
```

gems/smithy-client/spec/smithy-client/waiters_spec.rb
```diff
@@ -0,0 +1,708 @@
...
+
+            it 'fails when error does not match' do
+              client.stub_responses(:get_widget, StandardError)
+              expect do
+                client.wait_until(:error_type_matcher, input, max_wait_time: 60)
+              end.to raise_error(UnexpectedError)
+            end
+          end
+
+          context 'output matcher' do
```
mullermp: no need to have this context imo if you have no shared context in between

gems/smithy-client/spec/smithy-client/waiters_spec.rb
```diff
@@ -0,0 +1,708 @@
...
+                }
+              }
+            end
+
+            it 'checks acceptors in order' do
+              client.stub_responses(:get_widget, my_error)
+              expect do
+                client.wait_until(:acceptor_order_success_matcher, input, max_wait_time: 60)
+              end.to_not raise_error
+              client.stub_responses(:get_widget, my_error)
```
mullermp: Shouldn't need to repeat stub response.

mullermp (commented):
Nice! Clean ups are looking good. This should be very close.

mullermp (approved):

PR #293, opened 2025-04-28
Address some TODOs

mullermp:
Deleted some TODOs or addressed them.


gems/smithy-cbor/lib/smithy-cbor/decoder.rb
```diff
@@ -5,8 +5,7 @@
 module Smithy
   module CBOR
     # @api private
-    # rubocop:disable Metrics/ClassLength
-    class Decoder
+    class Decoder # rubocop:disable Metrics/ClassLength
```
jterapin, line 8: Is this `rubocop:disable` scoped to this class? If so, that's pretty neat to know.

gems/smithy-cbor/lib/smithy-cbor/deserializer.rb
```diff
@@ -63,32 +63,32 @@ def map(shape, values, target = nil)
...
-          target[name] = shape(member_shape.shape, value)
+          target[member_name] = shape(member_shape.shape, values[key])
         end
         target
       end
 
-      def union(shape, values, target = nil)
-        # TODO: delete target instead of checking key?
+      def union(shape, values, target = nil) # rubocop:disable Metrics/AbcSize
+        raise ArgumentError, "union value includes more than one key, received: #{values.keys}" if values.size > 1
```
jterapin, line 79: So we would want to raise than using the first key/value that we encounter? (I think using the first key/value was the V3 behavior if I remember correctly)

gems/smithy-client/lib/smithy-client/plugins/stub_responses.rb
```diff
@@ -10,7 +10,7 @@ class StubResponses < Plugin
           default: false,
           doc_type: 'Boolean',
           docstring: <<~DOCS)
-            When true, the client will return stubbed responses instead of networking requests.
+            When `true``, the client will return stubbed responses instead of networking requests.
```
jterapin:
```suggestion
            When `true`, the client will return stubbed responses instead of networking requests.
```

gems/smithy/lib/smithy/model.rb
```diff
@@ -2,11 +2,12 @@
 
 require_relative 'model/flattener'
 require_relative 'model/rbs'
-require_relative 'model/shape'
 require_relative 'model/operation_parser'
```
jterapin, line 4: nit: if we are ordering these by alphabetical order `rbs` goes after `operation_parser`

gems/smithy-cbor/lib/smithy-cbor/codec.rb
```diff
@@ -6,8 +6,6 @@
 module Smithy
   module CBOR
     # Codec that serializes and deserializes in CBOR format.
-    # TODO:
-    #   * Update implementation to handle event streams
```
jterapin, line 10: Do we not support eventstreams for CBOR?

projections/shapes/lib/shapes/client.rb
```diff
@@ -158,7 +157,7 @@ class Client < Smithy::Client::Base
     #    - `refresh_retry_token(retry_token, error_info)`
     #    - `record_success(retry_token)`
     # @option options [Boolean] :stub_responses
-    #  When true, the client will return stubbed responses instead of networking requests.
+    #  When `true``, the client will return stubbed responses instead of networking requests.
     #  By default fake responses are generated and returned. You can specify the response data
     #  to return or errors to raise by calling {Stubs#stub_responses}.
     #  @see Stubs
```
jterapin, line 163: Is there a way to fix this weird rendering? Not sure if this is easy to do
Screenshot 2025-04-29 at 10 13 56 AM

jterapin (approved):
Nice clean up!

gems/smithy-cbor/lib/smithy-cbor/codec.rb
mullermp: This TODO would linger a while. Also I think it wouldn't be implemented here? This would make the body of the event stream.

jterapin: That's fair - when the core eventstream components gets implemented, we prob do a circle around in all protocols

gems/smithy-cbor/lib/smithy-cbor/decoder.rb
mullermp: Yeah. I like this for when the lines are short.

gems/smithy-cbor/lib/smithy-cbor/deserializer.rb
mullermp: I'm not sure here. Kevin seems to suggest we should raise. I think using the first might be preferred, but then we have to code around `__type` which I can't seem to find any details for in cbor.

PR #294, opened 2025-05-02
Smithy JSON as a codec

mullermp:
Refactors out the json codec from AWS to its own gem.


gems/smithy-json/lib/smithy-json/serializer.rb
```diff
@@ -0,0 +1,122 @@
...
+        else
+          value
+        end
+      end
+
+      def list(shape, values)
+        values.collect do |value|
+          next if value.nil? && !sparse?(shape)
+
+          if value.nil? && sparse?(shape)
```
richardwang1124: Is the `&& sparse?(shape)` check here necessary since you already have a check for the sparse trait on line 54? Same question for line 69.

gems/smithy-cbor/lib/smithy-cbor/deserializer.rb
```diff
@@ -78,7 +78,7 @@ def structure(shape, values, target = nil)
       def union(shape, values, target = nil) # rubocop:disable Metrics/AbcSize
         raise ArgumentError, "union value includes more than one key, received: #{values.keys}" if values.size > 1
 
-        key, value = values.flatten
+        key, value = values.first
```
richardwang1124, line 71: Is this a bugfix or optimization change?

gems/smithy-json/lib/smithy-json/deserializer.rb
```diff
@@ -0,0 +1,138 @@
...
+        when 'NaN' then ::Float::NAN
+        when nil then nil
+        else value.to_f
+        end
+      end
+
+      def list(shape, values, target = nil)
+        target = [] if target.nil?
+        values.each do |value|
+          next if value.nil? && !sparse?(shape)
```
richardwang1124, line 48: For my understanding, does a shape not having the sparse trait imply that the shape cannot have nil values? If so, how can nil values be introduced to the shape?

richardwang1124 (approved):
Added a few comments to clarify my understanding.

gems/smithy-json/lib/smithy-json/deserializer.rb
mullermp: Yes, we would omit building/parsing nil values if there is no sparse trait.

gems/smithy-cbor/lib/smithy-cbor/deserializer.rb
mullermp: I guess neither. First should get the first key/value and unpack it to key and value. Flatten would do the same thing but would have extra assignments if the array was larger (but we validate against this)

gems/smithy-json/lib/smithy-json/serializer.rb
mullermp: Good catch!

PR #295, opened 2025-05-02
Shaperef

mullermp:
Remove MemberShape as a concept and use ShapeRef which contains the shape target and any traits it has. Some "container" seems to be necessary for codec traversal (we will wrap them in shape ref if not provided) - and operation input/output/errors aren't technically member shapes but rather references (because they have target shapes in the model). Currently, the visitor patterns are very awkward, for traits that must be checked at the member layer and then the shape layer (for example timestampFormat). This also removes some awkward accessors on shapes like "member by location name"


projections/weather/lib/weather/schema.rb
```diff
@@ -20,77 +20,66 @@ module Schema
...
+        operation.output = ShapeRef.new(target: GetCurrentTimeOutput)
         operation.traits = {"smithy.api#readonly" => {}}
       end)
       service.add_operation(:get_forecast, OperationShape.new do |operation|
         operation.id = "example.weather#GetForecast"
         operation.name = "GetForecast"
-        operation.input = GetForecastInput
-        operation.output = GetForecastOutput
+        operation.input = ShapeRef.new(target: GetForecastInput)
+        operation.output = ShapeRef.new(target: GetForecastOutput)
```
jterapin: What benefits does `ShapeRef` bring when you wrap it with Input and Output shapes? The only that is obvious to me is the ability to add/access the `metadata` attribute.

gems/smithy-schema/lib/smithy-schema/shapes.rb
```diff
@@ -30,28 +30,52 @@ def []=(key, value)
         end
       end
 
+      # A reference to a shape.
+      class ShapeRef
+        def initialize(options = {})
+          @target = options[:target]
+          @location = options[:location]
```
jterapin: I'm hoping that we stay away from `location` name since it makes sense with the context of C2j models but to me - I can't see that obvious correlation that it is the represented `memberName` part of a "parent" shape (when using to represent member shapes)

gems/smithy-schema/lib/smithy-schema/shapes.rb
```diff
@@ -30,28 +30,52 @@ def []=(key, value)
         end
       end
 
+      # A reference to a shape.
+      class ShapeRef
```
jterapin, line 34: I'm not sure if this is a good class name for this purpose. I always felt that it was bit off when working with V3. 
While it does contain a reference to the shape - it also contains all the details that are tied to the parent shape. I don't have any suggestion on what is a better name but something I was thinking.

jterapin (commented):
I just reviewed the ShapeRef section - I have additional thoughts but I will share them offline.

projections/weather/lib/weather/schema.rb
mullermp: Talked offline. The codecs (and any visitor) should operate on shaperef. This just makes it easier for our implementation if it's consistent. In the model, they are shaperefs because they target a shape.

gems/smithy-schema/lib/smithy-schema/shapes.rb
mullermp: Talked offline, I think location_name makes sense. Location is the protocol specific bit that we don't need anymore.

gems/smithy-schema/lib/smithy-schema/shapes.rb
mullermp: Talked offline. I can't really think of anything else but I do think that ShapeRef is actually very descriptive - it contains a reference to the shape as you said. It could be "target" or a method called "target" - I tried that and it didn't sit right.

projections/shapes/sig/shapes/schema.rbs
```diff
@@ -1,24 +1,24 @@
...
-    Long: Smithy::Schema::Shapes::IntegerShape
-    Map: Smithy::Schema::Shapes::MapShape
-    OperationInputOutput: Smithy::Schema::Shapes::StructureShape
-    Short: Smithy::Schema::Shapes::IntegerShape
-    String: Smithy::Schema::Shapes::StringShape
-    Structure: Smithy::Schema::Shapes::StructureShape
-    Timestamp: Smithy::Schema::Shapes::TimestampShape
-    Union: Smithy::Schema::Shapes::UnionShape
+    BigDecimal: Smithy::Schema::Shapes::bigDecimal
+    BigInteger: Smithy::Schema::Shapes::bigInteger
```
mullermp, line 4: I need to fix these.

gems/smithy-schema/lib/smithy-schema/shapes.rb
```diff
@@ -228,20 +216,21 @@ def initialize(options = {})
...
         attr_accessor :members_by_type
 
-        # @return [MemberShape]
-        def add_member(name, member_name, shape, type, traits: {})
-          member = MemberShape.new(member_name, shape, traits: traits)
-          @members[name] = member
-          @names_by_member_name[member_name] = name
+        # @return [ShapeRef]
+        def add_member(name, type, shape_ref)
+          super(name, shape_ref)
```
mullermp, line 230: Fix

gems/smithy-client/lib/smithy-client/rpc_v2_cbor/response_parser.rb
```diff
@@ -53,10 +53,10 @@ def extract_error(body, context)
 
         def parse_error_data(context, body, code)
           data = Schema::EmptyStructure.new
-          context.operation.errors.each do |error|
-            next unless error.id == code
+          context.operation.errors.each do |ref|
+            next unless ref.shape.id == code
 
-            data = @codec.deserialize(error, body, error.type.new)
+            data = @codec.deserialize(ref, body, ref.shape.type.new)
```
mullermp, line 59: Let type be made implicitly

gems/smithy-client/lib/smithy-client/rpc_v2_cbor/request_builder.rb
```diff
@@ -31,7 +31,7 @@ def apply_headers(context)
         end
 
         def apply_content_type_header(context)
-          input = context.operation.input
+          input = context.operation.input.shape
```
mullermp, line 34: Clean up

gems/smithy-client/lib/smithy-client/param_validator.rb
```diff
@@ -185,8 +185,8 @@ def validate_required_members(shape, values, errors, context)
         end
       end
 
-      def streaming_input?(shape)
-        shape.traits.include?('smithy.api#streaming')
+      def streaming_input?(ref)
```
mullermp, line 188: Clean up

gems/smithy-cbor/lib/smithy-cbor/serializer.rb
```diff
@@ -13,20 +13,21 @@ def initialize(options = {})
       end
 
       def serialize(shape, data)
-        return nil if shape == Prelude::Unit
+        ref = shape.is_a?(ShapeRef) ? shape : ShapeRef.new(shape: shape)
+        return nil if ref.shape == Prelude::Unit
```
mullermp, line 17: Optimize

mullermp (commented):
Notes for self

gems/smithy-schema/lib/smithy-schema/shapes.rb
```diff
@@ -30,28 +30,49 @@ def []=(key, value)
...
+          @location_name = options[:location_name]
+          @traits = options[:traits] || {}
+          @metadata = {}
+        end
+
+        # @return [Shape]
+        attr_accessor :shape
+
+        # @return [String, nil]
+        attr_reader :location_name
```
jterapin, line 46: Could we add a description here?

jterapin (approved):

PR #296, opened 2025-05-06
Cleanups

mullermp:
Rename location_name to member_name so it is self documenting.

Fix RBS for shape schema.

SERDE clean ups.


gems/smithy-json/lib/smithy-json/serializer.rb
```diff
@@ -11,10 +11,10 @@ def initialize(options = {})
       end
 
       def serialize(shape, data)
-        return nil if shape == Prelude::Unit
-
         ref = shape.is_a?(ShapeRef) ? shape : ShapeRef.new(shape: shape)
-        ::JSON.dump(shape(ref, data))
+        return nil if ref.shape == Prelude::Unit
```
jterapin, line 15: Nit: I don't think we need to explicitly return `nil` here, right?

jterapin (approved):
LGTM

gems/smithy-json/lib/smithy-json/serializer.rb
mullermp: No. I can remove this in the next change.

PR #297, opened 2025-05-07
Defaults

mullermp:
Handle the default trait.

Currently have two approaches to consider:

1) Handle via types. The biggest issue here is that input types are not used. It will require work to convert hashy input into types prior to serialization.

There are two sub-approaches - one that overrides initialize and the other that sets a defaults instance variable for Smithy::Schema::Structure to use implicitly.

2) Handle via codecs. This approach would not change anything major, but it requires all codecs to understand defaults. It also means that types do not have defaults built in. The codec would look at the default value and apply it if applicable.

(Note: we could implement both approaches but that seems wasteful)

~~I chose option 2 here after going down a rabbit hole. It seems cbor protocol should not serialize top level members on the client. Because of this, types would need to know protocol specific information. Protocols have client and server implementations, making it ideal to handle it at runtime as codecs (with a configurable option regarding top level).~~

I chose option 1 again, by using synthetic input/output. These synthetic shapes will have input/output traits, and if those traits are present, defaults generation does not apply.


gems/smithy-schema/lib/smithy-schema/structure.rb
```diff
@@ -4,6 +4,22 @@ module Smithy
   module Schema
     # A module mixed into Structs that provides utility methods.
     module Structure
+      #       <% if type.defaults.any? -%>
```
mullermp: Sub-approach 2 - the following code can handle it implicitly.

gems/smithy-cbor/lib/smithy-cbor/deserializer.rb
```diff
@@ -48,43 +48,56 @@ def map(ref, values, target = nil)
...
-          target[member_name] = shape(member_ref, values[key])
+          value = values[member_ref.member_name]
+          value = default(member_ref) if value.nil? && default?(member_ref.traits)
+          target[member_name] = shape(member_ref, value)
         end
         target
       end
 
       def union(ref, values, target = nil) # rubocop:disable Metrics/AbcSize
-        raise ArgumentError, "union value includes more than one key, received: #{values.keys}" if values.size > 1
```
jterapin, line 70: Do we have a documentation that confirms that we shouldn't raise when we receive more than one key? I agree that this is probably what we want to do but was curious.

gems/smithy-cbor/lib/smithy-cbor/deserializer.rb
```diff
@@ -48,43 +48,56 @@ def map(ref, values, target = nil)
...
-          next unless values.key?(name)
+          value = values[member_ref.member_name]
+          next if value.nil?
 
           target = ref.shape.member_type(member_name) if target.nil?
-          return target.new(shape(member_ref, values[name]))
+          return target.new(shape(member_ref, value))
         end
+
+        values.delete('__type')
```
jterapin, line 74: Should we do this earlier within the method? Feels like this `__type` doesn't really matter within union context

gems/smithy-client/lib/smithy-client/signers/http_basic.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'base64'
+
```
jterapin, line 4: Do we need to require this here too? I noticed we already require this under `Smithy::Client` level

gems/smithy-client/lib/smithy-client/stubbing/data_applicator.rb
```diff
@@ -18,30 +18,30 @@ def apply(data, stub)
 
         private
 
+        def shape(ref, value)
+          case ref.shape
+          when StructureShape then structure(ref.shape, value, ref.shape.type.new)
```
jterapin: Since we use structure's type here - should we do the same for unions?

gems/smithy-json/spec/smithy-json/codec_spec.rb
```diff
@@ -47,60 +43,60 @@ module JSON
           union: { string: 'string' }
         }
         data = data.merge(structure: data)
-        bytes = subject.serialize(structure_shape, data)
-        expect(subject.deserialize(structure_shape, bytes).to_h).to eq(data)
+        json = subject.serialize(structure_shape, data)
+        expect(subject.deserialize(structure_shape, json).to_h).to eq(data)
       end
 
       context 'structures' do
```
jterapin, line 51: Unrelated to defaults but we should probably add a test related to timestampFormat trait.
(to ensure that we resolve timestamp formats per [here](https://smithy.io/2.0/spec/protocol-traits.html#timestampformat-trait))

jterapin (approved):
Nice, I'll follow this PR format to update my Document PR to stay aligned. We should chat about what you mean by going with option 1.

gems/smithy-cbor/lib/smithy-cbor/deserializer.rb
mullermp: It's documented as such in Smithy - we're supposed to select the first non nil value, so we were just doing it wrong to begin with.

gems/smithy-cbor/lib/smithy-cbor/deserializer.rb
mullermp: No I believe we have to do it after - because __type could be a jsonName - so we first check all the members, and if we didn't parse anything, delete it, then check for an unknown.

gems/smithy-client/lib/smithy-client/signers/http_basic.rb
mullermp: I think it's best practice to require things where they are used even if they may have been required before. We know this file depends on base64 existing.

gems/smithy-json/spec/smithy-json/codec_spec.rb
mullermp: This is tested with protocol tests. I can add a todo for adding more codec tests.

gems/smithy-client/lib/smithy-client/stubbing/data_applicator.rb
mullermp: There's no union stubbing here. What did you have in mind?

gems/smithy-cbor/lib/smithy-cbor/deserializer.rb
jterapin: I see. CBOR protocol does not support jsonName trait so I was under the impression that we could just ignore `__type` entirely.

gems/smithy-client/lib/smithy-client/stubbing/data_applicator.rb
jterapin: I was thinking that if you stub runtime shape for `StructureShape` -  I figured we do the same for when we want to stub `UnionShape` subclasses but I guess that is highly unlikely for that to happen. I was just curious.l

gems/smithy-cbor/lib/smithy-cbor/deserializer.rb
mullermp: Oh. Yes, I suppose we could. I just copied what we had for JSON here.

gems/smithy-client/lib/smithy-client.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'base64'
+require 'bigdecimal'
```
alextwoods, line 4: Should we be adding bigdecimal as a required, general dependency? If I remember it can cause some issues for users without compilers/jruby and SDKs dont generally support it.  Maybe add it as a dependency in client gems only when the model requires it?

gems/smithy-cbor/spec/smithy-cbor/codec_spec.rb
```diff
@@ -83,31 +84,13 @@ module CBOR
...
-
         it 'deserializes unknown union members' do
           unknown_union_type = shape.member(:union).shape.member_type(:unknown)
           data = { union: { 'someThing' => 'someValue' } }
           deserialized = subject.deserialize(shape, CBOR.encode(data))
           expect(deserialized.union).to be_a(unknown_union_type)
           expect(deserialized.union.to_h).to eq(unknown: { name: 'someThing', value: 'someValue' })
         end
-
-        it 'raises when deserializing unions with more than one member' do
```
alextwoods, line 106: Is there a reason we're removing these tests?

gems/smithy-client/lib/smithy-client/handler_context.rb
```diff
@@ -33,7 +33,7 @@ def initialize(options = {})
       # @return [Base]
       attr_accessor :client
 
-      # @return [Hash] The hash of request parameters.
+      # @return [Hash, Struct] The request parameters as a Hash or a Struct.
```
alextwoods, line 36: I assume Struct here would be the runtime input shape?

This is a little confusing - I assume the reason for the union of types here is that at the start of the request we have hash params and after the parameter conversion handler runs, we have runtime shapes?

gems/smithy-client/lib/smithy-client/param_converter.rb
```diff
@@ -39,67 +40,67 @@ def c(ref, value)
...
 
       def structure(ref, values)
         values = c(ref, values)
-        if values.respond_to?(:each_pair)
-          values.each_pair do |k, v|
-            next if v.nil?
+        return if values.nil?
 
-            next unless ref.shape.member?(k)
+        type = @convert_structures ? ref.shape.type.new : values
```
alextwoods, line 73: When convert_structures is false, this should copy values right?  Otherwise the input is being modified

gems/smithy/lib/smithy/views/client/types.rb
```diff
@@ -88,9 +83,42 @@ def attribute_docstrings
...
               lines << "  #{docstring}"
             end
-            lines << "  @return [#{Model::YARD.type(@model, @id, @target)}]"
+            lines << "  @return [#{@doc_type}]"
             lines
           end
+
+          def default?
+            traits = @member.fetch('traits', {})
+            traits.key?('smithy.api#default') && !traits.key?('smithy.api#clientOptional')
```
alextwoods, line 92: This gets a little confusing for shape only gems / server generation - the smithy docs say 
> The [clientOptional trait](https://smithy.io/2.0/spec/type-refinement-traits.html#clientoptional-trait) applied to a member marked with the default trait causes [non-authoritative generators](https://smithy.io/2.0/guides/building-codegen/overview-and-concepts.html#client-generation) to ignore the default trait.

So, in a client context, we should use it, but in an authoritative (server) type we shouldn't.  I'm not sure if theres a good way to handle that in a server/client agnostic generation though and since we're targeting client generation, maybe this is good for now.

gems/smithy/lib/smithy/views/client/types.rb
```diff
@@ -88,9 +83,42 @@ def attribute_docstrings
...
+            when 'document' then document(default)
+            when 'enum', 'string' then "'#{default}'"
+            when 'timestamp' then timestamp(default)
+            else default
+            end
+          end
+
+          def document(default)
+            case default
+            when nil then 'nil'
```
alextwoods, line 109: Should a check for nil default be done above and default just be skipped completely if its nil or is this a special document case?

gems/smithy/lib/smithy/welds/synthetic_input_output.rb
```diff
@@ -0,0 +1,53 @@
...
+
+      def create_synthetic_input_shape(model, operation_id, operation)
+        input_target = operation['input']['target']
+        target = Model.shape(model, input_target)
+        return if target.fetch('traits', {}).include?('smithy.api#input') || input_target == 'smithy.api#Unit'
+
+        input_shape = {
+          'type' => 'structure',
+          'traits' => target.fetch('traits', {}).merge({ 'smithy.api#input' => {} }),
+          'members' => target.fetch('members', {}).merge({})
```
alextwoods: Do we need to do a deep copy of members and traits so that another weld that modifies the synthetic shape don't modify the original.  I know the merge will do a shallow copy and I can't remember the details of how these are structured so this may not apply.

gems/smithy/lib/smithy/welds/synthetic_input_output.rb
```diff
@@ -0,0 +1,53 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Welds
+    # Creates synthetic input and output shapes for operations that do not have them.
+    class SyntheticInputOutput < Weld
```
alextwoods, line 8: I don't remember all of the history/details on this - but I remember with our java generator we had similar logic which we ended up removing because the logic was pushed up stream.

alextwoods (commented):
Nice - looks good overall - I think the approach makes sense

gems/smithy-client/lib/smithy-client.rb
mullermp: I think @hsbt has work to make it compatible everywhere. We can check.

gems/smithy-cbor/spec/smithy-cbor/codec_spec.rb
mullermp: I think this was wrong behavior when I had chatted with Kevin. I'm not really sure if we should be raising. Let me think on this.

gems/smithy-client/lib/smithy-client/handler_context.rb
mullermp: Yeah, at the start of the request, we would have a hash OR struct, but eventually it will always be a struct. I'm not sure if we want to do this though - but it allows us to get values from types (like defaults) if we use the input type.

gems/smithy-client/lib/smithy-client/param_converter.rb
mullermp: `dup` is called by the converter prior to this function call if it's a hash.

gems/smithy/lib/smithy/views/client/types.rb
mullermp: Yeah. Good call out.. I'm not sure how else to handle this? I'll chat with Michael.

gems/smithy/lib/smithy/views/client/types.rb
mullermp: This is a document case.

gems/smithy/lib/smithy/welds/synthetic_input_output.rb
mullermp: Yes. This was on my radar. We need a deep copy.

gems/smithy/lib/smithy/welds/synthetic_input_output.rb
mullermp: I believe it needs to be handled by us - the java ModelTransformer does this effectively for us, but I don't have that. The issue is that defaults only apply to shapes not input/output, but if you have synthetics, it not only helps with model evolution, but it allows you to apply defaults at the non-top-level scope which is required by protocol tests.

gems/smithy-schema/lib/smithy-schema/shapes.rb
```diff
@@ -228,7 +233,7 @@ def initialize(options = {})
         # @return [ShapeRef]
         def add_member(name, type, shape_ref)
           @member_types[name] = type
-          @members_by_type[type] = shape_ref
+          @members_by_type[type] = [name, shape_ref]
```
jterapin, line 236: I was wondering why I suddenly seeing `__name` in the above but I see why now.

gems/smithy-schema/spec/support/schema_helper.rb
```diff
@@ -85,7 +85,10 @@ def sample_shapes
             'short' => { 'target' => 'smithy.api#Short' },
             'streamingBlob' => {
               'target' => 'smithy.ruby.tests#StreamingBlob',
-              'traits' => { 'smithy.api#default' => 'streamingBlob' }
+              'traits' => {
+                'smithy.api#required' => {},
+                'smithy.api#clientOptional' => {}
+              }
```
jterapin, line 91: Should we move this closer to the tests itself? Feedback: https://github.com/smithy-lang/smithy-ruby/pull/282#discussion_r2080815568

Probably could look over the `sample_shapes` to see if there's any opportunity to move test-specific traits closer.

gems/smithy/lib/smithy/welds/synthetic_input_output.rb
jterapin: In thinking of speculative "typed document" cases -  us manipulating the shape ids on our end will have a trickle effect when we receive a typed document from a service and the underlying discriminator does not match our structures within type registry (since we replaced the shape id with a new one) - unless we intend to keep both synthetic and modeled shapes.

Disregard this comment if this was handled already.

PR #299, opened 2025-05-21
Remove Errors namespace under Smithy::Client

mullermp:
Remove the Errors namespace under Smithy::Client to flatten it


gems/smithy-client/lib/smithy-client/dynamic_errors.rb
```diff
@@ -0,0 +1,82 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    # This module is mixed into another module, providing dynamic
```
jterapin, line 5: Nit: I know this is `@api private` so I'm thinking this does not matter much but wording of "this module is mixed into another module" does not sound right to me. Prob copied over from previously.

gems/smithy-client/lib/smithy-client.rb
```diff
@@ -26,6 +26,7 @@
 require_relative 'smithy-client/plugin'
 require_relative 'smithy-client/plugin_list'
 require_relative 'smithy-client/retry'
+require_relative 'smithy-client/service_error'
```
jterapin, line 29: Should `service_error` file be required prior to `dynamic_errors`? Since Dynamic error classes inherits from Service Error?

jterapin (approved):
Nice, i like the flattening - Everything looks fine but this [test file](https://github.com/smithy-lang/smithy-ruby/blob/decaf/gems/smithy-client/spec/smithy-client/errors_spec.rb) needs an update based on flattening changes

gems/smithy-client/lib/smithy-client/dynamic_errors.rb
mullermp: What do you mean? It's a module that is included in generated Errors module (it's a ruby mixin).

gems/smithy-client/lib/smithy-client.rb
mullermp: It should not be necessary because a dynamic error is not created before loading files.

gems/smithy-client/lib/smithy-client/dynamic_errors.rb
jterapin: Right but without context, i'm kind of like what other module would this applicable?

"This is a mixin module, intended to be used in a generated Errors module..." or "This module is intended to mix into the generated Errors module..."

PR #300, opened 2025-06-13
Populate input defaults as a plugin

mullermp:
Breaks the symmetry of the defaults trait implementation. Defaults are added into type initializers for schema related gems, but for clients, are really used for output. For input, add a plugin to the client to populate default values. This allows us to retain "hash only" input and not worry about structure/union type cases and do any weird conversion. We can expect a hash throughout.

### DIFF
--- a/gems/smithy-client/lib/smithy-client/default_params.rb
+++ b/gems/smithy-client/lib/smithy-client/default_params.rb
@@ -1,0 +1,88 @@
+# frozen_string_literal: true
+
+require 'base64'
+require 'time'
+
+module Smithy
+  module Client
+    # @api private
+    class DefaultParams
+      include Schema::Shapes
+
+      def initialize(ref)
+        @ref = ref
+      end
+
+      # @param [Hash] params
+      # @return [Hash]
+      def apply(params)
+        structure(@ref, params)
+      end
+
+      private
+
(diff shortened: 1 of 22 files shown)


gems/smithy-client/lib/smithy-client/default_params.rb
```diff
@@ -0,0 +1,88 @@
+# frozen_string_literal: true
+
+require 'base64'
+require 'time'
+
+module Smithy
+  module Client
+    # @api private
+    class DefaultParams
```
jterapin, line 9: Need spec file for this

jterapin (commented):
Simplifying to "hash only" input reduces overall complexity so I'm onboard. So when we codegenerated `schema.rb`, the shapes will include default trait if there is one. 

Could we have a "schema-only" service in our projections? (with some shapes with default trait). So that I can visually see on what this looks like. We probably want one of our projections to be schema-only (for development purposes).

gems/smithy-client/lib/smithy-client/default_params.rb
mullermp: It's covered by protocol tests but I can do another one.

jterapin (approved):

PR #302, opened 2025-06-19
request/response -> http_request/response

mullermp:
Rename request/response in context to http_request/response. It will be http or http2 (both are http). Other transports can have a separate accessor such as mqtt if/when that happens.


jterapin (approved):
Approved, we discussed ways to mitigate when newer transports are introduced:
* Update the HandlerContext to include other transports like `mqtt_request`
* MV bump smithy-client when use case arises

PR #303, opened 2025-06-19
Refactor signing and auth

richardwang1124:
Refactors signing and auth to use V3 context bag approach. Simplifies and removes AuthOptions/ResolvedAuth, removes identity and signer properties, and updates sign to use context. Uses discrete plugin approach: https://github.com/smithy-lang/smithy-ruby/pull/309


gems/smithy-client/lib/smithy-client/signers/http_api_key.rb
```diff
@@ -5,7 +5,11 @@ module Client
     module Signers
       # A signer that signs requests using the HTTP API Key Auth scheme.
       class HttpApiKey < Signer
-        def sign(request:, identity:, properties:)
+        # rubocop:disable Metrics/AbcSize
+        def sign(context)
+          request = context.http_request
+          identity = context[:auth].identity
+          properties = context.config.service.traits['smithy.api#httpApiKeyAuth']
```
richardwang1124: Is this the best way to retrieve this?

gems/smithy/lib/smithy/templates/client/auth_plugin.erb
```diff
@@ -83,29 +83,22 @@ module <%= module_name %>
             return
           end
 
-          identity_properties = auth_option.identity_properties
-          identity = identity_provider.identity(identity_properties)
+          identity = identity_provider.identity({})
```
richardwang1124: If we remove `identity_properties` should we also update identity to no longer take a parameter?

projections/shapes/lib/shapes/auth_resolver.rb
```diff
@@ -6,10 +6,10 @@ module ShapeService
   # Resolves the auth scheme from {AuthParameters}.
   class AuthResolver
     # @param [AuthParameters] parameters
-    # @return [Smithy::Client::AuthOption]
+    # @return [String]
     def resolve(parameters)
```
richardwang1124, line 10: So I'm still a little confused about auth parameters. When will they ever be used? Technically `region` should be an auth parameter for SigV4, but the AuthParameter class only has `name` and `documentation` members. If it did include region then we'd be able to take that parameter and add it to signer properties in resolve. However it seems like parameters has always been unused?

gems/smithy-client/lib/smithy-client/signers/anonymous.rb
```diff
@@ -5,7 +5,7 @@ module Client
     module Signers
       # A signer that does not sign requests.
       class Anonymous < Signer
-        def sign(**); end
+        def sign(context); end
         def reset(**); end
```
mullermp: Reset should take context too.

gems/smithy-client/lib/smithy-client/signers/http_api_key.rb
```diff
@@ -5,7 +5,11 @@ module Client
     module Signers
       # A signer that signs requests using the HTTP API Key Auth scheme.
       class HttpApiKey < Signer
-        def sign(request:, identity:, properties:)
+        # rubocop:disable Metrics/AbcSize
+        def sign(context)
+          request = context.http_request
+          identity = context[:auth].identity
```
mullermp: We can consider making an accessor for auth instead of metadata - it is non-optional (it can default to a no-auth).

gems/smithy-client/lib/smithy-client/signers/http_api_key.rb
mullermp: Yes.

gems/smithy-client/lib/smithy-client/signers/http_basic.rb
```diff
@@ -7,11 +7,11 @@ module Client
     module Signers
       # A signer that signs requests using the HTTP Basic Auth scheme.
       class HttpBasic < Signer
-        def sign(request:, identity:, **_options)
+        def sign(context)
           # TODO: does not handle realm or other properties
-          identity_string = "#{identity.username}:#{identity.password}"
+          identity_string = "#{context[:auth].identity.username}:#{context[:auth].identity.password}"
```
mullermp: This is fine, but you can store context[:auth] since it's used twice. Hash access is always slower than local variable.

gems/smithy/lib/smithy/templates/client/auth_plugin.erb
```diff
@@ -54,7 +54,7 @@ module <%= module_name %>
           }
 
           auth_options.each do |auth_option|
```
mullermp: Another thing to investigate - I'm open to pushing this handler code to smithy-client. We still need a generated plugin for defaults and other properties, but the handler can be made generic. It cuts down on generated code.

gems/smithy/lib/smithy/templates/client/auth_plugin.erb
mullermp: Yes. We can always add it to the method signature later (we discussed that earlier).

gems/smithy/lib/smithy/templates/client/auth_plugin.erb
```diff
@@ -83,29 +83,22 @@ module <%= module_name %>
             return
           end
 
-          identity_properties = auth_option.identity_properties
-          identity = identity_provider.identity(identity_properties)
+          identity = identity_provider.identity({})
 
           ResolvedAuth.new(
-            scheme_id: scheme_id,
```
mullermp, line 90: I think we still need scheme_id here right? Otherwise we are not sure what auth we actually resolved.

gems/smithy/lib/smithy/views/client/auth_resolver.rb
```diff
@@ -116,8 +116,7 @@ def add_registered_auth_schemes(auth_schemes, traits)
         end
 
         def render_auth_option(auth_scheme)
-          properties = @service_traits.fetch(auth_scheme, {})
-          "Smithy::Client::AuthOption.new(scheme_id: '#{auth_scheme}', signer_properties: #{properties})"
+          "'#{auth_scheme}'"
```
mullermp: This method doesn't make sense anymore.

projections/shapes/lib/shapes/auth_resolver.rb
mullermp: They are used to resolve the auth scheme. Operation name should be used here. Check the auth resolver generator, but I have a "simple case" and a "complex case" that is generated. This code here is the simple case so it's not used. The complex case would look at operation name, and if there was operation auth preference, it would resolve a different order. I think now that we use context for signing, auth parameter doesn't need region, because we would never determine what scheme to use based off region (if we needed to, we would need to add it as a param). The signer would pull the region it needed in that case.

gems/smithy/lib/smithy/templates/client/auth_plugin.erb
richardwang1124: Can signers sign for multiple auth schemes? If so then we would need to keep `scheme_id`.

gems/smithy-client/lib/smithy-client/signers/anonymous.rb
richardwang1124: For my understanding, what is reset used for? I don't see any usage in V4 so far and I don't think I see it used in Smithy Ruby either..

gems/smithy-client/lib/smithy-client/signers/http_api_key.rb
richardwang1124: Is there any harm in having both auth and metadata? Does metadata contain anything needed for auth/signing?

gems/smithy/lib/smithy/templates/client/auth_plugin.erb
richardwang1124: I think that makes sense, the only templated things in this file are the auth_schemes and identities.

gems/smithy/lib/smithy/views/client/auth_resolver.rb
richardwang1124: Yep refactor and remove this.

gems/smithy-client/lib/smithy-client/signers/anonymous.rb
mullermp: It should be used in the retry handler. If it's not, it should be. It should undo signing for retrying (signing may need time sensitive information)

gems/smithy-client/lib/smithy-client/signers/http_api_key.rb
mullermp: Context[] adds to the metadata hash. I'm saying that we could have context.auth instead because auth is an expected thing in smithy.

gems/smithy/lib/smithy/templates/client/auth_plugin.erb
mullermp: No I don't think it could sign multiple auth schemes currently but that is an interesting use case. I do think we may want to know what we are signing for and what scheme was resolved (maybe we do other logic based on that).

gems/smithy-client/lib/smithy-client/signers/anonymous.rb
richardwang1124: Hm I don't think I see it. The reset methods are defined for the signers but I don't see them being used in the retry handler. I do see one instance of `reset` used in `context.http_response.reset` but that uses http response's `reset` method.

gems/smithy-client/lib/smithy-client/signers/http_api_key.rb
```diff
@@ -37,6 +45,8 @@ def append_query_param(request, name, value)
         end
 
         def remove_query_param(request, name)
+          return unless request.endpoint.query
```
richardwang1124: Since we're resetting at the start of sign, if `request.endpoint.query` is nil then `split` in `CGI.parse` returns an error, so I added a guard clause here.

gems/smithy/lib/smithy/templates/client/auth_resolver.erb
```diff
@@ -5,9 +5,10 @@
 module <%= module_name %>
   # Resolves the auth scheme from {AuthParameters}.
   class AuthResolver
-    # @param [AuthParameters] parameters
-    # @return [Smithy::Client::AuthOption]
-    def resolve(parameters)
+    # @param [HandlerContext] context
+    # @return [String]
+    def resolve(context)
+      parameters = AuthParameters.create(context)
```
richardwang1124: This used to be called within the Auth plugin, but it's easier to access AuthParameters within AuthResolver so I updated the resolve method to take the context instead.

gems/smithy-client/lib/smithy-client/signers/http_api_key.rb
```diff
@@ -5,7 +5,12 @@ module Client
     module Signers
       # A signer that signs requests using the HTTP API Key Auth scheme.
       class HttpApiKey < Signer
-        def sign(request:, identity:, properties:)
+        # rubocop:disable Metrics/AbcSize
+        def sign(context)
```
jterapin: Rather than enabling and disabling this specific cop on this method. You could:
```
def sign(context) # rubocop:disable Metrics/AbcSize
```
Then, this disable will be limited to that method. Matt had this in his PR and it was neat!

gems/smithy-client/spec/smithy-client/plugins/checksum_required_spec.rb
```diff
@@ -17,6 +17,7 @@ module Plugins
           client_class.add_plugin(sample_client::Plugins::Auth)
           client_class.add_plugin(sample_client::Plugins::Endpoint)
           client_class.add_plugin(AnonymousAuth)
+          client_class.add_plugin(Auth)
```
jterapin: This new plugin shared very similar namespaces with the codegenerated auth. I'm not sure if this is a potential piece of concern over time. What do you think, Matt?

gems/smithy-client/lib/smithy-client/signers/http_api_key.rb
mullermp: Seems fine

gems/smithy-client/lib/smithy-client/plugins/auth.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    module Plugins
+      # @api private
+      class Auth < Plugin
```
mullermp: I don't think we need another plugin - just make the handler class live inside smithy-client and the code generated auth plugin should register it.

gems/smithy-client/spec/smithy-client/plugins/checksum_required_spec.rb
mullermp: This plugin should not exist IMO.

gems/smithy/lib/smithy/templates/client/auth_plugin.erb
```diff
@@ -29,87 +29,6 @@ module <%= module_name %>
...
-            @signer_properties = options[:signer_properties]
-            @identity = options[:identity]
-            @identity_properties = options[:identity_properties]
-          end
-
-          attr_accessor :scheme_id, :signer, :signer_properties, :identity, :identity_properties
-        end
-      end
-
-      handler(Handler, step: :sign, priority: 70)
```
mullermp, line 112: You would keep this line, but it would be `Smithy::Client::AuthSchemes::Handler` or something (hopefully we can just delete all those, but we will see).

gems/smithy/lib/smithy/templates/client/auth_resolver.erb
mullermp: I'm not sure about this one .. let me think. AuthResolver is a public concept. Otherwise, I think this belongs in auth_rules_code generation. Let's discuss.

gems/smithy-client/spec/smithy-client/plugins/checksum_required_spec.rb
richardwang1124: Removed the plugin

gems/smithy-client/lib/smithy-client/plugins/logging.rb
```diff
@@ -18,6 +18,7 @@ class Logging < Plugin
           :log_level,
           default: :info,
           doc_type: Symbol,
+          doc_default: ':info',
```
mullermp, line 21: This isn't working as expected - I think this should be `"':info'"` otherwise I think we may just have a small bug with option default generation. In v3 this handles correctly.

mullermp (approved):
MASSIVE

gems/smithy-client/lib/smithy-client/plugins/resolve_auth.rb
```diff
@@ -0,0 +1,86 @@
...
+                failures << "Auth scheme #{auth_option} was not enabled for this request"
+                next
+              end
+
+              identity_provider = context.config.auth_schemes[auth_option]
+              resolved_auth = try_load_auth_scheme(
+                auth_option,
+                identity_provider,
+                failures
+              )
```
jterapin: I think this fits in one-line with the 120 line limit? Disregard if it doesn't

gems/smithy-client/lib/smithy-client/plugins/resolve_auth.rb
```diff
@@ -0,0 +1,86 @@
...
+
+              return resolved_auth if resolved_auth
+            end
+
+            raise failures.join("\n")
+          end
+
+          def try_load_auth_scheme(auth_option, identity_provider, failures)
+            scheme_id = auth_option
```
jterapin: is this assignment necessary?

gems/smithy-client/lib/smithy-client/plugins/resolve_auth.rb
```diff
@@ -0,0 +1,86 @@
...
+
+            unless identity_provider
+              failures << "Auth scheme #{scheme_id} did not have an identity resolver configured"
+              return
+            end
+
+            identity = identity_provider.identity
```
jterapin, line 63: Same goes for this one as well

jterapin (approved):
LGTM, great simplification.

PR #306, opened 2025-06-24
Add v3 style unions

mullermp:
Use Ruby SDK v3 style union sealed classes - no longer generate `to_h` methods and make them structs just like structures. Usage:

```
case resp.union.member
when :foo then do_thing(resp.union.value)
when :unknown then do_unknown
end

# or

case resp.union
when Service::Types::Union::Foo then do_thing(resp.union.value)
when Service::Types::Union::Unknown then do_unknown
end
```

Parsing will use subclasses of a struct:

```
union = Service::Types::Union::Foo.new(foo: 'foo')

union.members # [:foo, :unknown] new method as well as other struct methods
union.member # [:foo], because it is set
union.value # foo, because it is set

# instead of

union = Service::Types::Union::Foo.new('foo') # delegation
```

### DIFF
--- a/gems/smithy-cbor/lib/smithy-cbor/deserializer.rb
+++ b/gems/smithy-cbor/lib/smithy-cbor/deserializer.rb
@@ -68,12 +68,12 @@ module Smithy
           next if value.nil?
 
           target = ref.shape.member_type(member_name) if target.nil?
-          return target.new(shape(member_ref, value))
+          return target.new(member_name => shape(member_ref, value))
         end
 
         values.delete('__type')
         key, value = values.first
-        ref.shape.member_type(:unknown).new(key, value)
+        ref.shape.member_type(:unknown).new(unknown: { key => value })
       end
 
       def sparse?(shape)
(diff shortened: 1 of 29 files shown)


gems/smithy-cbor/lib/smithy-cbor/deserializer.rb
```diff
@@ -68,12 +68,12 @@ def union(ref, values, target = nil) # rubocop:disable Metrics/AbcSize
...
 
           target = ref.shape.member_type(member_name) if target.nil?
-          return target.new(shape(member_ref, value))
+          return target.new(member_name => shape(member_ref, value))
         end
 
         values.delete('__type')
         key, value = values.first
-        ref.shape.member_type(:unknown).new(key, value)
+        ref.shape.member_type(:unknown).new(unknown: { key => value })
```
jterapin, line 76: I forgot where we land on handling a response with multiple union members.

jterapin (approved):

gems/smithy-cbor/lib/smithy-cbor/deserializer.rb
mullermp: I think we would either raise or take the first. We don't have to solve that now.

PR #307, opened 2025-06-25
Client log formatter

mullermp:
Adds a log formatter similar to SDK v3. It takes a string pattern and outputs to a configured logger.


gems/smithy-client/lib/smithy-client/log_formatter.rb
```diff
@@ -0,0 +1,229 @@
...
+
+      # Given a response, this will format a log message and return it as a string according to {#pattern}.
+      # @param [Smithy::Client::Response] response
+      # @return [String]
+      def format(response)
+        pattern.gsub(/:(\w+)/) { |sym| send("_#{sym[1..]}", response) }
+      end
+
+      # @api private
+      def method_missing(method_name, *args)
```
richardwang1124, line 71: What is this method used for? I don't see any usages in V3 either.

richardwang1124 (approved):

PR #309, opened 2025-06-26
Discrete Plugin Approach

richardwang1124:
Refactor signing/auth to couple signing logic with auth plugin. Adds new ResolveAuth plugin to resolve auth scheme.


gems/smithy-client/lib/smithy-client/plugins/anonymous_auth.rb
```diff
@@ -14,9 +13,14 @@ class AnonymousAuth < Plugin
           AnonymousProvider.new
         end
 
-        option(:anonymous_auth_scheme) do |_config|
-          Smithy::Client::AuthSchemes::Anonymous.new
+        class Handler < Client::Handler
+          def call(context)
+            Smithy::Client::Signers::Anonymous.new.sign(context)
```
mullermp: You could go even further and delete signers too and just push the logic in here - but let's see if this prototype works first. Also this should also check for noAuth.

gems/smithy/lib/smithy/templates/client/auth_plugin.erb
```diff
@@ -20,19 +20,19 @@ module <%= module_name %>
       option(
         :auth_schemes,
         doc_type: Hash,
-        rbs_type: 'Hash[String, Smithy::Client::AuthScheme]',
+        rbs_type: 'Hash[String, String]',
```
mullermp: I think this should map string to identity provider, and values will be populated by plugins.

gems/smithy/lib/smithy/templates/client/client.erb
```diff
@@ -133,14 +133,13 @@ module <%= module_name %>
         <%= protocols %>
       end
 
-
       # @api private
-      def identity_providers(context)
```
mullermp, line 138: I think we'll be able to just delete this method entirely because we can look at config.auth_scheme to get the providers to call.

gems/smithy-client/lib/smithy-client/auth/auth.rb
```diff
@@ -54,7 +54,6 @@ def try_load_auth_scheme(auth_option, auth_scheme, identity_providers, failures)
           {
             scheme_id: scheme_id,
             identity: identity_provider.identity,
-            signer: auth_scheme.signer
```
mullermp: Before I forget, I think we will need to check if identity actually returns something here, before returning from this method. It could raise or return nil and this continues.

gems/smithy-client/lib/smithy-client/plugins/anonymous_auth.rb
```diff
@@ -13,8 +13,8 @@ class AnonymousAuth < Plugin
           AnonymousProvider.new
         end
 
-        def before_initialize(_client_class, _options)
-          ResolveAuth.add_auth_scheme('smithy.api#noAuth', :anonymous_provider)
+        def after_initialize(client)
+          client.config.auth_schemes['smithy.api#noAuth'] = :anonymous_provider
```
mullermp: I think set these to client.config.whatever_provider.

But also - I think you can entirely remove the anonymous auth plugin. If noAuth scheme is resolved, nothing happens anyway. This needed to exist before, but I think it doesn't now right?

gems/smithy-client/lib/smithy-client/plugins/http_api_key_auth.rb
```diff
@@ -2,8 +2,7 @@
 
 require_relative '../http_api_key_provider'
 require_relative '../identities/http_api_key'
-require_relative '../signers/http_api_key'
-require_relative '../auth_schemes/http_api_key'
+require_relative 'resolve_auth'
```
mullermp: This doesn't need to be required anymore in these plugins.

gems/smithy-client/lib/smithy-client/plugins/http_api_key_auth.rb
```diff
@@ -28,9 +27,69 @@ class that responds to #identity and returns a {Smithy::Client::Identities::Http
...
+        # @api private
+        class Handler < Client::Handler
+          def call(context)
+            sign(context) if context.auth[:scheme_id] == 'smithy.api#httpApiKeyAuth'
+            @handler.call(context)
+          end
+
+          def sign(context) # rubocop:disable Metrics/AbcSize
+            reset(context)
+            request = context.http_request
```
mullermp: If you wanted to clean these up, you can just not assign these variables in reset and sign, since they are only referenced once at runtime (in each when case)

gems/smithy-client/lib/smithy-client/plugins/resolve_auth.rb
```diff
@@ -0,0 +1,87 @@
...
+          def resolve_auth(context, auth_options)
+            failures = []
+
+            raise 'No auth options were resolved' if auth_options.empty?
+
+            auth_options.each do |auth_option|
+              identity_provider_config_option = context.config.auth_schemes[auth_option]
+
+              unless identity_provider_config_option
+                failures << "Auth scheme #{auth_option} was not enabled " \
```
mullermp: You can just make all of these a single line string - this was when rubocop enforced 80 character lines, it's now 120.

gems/smithy-client/spec/smithy-client/plugins/checksum_required_spec.rb
```diff
@@ -43,13 +42,12 @@ module Plugins
         end
 
         it 'calculates checksums before signing' do
```
mullermp, line 43: Removing anonymous auth would break this - but I think it's probably better to wrap checksum required handler (the thing being tested) and checking that no Authorization header is set, and explicitly add bearer or something as an auth scheme here.

gems/smithy-client/spec/smithy-client/plugins/resolve_auth_spec.rb
```diff
@@ -0,0 +1,162 @@
...
+          expect(client.config).to respond_to(:auth_resolver)
+        end
+
+        it 'adds scheme ids to auth scheme hash' do
+          client_class.add_plugin(HttpApiKeyAuth)
+          client_class.add_plugin(HttpBasicAuth)
+          client_class.add_plugin(HttpBearerAuth)
+          client_class.add_plugin(HttpDigestAuth)
+          client
+          expect(client.config.auth_schemes['smithy.api#noAuth']).to equal(:anonymous_provider)
```
mullermp: Goes away

gems/smithy-client/spec/smithy-client/plugins/resolve_auth_spec.rb
```diff
@@ -0,0 +1,162 @@
...
+              resolved_auth
+            end
+            client.operation
+          end
+
+          it 'resolves auth for http api key auth' do
+            shapes['smithy.ruby.tests#SampleClient']['traits']['smithy.api#httpApiKeyAuth'] = {}
+            client_class.add_plugin(HttpApiKeyAuth)
+            handler = ResolveAuth::Handler
+            expect_any_instance_of(handler).to receive(:try_load_auth_scheme).and_wrap_original do |method, *args|
```
mullermp: Rather than do this - I think you just want to do something like
```
resp = client.operation
expect(resp.context.auth[:scheme_id]) ...
etc
```
This is more behavior driven and not implementation (handler wrapping)

gems/smithy-client/spec/smithy-client/plugins/resolve_auth_spec.rb
```diff
@@ -0,0 +1,162 @@
...
+              resolved_auth = method.call(*args)
+              expect(resolved_auth[:scheme_id]).to equal('smithy.api#httpDigestAuth')
+              expect(resolved_auth[:identity]).to be_a(Identities::HttpLogin)
+              resolved_auth
+            end
+            expect { client.operation }.to raise_error(NotImplementedError)
+          end
+
+          it 'resolves the first supported auth scheme' do
```
mullermp: Not clear in the test what is supposed to happen. You probably want an auth resolver to explicitly pick one over the other and ensure it's resolved that way. I think it's alphabetical by default but that's a code gen concern.

gems/smithy/lib/smithy/templates/client/client.erb
```diff
@@ -133,19 +133,14 @@ module <%= module_name %>
...
-<% end -%>
-      }
+      def errors_module
+        Errors
       end
 
       # @api private
-      def errors_module
-        Errors
+      def auth_resolver
```
mullermp: alphabetize these methods

gems/smithy/lib/smithy/views/client/client.rb
```diff
@@ -112,11 +107,6 @@ def option_default(option)
           default.gsub('<DEFAULT_PROTOCOL>', protocols.keys.first || 'nil')
         end
 
-        def weld_auth_schemes(welds)
-          weld_auth_schemes = welds.map(&:add_auth_schemes).reduce({}, :merge)
```
mullermp, line 116: We can remove this from Weld base class too.

gems/smithy/spec/interfaces/welds/auth_schemes_spec.rb
```diff
@@ -48,7 +48,7 @@ def remove_auth_schemes
       end
 
       it 'removes auth schemes from the client' do
-        expect(client.config.auth_schemes).not_to include('smithy.api#httpBearerAuth')
+        expect(client.config.auth_schemes).to include('smithy.api#httpBearerAuth')
```
mullermp: I don't think this test file makes too much sense now, since we no longer add auth schemes via code generation. This test would be wrong otherwise?

projections/shapes/lib/shapes/client.rb
```diff
@@ -57,11 +55,8 @@ class Client < Smithy::Client::Base
     #  When true, the request will sleep until there is sufficient client side capacity to retry
     #  the request. When false, the request will raise a `CapacityNotAvailableError` and will
     #  not retry instead of sleeping.
-    # @option options [ShapeService::AuthResolver] :auth_resolver
-    #  The auth resolver used to resolve authentication. Any object that responds to `#resolve(parameters)`.
-    # @option options [Hash] :auth_schemes
-    #  The auth schemes used to resolve authentication. The key is the scheme name as a String,
-    #  and the value is an initialized auth scheme class.
+    # @option options [#resolve(context)] :auth_resolver (<DEFAULT_AUTH_RESOLVER>)
```
mullermp: DEFAULT_AUTH_RESOLVER needs to be populated with a real value here. Check how this is done for DEFAULT_PROTOCOL, but we do a gsub.

projections/shapes/sig/shapes/client.rbs
```diff
@@ -4,8 +4,7 @@ module ShapeService
 
     def self.new: (
       ?adaptive_retry_wait_to_fill: bool,
-      ?auth_resolver: ShapeService::AuthResolver,
-      ?auth_schemes: Hash[String, Smithy::Client::AuthScheme],
+      ?auth_resolver: Smithy::Client::AuthResolver,
```
mullermp: This should be ShapeService::AuthResolver somehow. I don't believe we have an AuthResolver base class in Smithy Client.

mullermp (commented):
Nice! Great simplifications.

gems/smithy-client/lib/smithy-client/plugins/anonymous_auth.rb
richardwang1124: I swear previously when I tried to set the hash value to `options[:whatever_provider]` it would always return `nil` which is why I mapped it to the config symbol instead, but I think `client.config.whatever_provider` does actually give me the provider.

I can remove the plugin, but I'm assuming I still need to keep around some sort of mapping for noAuth just so that ResolveAuth still works. Otherwise how would it know if no auth options were resolved because the auth scheme was noAuth vs there were actually no resolved auth schemes?

gems/smithy-client/spec/smithy-client/plugins/resolve_auth_spec.rb
richardwang1124: See my previous comment but I think we might still need this. Or we could add like a special case in the ResolveAuth logic just for noAuth.

gems/smithy-client/spec/smithy-client/plugins/resolve_auth_spec.rb
richardwang1124: I was wondering how to access that context value, I didn't think to use `resp`'s context. Making tests more behavioral is something I need to keep working on.

gems/smithy-client/spec/smithy-client/plugins/resolve_auth_spec.rb
richardwang1124: I think I saw from the SEP that "A service shape can have any number of auth traits applied to it and an explicit priority order can be applied to the service or individual operations by using the [auth trait](https://smithy.io/2.0/spec/authentication-traits.html#auth-trait)." I wanted this case to test that ordering, but I think I'll need to update this test to modify the auth trait instead during setup.

gems/smithy/lib/smithy/views/client/client.rb
richardwang1124: I think we still need the `:add_auth_schemes` method for the AuthResolver. Although now that I think about it, shouldn't the AuthResolver's auth options be purely generated from the model? It shouldn't be dependent on whether or not the SDK supports the auth scheme right?

gems/smithy/spec/interfaces/welds/auth_schemes_spec.rb
richardwang1124: Yeah I don't think we need this whole test file anymore.

projections/shapes/lib/shapes/client.rb
richardwang1124: What real value populates DEFAULT_AUTH_RESOLVER? Isn't it a class? Also is `#resolve(context)` the right type for this config option?

gems/smithy-client/spec/smithy-client/plugins/resolve_auth_spec.rb
mullermp: Yes. My understanding is if auth trait is present, it makes that priority order. If not present, it's alphabetical.

gems/smithy-client/lib/smithy-client/plugins/resolve_auth.rb
```diff
@@ -0,0 +1,88 @@
...
+module Smithy
+  module Client
+    module Plugins
+      # @api private
+      class ResolveAuth < Plugin
+        option(
+          :auth_resolver,
+          doc_default: '<DEFAULT_AUTH_RESOLVER>',
+          doc_type: '#resolve(context)',
+          rbs_type: 'ShapeService::AuthResolver',
```
mullermp: I think doc_default will correctly populate this. If not, you'll have to do a similar replace.

gems/smithy-client/spec/smithy-client/plugins/checksum_required_spec.rb
```diff
@@ -43,13 +41,19 @@ module Plugins
...
-          signer = Signers::Anonymous.new
-          expect(signer).to receive(:sign).and_wrap_original do |method, context|
+          shapes['smithy.ruby.tests#SampleClient']['traits']['smithy.api#httpBearerAuth'] = {}
+          client_class.add_plugin(HttpBearerAuth)
+          checksum_handler = ChecksumRequired::Handler
+          expect_any_instance_of(checksum_handler).to receive(:call).and_wrap_original do |method, context|
+            expect(context.http_request.headers['Authorization']).to be_nil
+            method.call(context)
+          end
+          auth_handler = HttpBearerAuth::Handler
```
mullermp: I don't think you care to check the auth handler here. Checking `expect_any_instance_of(ChecksumRequired::Handler)` satisfies the test that authorization didn't happen by the time the handler is called.

gems/smithy-client/spec/smithy-client/plugins/resolve_auth_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+          client_class
+        end
+
+        let(:client) { client_class.new(stub_responses: true) }
+
+        it 'adds an :auth_resolver option to config' do
+          expect(client.config).to respond_to(:auth_resolver)
+        end
+
+        it 'adds scheme ids to auth scheme hash' do
```
mullermp: "auth scheme config hash"

gems/smithy-client/spec/smithy-client/plugins/resolve_auth_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+          client_class.add_plugin(HttpApiKeyAuth)
+          client_class.add_plugin(HttpBasicAuth)
+          client_class.add_plugin(HttpBearerAuth)
+          client_class.add_plugin(HttpDigestAuth)
+          client
```
mullermp: This is already implicitly referenced below - so you can remove this.

gems/smithy-client/spec/smithy-client/plugins/resolve_auth_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+          expect(client.config.auth_schemes['smithy.api#httpApiKeyAuth']).to be_a(HttpApiKeyProvider)
+          expect(client.config.auth_schemes['smithy.api#httpBasicAuth']).to be_a(HttpLoginProvider)
+          expect(client.config.auth_schemes['smithy.api#httpBearerAuth']).to be_a(HttpBearerProvider)
+          expect(client.config.auth_schemes['smithy.api#httpDigestAuth']).to be_a(HttpLoginProvider)
+        end
+
+        context 'resolving auth' do
```
mullermp: No need to make this and errors into contexts imo since there's no shared context between them.

gems/smithy-client/spec/smithy-client/plugins/resolve_auth_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+            expect(resp.context.auth[:scheme_id]).to equal('smithy.api#httpBasicAuth')
+            expect(resp.context.auth[:identity]).to be_a(Identities::HttpLogin)
+          end
+        end
+
+        context 'errors' do
+          it 'raises an error when no auth options were resolved' do
+            client_class.add_plugin(HttpApiKeyAuth)
+            handler = ResolveAuth::Handler
+            expect_any_instance_of(handler).to receive(:resolve_auth).and_wrap_original do |method, *args|
```
mullermp: What is args[1] here? I think a better way to do this is to clear out auth schemes config perhaps or remove from the model? (what are you trying to test here?)

gems/smithy/lib/smithy/views/client/client.rb
```diff
@@ -107,14 +102,15 @@ def option_tag(option)
...
           default = option.doc_default || option.default
-          return default unless option.name == :protocol
 
-          default.gsub('<DEFAULT_PROTOCOL>', protocols.keys.first || 'nil')
-        end
-
-        def weld_auth_schemes(welds)
-          weld_auth_schemes = welds.map(&:add_auth_schemes).reduce({}, :merge)
-          weld_auth_schemes.except(*welds.map(&:remove_auth_schemes).reduce([], :+))
+          if option.name == :protocol
```
mullermp: case statement

gems/smithy/lib/smithy/views/client/client.rb
```diff
@@ -107,14 +102,15 @@ def option_tag(option)
 
         def option_default(option)
           default = option.doc_default || option.default
```
mullermp: Not entirely related but I think actually we want to wrap default in strings if doc_type is a String or Symbol ? Maybe add a todo so we don't forget.

gems/smithy/lib/smithy/welds/auth/anonymous_auth.rb
```diff
@@ -11,20 +9,8 @@ def for?(_service)
...
-
       def add_auth_schemes
-        {
-          'smithy.api#noAuth' => {
-            auth_scheme_config_option: :anonymous_auth_scheme,
-            identity_provider_config_option: :anonymous_provider,
-            identity_type: Smithy::Client::Identities::Anonymous
-          }
-        }
+        ['smithy.api#noAuth']
```
mullermp: Is anonymous auth weld still necessary?

gems/smithy/spec/interfaces/welds/auth_schemes_spec.rb
```diff
@@ -39,16 +28,22 @@ def remove_auth_schemes
...
     context context do
-      include_context context, 'Weather'
+      include_context context, 'ServiceWithAuthTrait', fixture: 'auth/auth_trait'
 
-      let(:client) { Weather::Client.new }
+      let(:client) { ServiceWithAuthTrait::Client.new }
 
       it 'adds auth schemes to the client' do
-        expect(client.config.auth_schemes).to include('smithy.api#httpBasicAuth')
+        auth_resolver = client.class.auth_resolver.new
```
mullermp: We know this already - it's ServiceWithAuthTrait::AuthResolver and ServiceWithAuthTrait::AuthParameters

gems/smithy-client/spec/smithy-client/plugins/resolve_auth_spec.rb
richardwang1124: I actually don't even know if this is possible, but it seems like the error check within the resolve auth plugin is checking to see if AuthResolver returns no auth options. I think even if no auth schemes or traits were added it just default to noAuth right? But this test is just mocking the return value of AuthResolver.resolve to be [].

gems/smithy/lib/smithy/welds/auth/anonymous_auth.rb
richardwang1124: I don't think so since we added that special case. I'll remove this.

gems/smithy-client/spec/smithy-client/plugins/resolve_auth_spec.rb
mullermp: I pushed a fix.

.github/workflows/ci.yml
```diff
@@ -8,6 +8,7 @@ on:
   pull_request:
     branches:
       - decaf
+      - signer
```
mullermp, line 11: Can remove this in the next PR.

mullermp (approved):
Nice!

PR #310, opened 2025-07-03
User Agent

richardwang1124:
Add user agent plugin and test.


gems/smithy-client/spec/smithy-client/plugins/user_agent_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+          client_class.add_plugin(UserAgent)
+          client_class
+        end
+
+        let(:client) { client_class.new }
+
+        it 'adds user agent header to request' do
+          resp = client.operation
+          ua_header = resp.context.http_request.headers['user-agent']
+          expect(ua_header).to_not be_nil
```
richardwang1124, line 15: Not really sure the best way to test this since the metadata would be different depending on the os/environment/etc. I could mock the values, but would that defeat the purpose of this test?

gems/smithy-client/lib/smithy-client/plugins/user_agent.rb
```diff
@@ -0,0 +1,59 @@
...
+    module Plugins
+      class UserAgent < Plugin
+        class Handler < Client::Handler
+          def call(context)
+            set_user_agent(context)
+            @handler.call(context)
+          end
+
+          def set_user_agent(context)
+            context.http_request.headers['User-Agent'] = UserAgent.new(context).to_s
```
mullermp, line 20: I think to_s is called implicitly by setting of header. You can verify.

gems/smithy-client/lib/smithy-client/plugins/user_agent.rb
```diff
@@ -0,0 +1,59 @@
...
+          end
+
+          class UserAgent
+            def initialize(context)
+              @context = context
+            end
+
+            def to_s
+              ua = "smithy-ruby/#{File.read(File.expand_path('../../../VERSION', __dir__)).strip}"
```
mullermp: There should be a constant for this .. Smithy::Client::VERSION if it isn't already there.

gems/smithy-client/lib/smithy-client/plugins/user_agent.rb
```diff
@@ -0,0 +1,59 @@
...
+                  'other'
+                end
+              metadata = "os/#{os}"
+              local_version = Gem::Platform.local.version
+              metadata += "##{local_version}" if local_version
+              metadata + " md/#{RbConfig::CONFIG['host_cpu']}"
+            end
+
+            def language_metadata
+              "lang/#{RUBY_ENGINE}##{RUBY_ENGINE_VERSION} md/#{RUBY_VERSION}"
```
mullermp: This format is still a bit AWS-like. Not saying we shouldn't do this, but we could look at other tools for inspiration.

gems/smithy-client/spec/smithy-client/plugins/user_agent_spec.rb
mullermp: Yeah. I think you can either:
1) mimic implementation in test - not great but works on all platforms
2) quick sanity test. Is os/<something> set, Is lang/<something rubyish> set etc

Probably option 2 is fine.

gems/smithy-client/lib/smithy-client/plugins/user_agent.rb
richardwang1124: Although it's true that `to_s` is called for `Headers`, this `to_s` method is for constructing the user agent string, which is defined below.

mullermp: In headers, the assignment method is:

```
        def []=(key, value)
          @data[key.to_s.downcase] = value.to_s
        end
```
which should call to_s already on UserAgent.new(...)

richardwang1124: Oh I see what you mean. Yeah then I think I can remove this.

gems/smithy-client/lib/smithy-client/plugins/user_agent.rb
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    module Plugins
+      class UserAgent < Plugin
+        option(:user_agent_suffix)
```
mullermp: In v3 this was secret. I think we should fully document this as a public option in the generic client. In v4 we can keep it secret or remove it. Or, we can use app id concept instead. Let's discuss

gems/smithy-client/lib/smithy-client/plugins/user_agent.rb
```diff
@@ -0,0 +1,62 @@
...
+
+        class Handler < Client::Handler
+          def call(context)
+            set_user_agent(context)
+            @handler.call(context)
+          end
+
+          def set_user_agent(context)
```
mullermp: This method is kinda useless.

gems/smithy-client/spec/smithy-client/plugins/user_agent_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+
+require 'smithy-client/plugins/user_agent'
+
+module Smithy
+  module Client
+    module Plugins
+      describe UserAgent do
+        let(:client_class) do
+          client_class = ClientHelper.sample_client.const_get(:Client)
+          client_class.clear_plugins
```
mullermp: With my PR I stopped doing this. You can just get client class normally.

gems/smithy-client/spec/smithy-client/plugins/user_agent_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+          client_class.add_plugin(DummySendPlugin)
+          client_class.add_plugin(UserAgent)
+          client_class
+        end
+
+        let(:client) { client_class.new }
+
+        it 'adds user agent header to request' do
```
mullermp: Adds a

gems/smithy-client/spec/smithy-client/plugins/user_agent_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+          resp = client.operation
+          ua_header = resp.context.http_request.headers['user-agent']
+          expect(ua_header).to_not be_nil
+          expect(ua_header).to include('smithy-ruby')
+          expect(ua_header).to include('macos').or include('linux').or include('windows').or include('other')
+          expect(ua_header).to include('ruby')
+        end
+
+        it 'adds user agent suffix to user agent string when configured' do
```
mullermp: Adds a

mullermp (approved):

gems/smithy-client/spec/smithy-client/plugins/user_agent_spec.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+require_relative '../../spec_helper'
+
+require 'smithy-client/plugins/user_agent'
```
mullermp: Require is not necessary anymore

gems/smithy-client/lib/smithy-client/plugins/user_agent.rb
```diff
@@ -0,0 +1,65 @@
...
+
+module Smithy
+  module Client
+    module Plugins
+      # @api private
+      class UserAgent < Plugin
+        option(
+          :user_agent_suffix,
+          doc_type: String,
+          docstring: 'The suffix appended to the user agent string.'
```
mullermp: String -> header

gems/smithy-client/lib/smithy-client/pageable_response.rb
```diff
@@ -133,6 +133,14 @@ def next_page_params(params)
...
         new_params.merge!(@paginator.next_tokens(data).merge(params))
       end
+
+      def with_metric(&block)
+        Thread.current[:aws_sdk_core_user_agent_metric] ||= []
+        Thread.current[:aws_sdk_core_user_agent_metric] << 'C'
+        block.call
+      ensure
+        Thread.current[:aws_sdk_core_user_agent_metric].pop
+      end
```
jterapin: Should we put a TODO to revisit this approach? Just want to make sure that this does not go amiss

PR #311, opened 2025-07-08
Use plugins per protocol approach

mullermp:
Went back and forth on this a lot over the past few months. This approach is to just use one protocol per plugin.

Main advantages:
* protocols can set their own configs onto the client, like for codecs, simple json, or even stubbing.
* no special casing of welds - protocols are just plugins. Ordering still is necessary at the weld level.
* Protocol handlers act on listeners (over the wire) instead of on parsing a body into a string io and then parsing into a data structure.

Quirks (maybe or maybe cons):
* no protocol config option - protocols can be removed or swapped out at the client class level like any other plugin
* new protocols are also plugin/handler based


gems/smithy/lib/smithy/welds/default_plugins.rb
```diff
@@ -40,7 +39,6 @@ def add_plugins
           Smithy::Client::Plugins::PageableResponse => { require_path: "#{base_path}/pageable_response" },
           Smithy::Client::Plugins::ParamConverter => { require_path: "#{base_path}/param_converter" },
           Smithy::Client::Plugins::ParamValidator => { require_path: "#{base_path}/param_validator" },
-          Smithy::Client::Plugins::Protocol => { require_path: "#{base_path}/protocol" },
           Smithy::Client::Plugins::RaiseResponseErrors => { require_path: "#{base_path}/raise_response_errors" },
           Smithy::Client::Plugins::RequestCompression => { require_path: "#{base_path}/request_compression" },
```
jterapin, line 44: Unrelated to this PR: Is there a reason why request compression is considered a default plugin? I think it is relatively useless unless one of the operations has a `requestCompression` trait

gems/smithy/spec/interfaces/welds/protocols_spec.rb
```diff
@@ -1,53 +0,0 @@
-# frozen_string_literal: true
-
-require_relative '../../spec_helper'
-
-describe 'Welds: Protocols' do
```
jterapin, line 5: I'm not sure if there's a good place for this but we should probably test that plugins are added when service model has the protocol trait defined?

gems/smithy-client/lib/smithy-client/plugins/rpc_v2_cbor.rb
```diff
@@ -0,0 +1,21 @@
...
+module Smithy
+  module Client
+    module Plugins
+      # @api private
+      class RpcV2Cbor < Plugin
+        # @api private
+        option(:cbor_codec) { CBOR::Codec.new }
+
+        # @api private
+        option(:stubber) { Stubbing::RpcV2Cbor.new }
```
jterapin, line 15: Does this override other protocol plugins? or do we have to do something like `:cbor_stubber`? I'm just assuming here that we are going with loading all the protocol plugins. If not, disregard.

gems/smithy-client/lib/smithy-client/plugins/rpc_v2_cbor.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+require_relative '../rpc_v2_cbor'
+
+module Smithy
+  module Client
+    module Plugins
+      # @api private
+      class RpcV2Cbor < Plugin
```
jterapin, line 11: (not sure where to put this so leaving comments here)

What is the CX to swap protocols?  How do they fall back on the "default" protocol? You should include these examples in the PR description.  If we are going with this approach, we should have a plan outlined to accommodate it.

I don't recall if we plan to load all the supported protocol plugins from the get-go or just have customers `.remove_plugin` and `.add_plugin` ....

gems/smithy/lib/smithy/welds/protocols/rpc_v2_cbor.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+require 'smithy-client/plugins/rpc_v2_cbor'
+
+module Smithy
+  module Welds
+    module Protocols
+      # Adds the rpcv2Cbor protocol to the client if the service has the trait.
+      class RpcV2Cbor < Weld
```
jterapin: I'm guessing if a service supports multiple protocols - are all those protocols brought in? I wasn't sure if below approach has changed or not
```
module Weather
     ....
     add_plugin(Smithy::Client::Plugins::Rpcv2Cbor)
     add_plugin(Smithy::Client::Plugins::RestJson)
    ...
```

jterapin: Also - based on PR description, the following snippet is noted as a pro:
> no special casing of welds - protocols are just plugins. Ordering still is necessary at the weld level.

It seems like we still need this cbor-specific weld to add to the client so not sure if this could be considered a main advantage unless there's another route that you were planning to do

gems/smithy-client/lib/smithy-client/plugins/rpc_v2_cbor.rb
jterapin: Also - I think you shared this offline but PR description should include ways to mitigate situations where we DO need to make changes to the response body prior to parsing. For historical context.

gems/smithy/lib/smithy/welds/default_plugins.rb
mullermp: Yes, we could consider adding it as a plugin conditionally to services that have the trait applied on an operation. We could also add the handler per operation. We could otherwise keep it globally and ensure a constant lookup, which would be at most a single method call.

gems/smithy/spec/interfaces/welds/protocols_spec.rb
mullermp: That's already done by protocol tests.

gems/smithy-client/lib/smithy-client/plugins/rpc_v2_cbor.rb
mullermp: The idea would be the default/winning protocol would always define a stubber config that knows how to stub. I am open to other ideas. In v3 we had all of the classes in a switch case but requires core/client to know about all protocols.

gems/smithy-client/lib/smithy-client/plugins/rpc_v2_cbor.rb
mullermp: The CX is to do add plugin and remove plugin. I'm also extremely skeptical that customers will care about this functionality at all as a normal client configurable, so I think any "friction" is not a huge concern for us. From my experience, customers just want the service to function, and we should by default use something standard and performant. I don't think we care to really outline how to do all this unless there is reason to, like with an operational event.

gems/smithy/lib/smithy/welds/protocols/rpc_v2_cbor.rb
mullermp: In code generation, I think if multiple protocols are present, we will just select the first one that is preferred. If the non default one is preferred, a swap is needed, or simply an add (last one wins anyway). For generic smithy ruby, there is one protocol, soon more (rails JSON and xml). I will probably make a weld that understands all 3 and picks one. For AWS, we will remove the cbor protocol always, and then potentially add it back in the priority list in a weld that has all 7 protocols in it.

gems/smithy/spec/interfaces/welds/protocols_spec.rb
jterapin: That's fair - given we are going back to 1 protocol plugin added at default.

jterapin (approved):
Our team has decided to revert to the V3 design of protocol plugins, establishing a default plugin based on priority list. We also have clear pathways for protocol switching (if needed) and the ability to modify body through handlers/listeners.

PR #312, opened 2025-07-11
Feature ids

richardwang1124:
Add features module to capture generic feature ids.


gems/smithy-client/lib/smithy-client/plugins/protocol.rb
```diff
@@ -16,7 +16,18 @@ class Protocol < Plugin
...
+          end
+
+          def with_metric(protocol, &block)
+            metric = protocol.is_a?(Smithy::Client::RPCv2CBOR::Protocol) ? 'M' : nil
+            return block.call unless metric
+
+            Thread.current[:smithy_ruby_user_agent_metric] ||= []
+            Thread.current[:smithy_ruby_user_agent_metric] << metric
+            block.call
+          ensure
```
richardwang1124: I copied over the ensure from our V3 code, but is it necessary? I think in this particular case it's actually incorrect, since if the protocol isn't CBOR it would just call the block and return but the ensure results in popping from `Thread.current[:smithy_ruby_user_agent_metric]` which may not be defined.

gems/smithy-client/lib/smithy-client/pageable_response.rb
```diff
@@ -133,6 +133,14 @@ def next_page_params(params)
         new_params = context.params.except(*prev_tokens)
         new_params.merge!(@paginator.next_tokens(data).merge(params))
       end
+
+      def with_metric(&block)
+        Thread.current[:aws_sdk_core_user_agent_metric] ||= []
```
mullermp: Discussed offline, but we should evaluate two approaches (or combined) - context/anonymous handler and thread local variable.

gems/smithy-client/lib/smithy-client/plugins/protocol.rb
mullermp: Ensure is necessary but I think the structure of this is wrong and creates that problem.

gems/smithy-client/lib/smithy-client/waiters/poller.rb
```diff
@@ -87,6 +87,14 @@ def anyStringEquals?(actual, expected)
           actual.any? { |value| value == expected }
         end
         # rubocop:enable Naming/MethodName
+
+        def with_metric (&block)
+          Thread.current[:smithy_ruby_user_agent_metric] ||= []
+          Thread.current[:smithy_ruby_user_agent_metric] << 'B'
```
mullermp: We should use string identifiers here.

gems/smithy-client/lib/smithy-client/features.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    # @api private
+    module Features
+      def self.clear
```
mullermp: I don't see this method used anywhere. We might be able to default it to an empty list when Features is required/loaded (at the top of the module).

gems/smithy-client/lib/smithy-client/features.rb
```diff
@@ -0,0 +1,20 @@
...
+        Thread.current[:smithy_ruby_user_agent_metric] = []
```
mullermp: Abstract away the naming from user agent. That's how it's used in v4. It's just a smithy ruby feature or metric. Not sure about the naming.

gems/smithy-client/lib/smithy-client/features.rb
```diff
@@ -0,0 +1,20 @@
...
+      end
+
+      def self.with_metric(*metrics, &block)
+        Thread.current[:smithy_ruby_user_agent_metric] ||= []
+        Thread.current[:smithy_ruby_user_agent_metric].concat(metrics)
+        block.call
+      ensure
+        Thread.current[:smithy_ruby_user_agent_metric].pop(metrics.size)
```
mullermp: You probably will need to add test cases for this module and method. We will want to ensure that metrics are added and removed in a stack like fashion, and probably also ensure it doesn't affect multi threading cases. This would be a good exercise for you to understand threads.

gems/smithy-client/lib/smithy-client/plugins/request_compression.rb
```diff
@@ -149,6 +149,14 @@ def update_content_encoding(encoding, context)
             end
           end
 
+          def with_metric(encoding, &block)
+            if encoding == 'gzip'
+              Features.with_metric('GZIP_REQUEST_COMPRESSION', &block)
```
mullermp: Alternative naming might be Metrics.track {} or Features.track {}. It depends if we want to refer to these as features or metrics.

We also should look at observability requirements and whether features make sense to track that way.

gems/smithy-client/lib/smithy-client/features.rb
richardwang1124: Added to `request.new`

gems/smithy-client/spec/smithy-client/features_spec.rb
```diff
@@ -0,0 +1,81 @@
+# frozen_string_literal: true
+
+require_relative '../spec_helper'
+
+module Smithy
+  module Client
+    describe Features do
+      describe '#clear' do
+        it 'clears features' do
+          Thread.current[:smithy_ruby_features] = ['test']
```
mullermp: implementation detail - You could do maybe

```
Features.track('A') do
  Features.clear
  expect() to be empty
end
```

gems/smithy-client/spec/smithy-client/features_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+          Features.clear
+          expect(Features.list).to be_empty
+        end
+      end
+
+      describe '#track' do
+        before(:each) { Features.clear }
+
+        it 'tracks and removes a feature' do
+          proc = proc { expect(Features.list).to eq(%w[A]) }
```
mullermp: What's with the procs? I don't think they are necessary.

gems/smithy-client/lib/smithy-client/features.rb
```diff
@@ -0,0 +1,23 @@
...
+      end
+
+      def self.track(*metrics, &block)
+        Thread.current[:smithy_ruby_features].concat(metrics)
+        block.call
+      ensure
+        Thread.current[:smithy_ruby_features].pop(metrics.size)
+      end
+
+      def self.list
```
mullermp: list is probably an ok name, but maybe "tracked"

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
```diff
@@ -117,6 +117,14 @@ def reset_response(context, response)
             context.http_response.reset
             response.error = nil
           end
+
+          def with_metric(retry_strategy, &block)
+            case retry_strategy
+            when Retry::Standard then Features.track('RETRY_MODE_STANDARD') { block.call }
```
mullermp: You can pass &block to these, or anonymous & (in all of the cases you have)

gems/smithy-client/spec/smithy-client/features_spec.rb
```diff
@@ -0,0 +1,88 @@
+# frozen_string_literal: true
+
+require_relative '../spec_helper'
+
+module Smithy
+  module Client
+    describe Features do
+      describe '#clear' do
+        it 'clears features' do
+          Thread.current[:smithy_ruby_features] = %w[A]
```
mullermp: This is an implementation detail. I think this is better:
```
Features.track('A') do
  Features.clear
  expect empty
end
```

mullermp (approved):

gems/smithy-client/lib/smithy-client/features.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    # @api private
+    module Features
```
jterapin, line 6: Since there's multiple module methods defined here - we could use `class << self` syntax than doing having to do `self.` whenever you have to defined a method.

gems/smithy-client/lib/smithy-client/features.rb
```diff
@@ -0,0 +1,23 @@
...
+      def self.clear
+        Thread.current[:smithy_ruby_features] = []
+      end
+
+      def self.track(*metrics, &block)
```
jterapin: I don't recall where you and Matt landed on this metrics vs features conversation but could we call these features? Metrics is an overloaded term but when I think of metrics - I think it is "measuring" some kind of value than tracking. 

Down the road - when we implement observability metrics - we are going to be using the term, "Metrics" quite heavily.

gems/smithy-client/lib/smithy-client/features.rb
jterapin: agreed with above

gems/smithy-client/lib/smithy-client/rpc_v2_cbor/handler.rb
```diff
@@ -7,7 +7,7 @@ module RpcV2Cbor
       class Handler < Client::Handler
         def call(context)
           build_request(context)
-          response = @handler.call(context)
+          response = Features.track('PROTOCOL_RPC_V2_CBOR') { @handler.call(context) }
```
jterapin: Friendly reminder that whenever we do the aws-variation of the plugin, we will need to bring this.

jterapin (approved):
Nice! Great job on this side quest

gems/smithy-client/lib/smithy-client/rpc_v2_cbor/handler.rb
mullermp: Good point - we can probably not bother tracking this here and delete it.

gems/smithy-client/lib/smithy-client/plugins/request_compression.rb
```diff
@@ -149,6 +149,14 @@ def update_content_encoding(encoding, context)
             end
           end
 
+          def track_feature(encoding, &block)
+            if encoding == 'gzip'
```
mullermp: `case when` would be appreciated here.

mullermp (approved):
Awesome!! Super clean

PR #313, opened 2025-07-12
Improve test building and running

mullermp:
List of changes:
- Add model/skip_tests to gem files so dependencies can grab it
- Use --aut to get minified json but only after validating using smithy build config.
- Move protocol tests out of fixtures and use `smithy build` to fetch contents automatically
- Use skip_tests from smithy gem model folder
- Use smithy-build.json for source of truth for protocol tests
- endpoint_tests -> endpoints in fixtures


gems/smithy-client/lib/smithy-client/protocol_spec_matcher.rb
```diff
@@ -13,7 +13,7 @@
 
     def match_hash(actual, expected)
       expected.each do |key, value|
-        expect(actual).to include(key)
+        expect(actual).not_to include(key) unless expected.key?(key)
```
richardwang1124: I might be misunderstanding, but when would `expect(actual).not_to include(key)` ever be evaluated? Shouldn't `expected.key?(key)` always be true since we're looping through each key, value pair in `expected`?

tasks/smithy.rake
```diff
@@ -5,88 +5,56 @@ require 'rspec/core/rake_task'
...
-    include_paths = ['gems/smithy/spec/support/matchers']
     sig_paths = %w[gems/smithy-client/sig gems/smithy-schema/sig]
     rbs_targets = %w[Smithy Smithy::* Smithy::Client Smithy::Schema Smithy::Client::* Smithy::Schema::*]
-    Dir.glob("gems/smithy/spec/fixtures/#{suite}/*/model.json") do |model_path|
+    Dir.glob("gems/smithy/spec/#{suite}/*/model.json") do |model_path|
       test_name = model_path.split('/')[-2]
       test_module = test_name.gsub('-', '').camelize
-      plan = SpecHelper.generate_gem(test_module, :client, fixture: "#{suite}/#{test_name}")
+      model = JSON.load_file(model_path)
+      plan = SpecHelper.generate_gem(test_module, :client, model: model)
```
jterapin:
```suggestion
      plan = SpecHelper.generate_gem(test_module, :client, model: JSON.load_file(model_path))
```

tasks/smithy.rake
```diff
@@ -100,19 +68,63 @@ namespace :smithy do
...
+    end
+    if failures.any?
+      puts 'Fixture models out of sync:'
+      failures.each { |m| puts "\t#{m}" }
+      raise 'Fixture models are out of sync. Run `bundle exec rake smithy:sync-fixtures` to correct.'
+    end
+  end
+
+  desc 'Build the upstream protocol tests and copy the source JSON to the test folder'
+  task 'sync-protocol-tests' do
```
jterapin, line 96: Do we want to include syncing of protocol tests within our CI workflow?

jterapin (approved):
Nice improvements.

richardwang1124 (approved):
Nice work, left a question. Also, does the README need to be updated?

tasks/smithy.rake
mullermp: I thought about this. Maybe. They shouldn't change since we are fixed to a smithy version, but they would when we upgrade. We might want to guard against that.

gems/smithy-client/lib/smithy-client/protocol_spec_matcher.rb
mullermp: Good catch - I think I also don't need this anymore?

PR #314, opened 2025-07-15
Add name field to shapes

mullermp:
Add name field to shapes. Some protocols do matching off name, so this is more convenience. Also makes every property of shape an accessor.


richardwang1124 (approved):
LGTM

PR #315, opened 2025-07-16
Support vendor params in protocol tests

mullermp:
Support vendor params hook in welds for protocol tests. These are needed for additional assertions.

Usage might be like:
```
      def protocol_test_vendor_code
        {
          'aws.protocoltests.config#ErrorCodeParams' => ErrorCodeParams.new
        }
      end

      # Used for generating error code expectations in protocol tests.
      class ErrorCodeParams
        def error_expect_code(params)
          <<~CODE
            expect(e.code).to eq('#{params['code']}')
            # except(e.type).to eq('#{params['type']}')
          CODE
        end
      end
```


jterapin (approved):
LGTM

PR #316, opened 2025-07-17
Do not override message errors and use magic message or default instead

mullermp:
This change removes the 'message' override for generated service errors and instead relies on the modeled data message if present or the default of the class name. The parsing of message in protocol handlers appeared to be a legacy consideration before modeled error messages. Also fixes rpcv2 error parsing correctness.


gems/smithy-client/lib/smithy-client/rpc_v2_cbor/error_handler.rb
```diff
@@ -10,9 +10,9 @@ def call(context)
           # 200 range for error handling only for this case.
           @handler.call(context).on_done(200..599) do |response|
             if !valid_response?(context)
-              code, message, data = http_status_error(context)
-              response.error = build_error(context, code, message, data)
-            elsif (300..599).cover?(context.http_response.status_code)
+              code, data = http_status_error(context)
+              response.error = build_error(context, code, data)
+            elsif (400..599).cover?(context.http_response.status_code)
```
richardwang1124, line 15: Was this previously incorrect? 300 status codes are for redirects right?

gems/smithy-client/lib/smithy-client/service_error.rb
```diff
@@ -8,14 +8,13 @@ module Client
...
       # @param [Structure] data
-      def initialize(context, message, data)
+      def initialize(context, data)
         @code = self.class.code
         @context = context
-        @message = extract_message(message, data)
         @data = data
-        super(@message)
+        message = data.message if data.respond_to?(:message)
+        super(message)
```
richardwang1124, line 17: Is this how messages are defaulted to class name? I see that the `Exception` class' `message` method returns either the provided message or name of the class.

gems/smithy-client/lib/smithy-client/rpc_v2_cbor/error_handler.rb
```diff
@@ -26,60 +26,52 @@ def valid_response?(context)
           req_header == resp_header
         end
 
-        # TODO: Fix this
-        # This is not correct per protocol tests. Some headers will determine the error code.
-        # If the body is empty, there is still potentially an error code from the header, but
-        # we are making a generic http status error instead. In a new major version, we should
-        # always try to extract header, and during extraction, check headers and body.
```
richardwang1124, line 33: Did we fix this? I don't see us checking the headers.

gems/smithy-client/lib/smithy-client/rpc_v2_cbor/error_handler.rb
mullermp: Yeah. The specification says 4xx and 5xx. It was probably benign but it should never have 3xx.

gems/smithy-client/lib/smithy-client/service_error.rb
mullermp: It's defaulted to class name as part of RuntimeError which this inherits from. `super(nil)` will have that behavior.

gems/smithy-client/lib/smithy-client/rpc_v2_cbor/error_handler.rb
mullermp: This was copied from AWS JSON. For cbor, the error is only in the body, so this is fine.

jterapin (approved):

PR #318, opened 2025-07-25
Smithy XML gem

mullermp:
Adds the Smithy XML gem for generic XML parsing. It has the same 5 engines as SDK v3. It currently passes AwsQuery protocol tests.


gems/smithy-client/sig/smithy-client/interfaces.rbs
```diff
@@ -2,10 +2,6 @@ module Smithy
   module Client
     type endpoint_url = String | URI::HTTP | URI::HTTPS
 
-    interface _IdentityProvider
```
richardwang1124: Do we not need this anymore?

gems/smithy-xml/lib/smithy-xml/parser/frame.rb
```diff
@@ -0,0 +1,341 @@
...
+            #   UnknownMemberFrame.new(xml_name, self, nil, @result)
+          else
+            NullFrame.new(xml_name, self)
+          end
+        end
+
+        def consume_child_frame(child) # rubocop:disable Metrics/AbcSize
+          case child
+          when MapEntryFrame
+            @result[@member[:name]] ||= {}
```
richardwang1124, line 256: I noticed that the V3 code does not have this. Should we fix this in V3? Not sure how possible/likely that `@result[@member[:name]]` would be `nil`.

gems/smithy/spec/spec_helper.rb
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require 'byebug'
```
richardwang1124, line 3: Is this needed?

richardwang1124 (commented):
Nice work, left a few questions. Most changes are very similar to V3.

gems/smithy-client/sig/smithy-client/interfaces.rbs
mullermp: It wasn't used anywhere. But I think you'll need it and we should have used it in here somewhere.

gems/smithy-xml/lib/smithy-xml/parser/frame.rb
mullermp: In the v3 code we were using DefaultList and DefaultMap. I think this was wrong because even when there weren't any values, we would parse empty lists and maps. There is a patch in the json parser in v3 that has to maintain this behavior from query to json compatibility. I'm not 100% but this should have prevented this, I will double check.

gems/smithy/spec/spec_helper.rb
mullermp: I think I'd like to add these in our spec helpers so that I don't have to require it every time I want to byebug.

gems/smithy-xml/lib/smithy-xml/codec.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Xml
+    # @api private
+    class Codec
+      # @param [Hash] options
+      def initialize(options = {})
+        @options = options
```
jterapin, line 9: Similar to `json_name` setting - I'm thinking `xml_name` could be a possible setting.

gems/smithy-xml/lib/smithy-xml/builder.rb
```diff
@@ -0,0 +1,148 @@
...
+        target ||= []
+        @builder = DocBuilder.new(target: target, indent: @indent, pad: @pad)
+        structure(location_name(ref, ref.member_name), ref, data)
+        target.join
+      end
+
+      private
+
+      def shape(name, ref, value)
+        case ref.shape
```
jterapin, line 27: Do we need to cover Boolean shape here? Serialized as "true" or "false" according to smithy docs

gems/smithy-xml/lib/smithy-xml/parser/frame.rb
```diff
@@ -0,0 +1,341 @@
...
+          @result << child.result unless child.is_a?(NullFrame)
+        end
+      end
+
+      # @api private
+      class MapEntryFrame < Frame
+        def initialize(xml_name, *args)
+          super
+          @key_name = @ref.shape.key.traits['smithy.api#xmlName'] || 'key'
+          @key = Frame.new(xml_name, self, @ref.shape.key)
```
jterapin, line 171: Supposed to be StringFrame here per documenation?

gems/smithy/spec/spec_helper.rb
jterapin: Yeah, I'd like that.

gems/smithy-xml/spec/spec_helper.rb
```diff
@@ -0,0 +1,113 @@
...
+SimpleCov.start do
+  add_filter '/spec/'
+  add_filter 'gems/smithy/'
+  add_filter 'gems/smithy-cbor/'
+  add_filter 'gems/smithy-client/'
+  add_filter 'gems/smithy-json/'
+  add_filter 'gems/smithy-schema/'
+end
+
+require 'smithy-xml'
```
jterapin, line 13: Most of the actual testing will be protocol tests but probably should cover similar test cases that we have in place for smithy-json.

gems/smithy-xml/lib/smithy-xml/parser.rb
```diff
@@ -0,0 +1,109 @@
...
+          @column = column
+          super(msg)
+        end
+
+        # @return [Integer, nil]
+        attr_reader :line
+
+        # @return [Integer, nil]
+        attr_reader :column
+      end
```
jterapin: We should probably move where we set the smithy-json's `ParserError` to keep these gems in a similar format if possible - https://github.com/smithy-lang/smithy-ruby/blob/decaf/gems/smithy-json/lib/smithy-json.rb#L13

gems/smithy-xml/lib/smithy-xml/parser.rb
```diff
@@ -0,0 +1,109 @@
...
+
+      def initialize(options = {})
+        @engine = options[:engine] || self.class.engine
+      end
```
jterapin, line 12: In smithy-json, engines are defined within `Smithy::Json` namespace - any strong preference between two setups? I guess it actually makes more sense to be under Parser since it's really used for those intentions

jterapin (approved):
Nice!

gems/smithy-xml/lib/smithy-xml/parser.rb
mullermp: Yeah. I don't think we necessarily have to follow symmetry here. The engine is for parsing only, and I think we could eventually promote it to be for building too if we wanted to, but I don't think we should let a builder handle this, at least from the client pov. For servers, possibly.

gems/smithy-xml/lib/smithy-xml/parser.rb
mullermp: Yes, I can move this one up.

gems/smithy-xml/spec/spec_helper.rb
mullermp: Yes. I have a worklog where I've added smithy-xml unit tests. I think it makes sense to do this alongside rest-xml because that includes the building portion (query/ec2 has xml only on output).

gems/smithy-xml/lib/smithy-xml/parser/frame.rb
mullermp: Yes. It ends up being a StringFrame. Frame.new will look at the shape and make the appropriate class.

gems/smithy-xml/lib/smithy-xml/builder.rb
mullermp: Yes, I believe it's handled by the default case `node(name, ref, value.to_s)`. The rest xml tests should exercise this for building.

gems/smithy-xml/lib/smithy-xml/codec.rb
mullermp: Yes, but I don't think we need it right now. For rest json and json rpc we do need that distinction.

PR #319, opened 2025-07-30
Add IdentityProvider

richardwang1124:
Add IdentityProvider module.

### DIFF
--- a/gems/smithy-client/lib/smithy-client/identity_resolver.rb
+++ b/gems/smithy-client/lib/smithy-client/identity_resolver.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    module IdentityResolver
+      # @return [Identity]
+      attr_reader :identity
+
+      # @return [Boolean]
+      def set?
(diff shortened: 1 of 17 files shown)


gems/smithy-client/lib/smithy-client/identity_resolver.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    # This module provides basic accessors and methods for an
+    # Identity Resolver class which resolves an Identity.
+    module IdentityResolver
```
mullermp: You should include this in the existing identity providers in this package. (Also should this be called Identity Provider?)

Can you also add rbs for this and the other providers in here?

richardwang1124: I think strictly following SRA it would be IdentityResolver, but I think IdentityProvider is fine. Also the SRA uses IdentityProvider in places too, not very consistent.

mullermp (approved):
You can remove the _IdentityProvider interface too.

PR #322, opened 2025-08-11
Rest xml

mullermp:
Changes necessary for Rest XML in v4.

Also change member_name to location_name, because we now have to add the location name for top level input/output for xml serialization, and previous name didn't make sense.


gems/smithy-xml/spec/smithy-xml/parser_spec.rb
```diff
@@ -0,0 +1,281 @@
...
+              <timestamp>#{time.to_i}</timestamp>
+              <union>
+                <string>string</string>
+              </union>
+            </Structure>
+          XML
+          expect(subject.parse(structure_shape, bytes).to_h).to eq(expected)
+        end
+
+        it 'parses structures with jsonName' do
```
jterapin:
```suggestion
        it 'parses structures with xmlName' do
```

gems/smithy/spec/smithy/model/shape_spec.rb
```diff
@@ -43,13 +43,13 @@ module Model
         end
       end
 
-      describe '.member_name' do
+      describe '.location_name' do
         it 'returns the member name' do
```
jterapin:
```suggestion
        it 'returns the location name' do
```

gems/smithy/spec/smithy/model/shape_spec.rb
```diff
@@ -43,13 +43,13 @@ module Model
...
-          expect(Shape.member_name(absolute_id)).to eq(member_name)
+          expect(Shape.location_name(absolute_id)).to eq(member_name)
         end
 
         it 'returns nil when there is no member name' do
```
jterapin:
```suggestion
        it 'returns nil when there is no location name' do
```

gems/smithy/lib/smithy/model/shape.rb
```diff
@@ -39,7 +39,7 @@ def self.name(id)
       # Optional member name
       # @param [String] id Shape ID
       # @return [String, nil] Member name
```
jterapin:
```suggestion
      # Optional location name
      # @param [String] id Shape ID
      # @return [String, nil] location name
```

jterapin (approved):
LGTM

gems/smithy-xml/lib/smithy-xml/builder.rb
```diff
@@ -17,7 +17,7 @@ def build(shape, data, target = nil)
         ref = shape.is_a?(ShapeRef) ? shape : ShapeRef.new(shape: shape)
         target ||= []
         @builder = DocBuilder.new(target: target, indent: @indent, pad: @pad)
-        structure(location_name(ref, ref.member_name), ref, data)
+        structure(ref.location_name || ref.shape.traits['smithy.api#xmlName'] || ref.shape.name, ref, data)
```
jterapin, line 20: Was there any documentation on resolving XML element name? I looked in Smithy Docs and couldn't find anything that jumps out. I was hoping to save the reference link for future me.

gems/smithy-xml/lib/smithy-xml/builder.rb
```diff
@@ -76,35 +77,54 @@ def map(name, ref, values) # rubocop:disable Metrics/AbcSize
...
+            shape(location_name(member_ref, member_ref.location_name), member_ref, values[member_name])
           end
         end
       end
 
       def structure_attrs(ref, values)
         ref.shape.members.each_with_object({}) do |(member_name, member_ref), attrs|
           if xml_attribute?(member_ref) && values.key?(member_name)
-            attrs[location_name(member_ref, member_ref.member_name)] = values[member_name]
+            attrs[location_name(member_ref, member_ref.location_name)] = values[member_name]
```
jterapin, line 95: Do we need to ensure whether if the member shape supports xmlAttribute trait? Or is this something that is caught through smithy build?

[Smithy Docs](https://smithy.io/2.0/spec/protocol-traits.html#xmlattribute-trait) says:
> Structure members that target boolean, number, string, or timestamp

gems/smithy/lib/smithy/views/client/schema.rb
```diff
@@ -112,13 +109,22 @@ def paginated?
...
           def paginator
-            "Paginators::#{Model::Shape.name(@id)}.new"
+            "Paginators::#{@name}.new"
           end
 
           private
 
+          def build_input(input)
+            ShapeRef.new(@service, nil, input)
+          end
```
jterapin, line 119: Do we no longer need to reference the location name for input/output shapes?

jterapin (approved):
Looks good, just have few questions for my learnings.

gems/smithy/lib/smithy/views/client/schema.rb
mullermp: It doesn't right now. I'm not sure if we should or shouldn't. I am not confident in the builder logic. We could go back to member name or keep location name.

gems/smithy-xml/lib/smithy-xml/builder.rb
mullermp: It should be caught at build time. The structure visitor will just check for the trait for any structure member.

gems/smithy-xml/lib/smithy-xml/builder.rb
mullermp: No, the docs are not great regarding XML. The only behavior I really saw was [this](https://smithy.io/2.0/spec/protocol-traits.html#xmlname-on-structures-and-unions) which says the xml name trait doesn't influence serialization of members that target it. There is a protocol test for it. However, I'm not sure if this logic is entirely correct but it works.

gems/smithy/lib/smithy/views/client/schema.rb
jterapin: Gotcha. We can keep the location name for now.

PR #323, opened 2025-08-12
Support VERSION file

mullermp:
Generates a VERSION and CHANGELOG.md file that persists if rebuilt. We have a separate process that reads changelog updates and updates the version file.


gems/smithy/lib/smithy/templates/client/gemspec.erb
```diff
@@ -4,10 +4,11 @@
 
 Gem::Specification.new do |spec|
   spec.name        = '<%= gem_name %>'
-  spec.version     = '<%= gem_version %>'
+  spec.version     = File.read(File.expand_path('../VERSION', __FILE__))
   spec.summary     = 'Generated gem using Smithy'
   spec.authors     = ['Smithy Ruby']
```
jterapin, line 9: Any thoughts on how to go about codegenerating these + email? I'm thinking of a custom smithy trait. I'm not sure if `smithy-build.json` carries any other metadata that we can retrieve

gems/smithy/lib/smithy/generators/client.rb
```diff
@@ -29,6 +32,8 @@ def source
       # rubocop:disable Metrics/AbcSize
       def gem_files
         Enumerator.new do |e|
+          e.yield 'VERSION', Views::Client::Version.new(@plan).render
+          e.yield 'CHANGELOG.md', Views::Client::ChangelogMd.new.render
```
jterapin, line 36: Should we include the `LICENSE.txt` as well?

jterapin (approved):
LGTM!

gems/smithy/lib/smithy/generators/client.rb
mullermp: I think the gemspec with SPDX Apache-2.0 is sufficient.

gems/smithy/lib/smithy/templates/client/gemspec.erb
mullermp: I think it would be through welds.

PR #324, opened 2025-09-05
SSO Credentials Changes

richardwang1124:
Updates HTTPBearer to Token, HTTPApiKey to ApiKey and handles optional auth


gems/smithy-client/lib/smithy-client.rb
```diff
@@ -53,6 +53,9 @@
 require_relative 'smithy-client/identity'
 require_relative 'smithy-client/identity_provider'
 require_relative 'smithy-client/refreshing_identity_provider'
+require_relative 'smithy-client/identities/http_api_key'
+require_relative 'smithy-client/identities/http_login'
+require_relative 'smithy-client/identities/token'
```
mullermp: The auth plugins should require these - we should not need these. What issues were you seeing?

gems/smithy-client/lib/smithy-client/identities/token.rb
```diff
@@ -3,15 +3,21 @@
...
           super(**)
         end
 
         # @return [String, nil]
         attr_reader :token
+
+        # Removing the token from the default inspect string.
+        # @api private
+        def inspect
+          "#<#{self.class.name} token=[FILTERED]>"
```
mullermp: I think this would need to include expiration too right? I think a better implementation may just be to gsub out `@token="contents"` with a regex and replace it with filtered.

gems/smithy-client/lib/smithy-client/plugins/resolve_auth.rb
```diff
@@ -38,7 +38,7 @@ def resolve_auth(context, auth_options)
             auth_options.each do |auth_option|
               # Anonymous auth does not have a plugin and does not sign,
               # so if auth scheme is noAuth then just return scheme_id.
-              return { scheme_id: auth_option } if auth_option == 'smithy.api#noAuth'
+              return { scheme_id: auth_option } if %w[smithy.api#noAuth smithy.api#optionalAuth].include?(auth_option)
```
mullermp: Talked offline, but I think this is correct, and the generator needs a change.

gems/smithy-client/lib/smithy-client.rb
```diff
@@ -53,6 +53,9 @@
 require_relative 'smithy-client/identity'
 require_relative 'smithy-client/identity_provider'
 require_relative 'smithy-client/refreshing_identity_provider'
+require_relative 'smithy-client/identities/http_api_key'
```
mullermp: I think similar to token, this should just be ApiKey.

gems/smithy-client/lib/smithy-client.rb
richardwang1124: When I try to use `Smithy::Client::Identities::Token` in V4 without requiring, it gives me the following error:
```
/Users/wangrch/Documents/aws-sdk-ruby-v4/gems/aws-sdk-ssooidc/lib/aws-sdk-ssooidc/ssooidc_token.rb:49:in 'Aws::SSOOIDC::SSOOIDCToken#refresh': uninitialized constant Smithy::Client::Identities (NameError)

        @identity = Smithy::Client::Identities::Token.new(token: token_jso...
                                  ^^^^^^^^^^^^
Did you mean?  Smithy::Client::Identity
```
If only the auth plugins are requiring these then I can see why there's an issue. SSOCredentials doesn't use Token auth so the plugin wouldn't be added and Token wouldn't be required, which is why I run into the uninitialized constant error.

mullermp: Yes, but in v4, you can require it there in the SSO class. I don't think we should globally require it here.

gems/smithy-client/lib/smithy-client/identities/token.rb
```diff
@@ -0,0 +1,26 @@
...
+      class Token < Identity
+        def initialize(token:, **)
+          @token = token
+          super(**)
+        end
+
+        # @return [String, nil]
+        attr_reader :token
+
+        alias original_inspect inspect
```
mullermp: No need for this I don't think. You can always print the value.

gems/smithy-client/lib/smithy-client/identities/token.rb
```diff
@@ -0,0 +1,26 @@
...
+
+        # Removing the token from the default inspect string.
+        # @api private
+        def inspect
+          original_inspect.gsub(/@token="[^"]*"/, '@token=[FILTERED]')
```
mullermp: I think this is fine. Keys and tokens should not have quote characters right?

gems/smithy-client/sig/smithy-client/identities/api_key.rbs
```diff
@@ -4,6 +4,7 @@ module Smithy
       class HttpApiKey < Identity
         def initialize: (key: String, **untyped options) -> void
         attr_reader key: String
+        def inspect: () -> String
       end
```
mullermp: Inspect is a standard method on Object so no need to put it in rbs

gems/smithy/spec/interfaces/client/auth_resolver_spec.rb
```diff
@@ -51,6 +51,12 @@
           auth_options = subject.resolve(params)
           expect(auth_options).to eq(['smithy.api#noAuth'])
         end
+
+        it 'returns a noAuth option for the operation with optionalAuth trait' do
+          params = AuthTrait::AuthParameters.new(operation_name: :operation_f)
+          auth_options = subject.resolve(params)
+          expect(auth_options).to eq(['smithy.api#noAuth'])
```
mullermp, line 91: This seems wrong. Wouldn't there be a list of 3 auths here? The two bound to the service and then no auth should be at the end.

You probably want a subsequent test for a service with no auth at all and optionalAuth (it should be benign in that case and it shouldn't have no auth twice).

gems/smithy-client/lib/smithy-client/identities/token.rb
richardwang1124: I can't confirm for certain that they shouldn't contain quote characters. The token provider SEP just says the SSO token is a opaque UTF-8 string. I'll update the regex pattern to handle `\"`, since I'm assuming escaped double quotes would be the only way a quote character could appear within the token.

gems/smithy-client/lib/smithy-client/identities/token.rb
richardwang1124: Can you explain this some more? I needed some way to access the original inspect's results to do the gsub.

gems/smithy/spec/interfaces/client/auth_resolver_spec.rb
richardwang1124: Just checked with Manuel from the Smithy team and I think you're right, it should be [httpBasicAuth, httpDigestAuth, noAuth]. I think our auth_resolver needs to be updated then, I'll take a look.

gems/smithy-client/lib/smithy-client/identities/token.rb
mullermp: Talked offline, use `super`.

gems/smithy/spec/fixtures/auth/no_auth/model.smithy
```diff
@@ -0,0 +1,20 @@
...
+
+service ServiceWithNoAuth {
+    version: "2020-01-29"
+    operations: [
+        OperationH
+        OperationI
+    ]
+}
+
+// This operation does not have the @auth trait and is bound to a service
```
mullermp, line 13: These comments are wrong. It does not have the `@auth` trait and it does not have an auth trait (`@auth` mentioned twice).

gems/smithy/spec/interfaces/client/auth_resolver_spec.rb
```diff
@@ -24,6 +24,13 @@
           auth_options = subject.resolve(params)
           expect(auth_options).to eq(['smithy.api#httpDigestAuth'])
         end
+
+        it 'returns the auth options for the operation with the optionalAuth trait' do
+          params = NoAuthTrait::AuthParameters.new(operation_name: :operation_g)
+          auth_options = subject.resolve(params)
+          expect(auth_options).to eq(%w[smithy.api#httpBasicAuth smithy.api#httpBearerAuth smithy.api#httpDigestAuth
```
mullermp: multiline

gems/smithy/lib/smithy/views/client/auth_resolver.rb
```diff
@@ -93,10 +93,12 @@ def operation_auth_schemes(operation)
           if operation_auth?(operation)
             operation_auth = operation_traits.fetch('smithy.api#auth', [])
             add_auth_schemes_from_auth_trait(auth_schemes, operation_auth)
+          elsif optional_operation_auth?(operation)
```
mullermp: I'm skeptical of this. Let's chat offline.

mullermp (approved):
LGTM assuming minor fixes from prev comments.

PR #327, opened 2025-09-16
Upfront Identity Fetching

mullermp:
Rework of identity and credentials to be more like SDK v3.

Major changes:

* Simplify identity classes by removing Identity base
* Make all identity providers refreshable by default
* Add expiration and set? into the identity provider interface
* Resolve credentials up front on client config


gems/smithy-client/lib/smithy-client/plugins/resolve_auth.rb
```diff
@@ -36,41 +36,26 @@ def resolve_auth(context, auth_options)
...
-                failures << "Auth scheme #{auth_option} was not enabled for this request"
+              unless context.config.auth_schemes.key?(scheme_id)
+                failures << "Auth scheme #{scheme_id} was not enabled for this request"
                 next
               end
 
-              identity_provider = context.config.auth_schemes[auth_option]
-              resolved_auth = try_load_auth_scheme(auth_option, identity_provider, failures)
+              identity_provider = context.config.auth_schemes[scheme_id]
+              unless identity_provider
```
richardwang1124: Do we need to check that the identity_provider is set?

gems/smithy-client/lib/smithy-client/refreshing_identity_provider.rb
```diff
@@ -10,20 +10,23 @@ module RefreshingIdentityProvider
       SYNC_EXPIRATION_LENGTH = 300 # 5 minutes
       ASYNC_EXPIRATION_LENGTH = 600 # 10 minutes
 
-      def initialize
+      def initialize(_options = {})
```
richardwang1124, line 13: Why is the options parameter added back into the method header?

gems/smithy-client/lib/smithy-client/plugins/resolve_auth.rb
mullermp: It should have already done that on client config, right? It would either be nil or a provider if set.

gems/smithy-client/lib/smithy-client/refreshing_identity_provider.rb
mullermp: I suppose this is not necessary right now, but I may want to add back the before refresh callback. Is there a use case maybe with token file reading in SSO?

gems/smithy-client/lib/smithy-client/plugins/resolve_auth.rb
richardwang1124: If the customer does not provide a provider, then yes, our default option would either be nil or a provider which is set. However, its possible that a customer may provide a provider that isn't set right? Should we still return the auth scheme even if the provider isn't set, or should we continue checking for other auth schemes? From the previous resolve auth code, I think we checked that not only does the auth scheme have a configured provider, but also that its identity is resolved.

gems/smithy-client/lib/smithy-client/refreshing_identity_provider.rb
richardwang1124: I do see that in v3 we have the `before_refresh` option in certain providers and `@before_refresh` member in the `RefreshingCredentials`, but I don't immediately see where this is used in the SDK code. Going through the default chain at least I don't think we use `before_refresh` for SSOCredentials. What use case were you thinking of?

mullermp: It's used by customers. My thought was we can make it a list of callbacks and not just one proc, then we can register a default one that also reread the token file instead of building it into refresh, but I don't know if we care to do that. It was just a thought.

gems/smithy-client/lib/smithy-client/plugins/resolve_auth.rb
mullermp: That's a good call out. If they explicitly configured a provider and it is not set, then yes I think it makes sense to exhaust other modeled auth types. If you'd like to make the change and a test case for that in the resolve auth plugin go for it.

gems/smithy-client/spec/smithy-client/plugins/resolve_auth_spec.rb
```diff
@@ -87,22 +83,22 @@ def resolve(_)
...
+          client = client_class.new(stub_responses: true, api_key_provider: nil)
+          expect { client.operation }.to raise_error(/did not have an identity provider configured/)
         end
 
         it 'raises an error when identity resolver fails to resolve identity' do
           shapes['smithy.ruby.tests#SampleClient']['traits']['smithy.api#httpApiKeyAuth'] = {}
           client_class.add_plugin(HttpApiKeyAuth)
-          provider = HttpApiKeyProvider.new('stubbed_key')
+          provider = ApiKeyProvider.new(key: 'stubbed_key')
           expect(provider).to receive(:identity).and_return(nil)
```
mullermp: We can improve this test. Just pass an empty instance of the provider. Set? Should return false on its own.

gems/smithy-client/lib/smithy-client/plugins/resolve_auth.rb
```diff
@@ -36,40 +36,30 @@ def resolve_auth(context, auth_options)
...
+                failures << "Auth scheme #{scheme_id} did not have an identity provider configured"
+                next
+              end
 
-          def try_load_auth_scheme(scheme_id, identity_provider, failures)
-            unless identity_provider
-              failures << "Auth scheme #{scheme_id} did not have an identity resolver configured"
-              return
-            end
+              unless identity_provider.identity
```
mullermp: I think we want to check set?

gems/smithy-client/lib/smithy-client/plugins/http_basic_auth.rb
```diff
@@ -1,43 +1,44 @@
 # frozen_string_literal: true
 
-require_relative '../http_login_provider'
-require_relative '../identities/http_login'
+require 'base64'
```
richardwang1124: Is this require necessary?

richardwang1124 (approved):
I think it looks good overall and changes to V4 are minor. Let me know if you end up keeping the options in the RefreshingIdentityProvider initializer.

PR #328, opened 2025-09-29
Endpoints auth

mullermp:
Resolves auth using an auth module instead of entirely in the resolve auth plugin. This is necessary so that presigners and other auth related utilities can resolve auth. The auth is now once again an auth scheme and has an attached signer and identity provider, so that signing can be done.

Also implements auth scheme preference config.


gems/smithy-client/lib/smithy-client/api_key_signer.rb
```diff
@@ -0,0 +1,60 @@
...
+
+module Smithy
+  module Client
+    # Signs requests with the ApiKey identity.
+    class ApiKeySigner
+      def initialize(options = {})
+        @name = options[:name]
+        @in = options[:in]
+        @scheme = options[:scheme]
+      end
```
jterapin, line 11: Are these intended to be public interfaces - if so, we should include more documentation about init options?  Here, other signers and auth scheme.

gems/smithy-client/lib/smithy-client/auth.rb
```diff
@@ -0,0 +1,106 @@
...
+              properties[key] = value
+            end
+            normalized_endpoint_schemes << { scheme_id: normalized_scheme_id, signer_properties: properties }
+          end
+          resolved_auth_options = prioritize_auth_options(normalized_endpoint_schemes, config.auth_scheme_preference)
+          resolve_auth_scheme(config.auth_schemes, resolved_auth_options)
+        end
+
+        def resolve_without_endpoint_auth(config, auth_parameters)
+          auth_options = config.auth_resolver.resolve(auth_parameters)
```
richardwang1124, line 42: If I'm understanding this correctly, endpoint auth completely overrides the modeled auths?

gems/smithy-client/lib/smithy-client/auth.rb
```diff
@@ -0,0 +1,106 @@
...
+            normalized_scheme_id = endpoint_auth_schemes_map[scheme['name']]
+            next unless normalized_scheme_id
+
+            properties = {}
+            scheme.each do |key, value|
+              next if key == 'name'
+
+              properties[key] = value
+            end
+            normalized_endpoint_schemes << { scheme_id: normalized_scheme_id, signer_properties: properties }
```
richardwang1124, line 35: I'm assuming we'll need to make changes to sigv4 auth in V4 as well following these changes. Have you tested with V4?

gems/smithy-client/lib/smithy-client/plugins/resolve_auth.rb
```diff
@@ -10,10 +10,18 @@ class ResolveAuth < Plugin
...
-          docstring: 'An object that resolves authentication schemes for request signing'
+          docstring: 'An object that resolves authentication schemes for request signing.'
         )
 
-        # @api private
+        option(
+          :auth_scheme_preference,
+          doc_type: 'Array<String>',
+          rbs_type: 'Array[String]',
+          docstring: 'A list of preferred authentication schemes to use when making a request.'
```
jterapin: Should we expand on what schemes are available (so that users know what valid values to pass in)? Also - may want to state the array is in priority order.

jterapin (approved):
LGTM but there appears to be some RBS failure - FYI.

richardwang1124 (approved):
Looks good overall to me. There's a few failing RBS tests.

gems/smithy-client/lib/smithy-client/api_key_signer.rb
mullermp: I intended them to be API private for now. I'm also not sure yet so I'm going to wait on documentation.

gems/smithy-client/lib/smithy-client/plugins/resolve_auth.rb
mullermp: Maybe - I'd need a way for this documentation to have a complete list. That might be complex. But I can improve the docs to say use the full scheme ids and priority order.

gems/smithy-client/lib/smithy-client/auth.rb
mullermp: Yeah. I think we should do it that way. It's less complex and I think endpoint auth is guaranteed to work.

gems/smithy-client/lib/smithy-client/auth.rb
mullermp: Yeah, I tested with v4 and I opened a separate PR.

PR #329, opened 2025-10-02
Add auth option class + signer public

mullermp:
Adds auth option to the auth resolver to make that more explicit.

Makes the signer classes public.

Re-enables smithy rbs (disabled by accident)


gems/smithy-client/lib/smithy-client/auth_option.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    # Contains information about candidate authentication schemes.
```
jterapin, line 5: Documentation on options?

jterapin (approved):
LGTM

PR #331, opened 2025-10-13
Do not default endpoint with stub responses

mullermp:
Do not default the endpoint with stub responses.

In AWS clients, we want to default to a stubbed region which should make a valid regional endpoint, used for presigning or signing tests.

Other solutions may involve wrapping the default endpoint provider with a stubbed one that checks stubbed responses?


jterapin (approved):
LGTM

richardwang1124 (approved):

PR #332, opened 2025-11-05
Update features to use set

richardwang1124:
Track features using a set to prevent duplicate metrics.


gems/smithy-client/spec/smithy-client/features_spec.rb
```diff
@@ -57,6 +57,21 @@ module Client
         end
         expect(Features.tracked).to be_empty
       end
+
+      it 'does not track duplicate features' do
```
mullermp, line 61: I think amend this test to push A B at the same time and assert that both get removed together. This properly tests the add/remove logic you made.

mullermp (approved):

jterapin (approved):

PR #334, opened 2025-12-03
Support before_refresh

richardwang1124:
Support before_refresh procs for RefreshingIdentityProvider.


jterapin (approved):
LGTM, nice helpers and looks about same as V3

PR #339, opened 2026-05-27
fix: disable Net::HTTP automatic gzip decompression

jterapin:
## Summary
- Sets a default empty `accept-encoding` header in `net_headers_for`, preventing Net::HTTP from adding `Accept-Encoding: gzip` and enabling automatic decompression
- Uncomments existing tests that were written for this behavior in #222

## Context
Taken care of as part of oncall. V3 already does this at [`seahorse/client/net_http/handler.rb:185`](https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-core/lib/seahorse/client/net_http/handler.rb#L185)

### DIFF
--- a/gems/smithy-client/lib/smithy-client/net_http/handler.rb
+++ b/gems/smithy-client/lib/smithy-client/net_http/handler.rb
@@ -137,7 +137,9 @@ module Smithy
         # @param [Http::Request] request
         # @return [Hash<String, String>]
         def net_headers_for(request)
-          headers = {}
+          # Net::HTTP defaults decode_content=true and adds Accept-Encoding: gzip.
+          # Setting an empty value prevents automatic decompression.
+          headers = { 'accept-encoding' => '' }
           request.headers.each_pair do |key, value|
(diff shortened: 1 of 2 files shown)


gems/smithy-client/lib/smithy-client/net_http/handler.rb
```diff
@@ -137,7 +137,9 @@ def net_http_request_class(request)
         # @param [Http::Request] request
         # @return [Hash<String, String>]
         def net_headers_for(request)
-          headers = {}
+          # Net::HTTP defaults decode_content=true and adds Accept-Encoding: gzip.
+          # Setting an empty value prevents automatic decompression.
+          headers = { 'accept-encoding' => '' }
```
richardwang1124: Thoughts on setting this to `'identity'` explicitly? Seems like both result in similar behavior but using `'identity'` may be more semantically correct and less ambiguous.

richardwang1124 (approved):

PR #341, opened 2026-05-28
feat: Add transfer encoding plugin

sichanyoo:

//

For reviewers:
- This is the port of the [V3 transfer encoding plugin](https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-core/lib/aws-sdk-core/plugins/transfer_encoding.rb).
- The `transfer-encoding: chunked` header is used when the client sending the request doesn't know the size of the payload in advance (e.g., for streaming real time data from pipe), and end of body is signaled with a 0-byte chunk at the end. The logic in V3 transfer encoding plugin specifically handles that case by checking that payload is a streaming type with no known length & is unsigned payload (this case is hit by small set of services like Lex, Transcribe, etc. for sending real-time stream of data). Note that signed payload with no known length gets rejected by our signer so it doesn't need to be covered by TE plugin.
- There's another usecase for `transfer-encoding: chunked` that is specific to AWS, which is related to [aws.api#awsChunked trait](https://smithy.io/2.0/aws/aws-core.html#aws-api-awschunked-trait). AWS chunked encoding allows sending payload in chunks with running checksum and signature, which reduces upfront cost of having to read the entire body to calculate checksum and signature. As you'd expect, S3 operations like put object and upload part is why this functionality exists. The trait is part of Smithy spec as of 1.69.0, but service models have not been updated to use awsChunked trait just yet. At the moment, whether to use aws chunking for payload or not is internal implementation unique to each SDK. When using aws chunked encoding for S3 operations, SEP specifies adding `transfer-encoding: chunked` header and `x-amz-decoded-content-length` header, where `x-amz-decoded-content-length` header is the raw size of the chunk before it is encoded with aws chunk encoding. The reason S3 needs raw size is to be able to allocate space for the object as it is arriving. S3 also accepts `content-length` in place of `transfer-encoding: chunked`, if SDK provides length of encoded chunk in `content-length`.
- Ruby V3 SDK wraps the stream into [`AwsChunkedTrailerDigestIO`](https://github.com/aws/aws-sdk-ruby/blob/c660b98d16b39ad8034422e3d5028ce9a61141a2/gems/aws-sdk-core/lib/aws-sdk-core/plugins/checksum_algorithm.rb#L482) that exposes size of the encoded chunk via simple calculation done on raw size. Because of this, `transfer-encoding: chunked` isn't used for aws chunk streaming in Ruby V3 SDK, as noted in comment [here](https://github.com/aws/aws-sdk-ruby/blob/c660b98d16b39ad8034422e3d5028ce9a61141a2/gems/aws-sdk-core/lib/aws-sdk-core/plugins/checksum_algorithm.rb#L401-L403). Instead, it just uses `content-length` header.
- Given only the generic use of `transfer-encoding: chunked` as outlined by HTTP spec is covered by V3 transfer encoding plugin and smithy-ruby is for generic code, the plugin was just ported over as-is, without handling aws chunked logic. The `transfer-encoding: chunked` usecase for aws chunked streaming can be considered when flexible checksum work is picked up.
- Note that the [aws.auth#unsignedPayload trait](https://smithy.io/2.0/aws/aws-auth.html#aws-auth-unsignedpayload-trait) is namespaced to AWS but isn't AWS-specific and is considered generic to Smithy, kind of like `aws.auth#sigv4`.


gems/smithy-client/lib/smithy-client/plugins/transfer_encoding.rb
```diff
@@ -0,0 +1,47 @@
...
+            elsif unsigned_payload?(context)
+              context.http_request.headers['Transfer-Encoding'] = 'chunked'
+            end
+          end
+
+          def streaming_member_ref(context)
+            context.operation.input.shape.members.detect do |_, ref|
+              ref.shape.traits.key?('smithy.api#streaming')
+            end&.last
+          end
```
jterapin, line 35: Mostly a note for myself but we could optimize to find out whether there is a streaming member ref at codegen instead of runtime. I have an assigned task to revisit shapes anyways.

gems/smithy-client/lib/smithy-client/plugins/transfer_encoding.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+module Smithy
+  module Client
+    module Plugins
+      # @api private
+      class TransferEncoding < Plugin
```
jterapin, line 10: One thing I want to track for follow-up:

```
TODO: Verify whether aws.api#awsChunked trait is supported in smithy-ruby's 
model/codegen. If it is (or once service models adopt it), revisit this plugin to
handle AWS chunked transfer encoding alongside the generic HTTP case.
```

Not blocking this PR, just want to make sure we don't lose track of it.

gems/smithy-client/lib/smithy-client/plugins/transfer_encoding.rb
```diff
@@ -0,0 +1,47 @@
...
+            apply_transfer_encoding(context, payload_ref.shape) if payload_ref
+            @handler.call(context)
+          end
+
+          private
+
+          def apply_transfer_encoding(context, payload)
+            return if context.http_request.body.respond_to?(:size)
+            if requires_length?(payload)
+              raise 'Required `Content-Length` value missing for the request.'
```
jterapin: Wow, I'm surprised that we don't have any smithy-client level errors other than service error!

This is going to raise a generic `RuntimeError` class so maybe worth defining `MissingContentLength`? We could definite the error class here -or- have it under a `Errors` namespace so we can use it elsewhere as well.

gems/smithy-client/spec/smithy-client/plugins/transfer_encoding_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+              def call(context); end
+            end.new
+          end
+          let(:handler) { TransferEncoding::Handler.new(noop_handler) }
+
+          let(:unsizable_body) do
+            io = StringIO.new('data')
+            io.singleton_class.undef_method(:size)
+            io.singleton_class.undef_method(:length)
+            io
```
jterapin: This works but we could consider using `double('data')` here instead of undefining methods on a real `StringIO`. 

Context: https://rspec.info/features/3-13/rspec-mocks/

gems/smithy-client/spec/smithy-client/plugins/transfer_encoding_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+          end
+
+          def build_context(traits: {}, operation_traits: {}, body: StringIO.new('data'))
```
jterapin: The other plugin specs (i.e. `host_prefix`, etc) test through `client.operation(...)`, which tests the behavior rather than internal class constructors. Could we follow a similar pattern here?

jterapin (commented):
Hey, thanks for taking this on. So many moving pieces! Hopefully, we will get clarification upon closer in.

gems/smithy-client/lib/smithy-client/plugins/transfer_encoding.rb
sichanyoo: Yeah that caught me off guard as well. In V3 I saw that `MissingContentLength` error is literally only used by transfer encoding plugin, so in terms of reusing the error it probably isn't necessary to define it. As for compatibility with V3, since we can't namespace the error to Aws like it is in V3 atm, whatever we do it's gonna break customer that was catching it by the error name `MissingContentLength`. If you want this to be defined as error nevertheless, I can def add it, lmk.

jterapin: Ah, this is going to be fun. The separation of generic and AWS gets interesting.

I see that we do have `MissingContentLength` error defined in Core namespace so I imagined our compatibility path is shimming the aliases (or maybe we walk back on the namespace changes). 

This may be a case where you need to have _two_ separate transfer encoding plugins to cover generic and AWS-specific cases (raises AWS-specific error). See how we handled UserAgent plugins.

But we could try consolidating into having a single Smithy plugin but will break AWS customers who are migrating from V3 to V4 (IF there's any that catches that error class). Given that most AWS use cases will include a Content-Length path, maybe this could be considered as a low risk break? Let me know what you think!

I also don't think it's much of a lift to define this error class (it's an one-linear) and allows callers to distinguish it from any other RuntimeError in their application

sichanyoo: Yeah that (consolidating into having single Smithy plugin) sounds good; and I'll change it to define it into class 👍

gems/smithy-client/lib/smithy-client/plugins/transfer_encoding.rb
sichanyoo: Added TODO comment as suggested.

gems/smithy-client/lib/smithy-client/plugins/transfer_encoding.rb
jterapin: > I see that we do have MissingContentLength error defined in Core namespace so I imagined our compatibility path is shimming the aliases (or maybe we walk back on the namespace changes).

This is my mistake. I was looking at the wrong branch here. ;)

gems/smithy-client/spec/smithy-client/plugins/transfer_encoding_spec.rb
sichanyoo: Changed as suggested.

gems/smithy-client/spec/smithy-client/plugins/transfer_encoding_spec.rb
sichanyoo: Changed as suggested. Only complication was that sample client used for tests in smithy-ruby uses RPCv2CBOR, and RPCv2CBOR plugin serializes body even when payload is stream instead of just passing it through. To mimic that behavior I swapped it out with passthrough plugin.

jterapin (approved):
LGTM! Thank you for working on this.

PR #343, opened 2026-06-08
feat: Retry alignment

sichanyoo:


gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
```diff
@@ -6,66 +6,31 @@ module Plugins
...
-            * Any instance of a class that implements the following methods:
-              - `acquire_initial_retry_token(token_scope)`
-              - `refresh_retry_token(retry_token, error_info)`
-              - `record_success(retry_token)`
+            Specifies which retry algorithm to use. Values are:
+
+            * `standard` - A standardized set of retry rules across the AWS SDKs.
+              This includes support for retry quotas, which limit the number of
+              unsuccessful retries a client can make. This is the default
+            value if no retry mode is provided.
```
richardwang1124: nit: spacing

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
```diff
@@ -76,23 +41,24 @@ class RetryErrors < Plugin
             not retry instead of sleeping.
           DOCS
 
+        # @api private undocumented
+        option(:retry_strategy)
```
richardwang1124: Is there a reason why we're using this instead of `retry_mode`?

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
```diff
@@ -6,66 +6,31 @@ module Plugins
       # @api private
       class RetryErrors < Plugin
         option(
-          :retry_strategy,
```
richardwang1124, line 9: Juli might have more context, but should the class option be removed? I think it allowed customers to implement their own custom retry strategies, but not sure if we still want to support that or not.

gems/smithy-client/lib/smithy-client/plugins/stub_responses.rb
```diff
@@ -28,12 +28,6 @@ def add_handlers(handlers, config)
           handlers.add(StubHandler, step: :send)
         end
 
-        def after_initialize(client)
```
richardwang1124, line 31: Why is this removed? In V3 retries are disabled when stub_responses is true since customers need to mock the response behavior.

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
```diff
@@ -76,23 +41,24 @@ class RetryErrors < Plugin
...
             when 'adaptive'
               Retry::Adaptive.new(
-                max_attempts: config.retry_max_attempts,
-                backoff: config.retry_backoff,
+                max_attempts: config.max_attempts,
                 wait_to_fill: config.adaptive_retry_wait_to_fill
               )
             else
-              config.retry_strategy
+              raise ArgumentError, 'Must provide either standard` or `adaptive` for retry_mode'
```
richardwang1124: Missing a backtick at the beginning: standard`

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
sichanyoo: SEP doesn't require custom retry strategy configuration & afaik specific configurable options exposed in V3's retry plugin [also only apply to legacy](https://github.com/aws/aws-sdk-ruby/blob/6d2decc949d0f819c0d6a5164529c1080c92d056/gems/aws-sdk-core/lib/aws-sdk-core/plugins/retry_errors.rb#L13-L95). I definitely might be missing something but it seems V3 doesn't support fully customizable strategy either.

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
sichanyoo: Fixed spacing

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
sichanyoo: It felt good to separate out `retry_mode` string config from the resolved strategy object. Storing resolved strategy in `retry_mode` felt a little awkward. But I'm open to discussion if you feel strongly / there's some rubyist convention to not do what I did.

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
sichanyoo: Fixed typo

gems/smithy-client/lib/smithy-client/plugins/stub_responses.rb
sichanyoo: Instead of having this logic be here, I moved it off to `add_handlers` in the retry error plugin. It's sort of annoying because we use this same plugin in AWS side also, and this doesn't remove AWS-specific retry plugin. Actually, alternative & better way might be having a AWS-specific stub_responses.rb plugin that replaces this, so we can still have all the plugins to remove in one place. Lmk what you think.

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
jterapin: I'm not sure if I like the idea of using both terminology since it is hard to different what makes these two verbiage different (to me at least! Coming from a non-SRA/smithy perspective). But I understand the motivation. Looking at Smithy/AWS SDK retries documentation seems to point me to similar verbiage so I think this is the most reasonable one:
* https://docs.aws.amazon.com/sdkref/latest/guide/feature-retry-behavior.html
* https://smithy.io/2.0/guides/client-guidance/retries.html

That being said... I'm unsure whether we want to have this as a private config for customers. I'm thinking of the following flow:

* customer passes no default for `retry_mode`, we create `Retry::Standard` instance using default values and set it as a value for `retry_strategy`
* customer passes in `:standard` but sets the max attempts to 10, we create `Retry::Standard` instance using the passed-in values and set it as a value for `retry_strategy`
* customer passes in their OWN custom retry instance to `:retry_strategy` and we will respect this customization

Overall, if a `retry_strategy` has an instance, we have no need to check for other retry options.

We can have this discussion offline for better alignment.

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
jterapin: In V3, a customer could pass their own custom `Proc` to influence how backoffs work but that is under the `:legacy` umbrella. We could have a way to accommodate this legacy support via migration gem. 

While SEP does not mention custom retries, I did see some references in Smithy docs: https://smithy.io/2.0/guides/client-guidance/retries.html

For customers who wants to NOT use migration gem BUT want custom retries to be a thing. They could opt to implement their own via duck typing our retries interface (aka below). Other ways are replacing this plugin with their own. Or even utilizing custom welds. 

Note: I added more context in a separate comment below on what we need to do to resolve this.

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
```diff
@@ -6,66 +6,31 @@ module Plugins
       # @api private
       class RetryErrors < Plugin
         option(
-          :retry_strategy,
+          :retry_mode,
```
jterapin, line 9: +1 on moving back to V3 naming.

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
```diff
@@ -6,66 +6,31 @@ module Plugins
...
-            * `adaptive` - An experimental retry strategy that includes all the functionality of the
-              `standard` strategy along with automatic client side throttling. This is a provisional
-              strategy that may change behavior in the future.
-            * Any instance of a class that implements the following methods:
-              - `acquire_initial_retry_token(token_scope)`
-              - `refresh_retry_token(retry_token, error_info)`
-              - `record_success(retry_token)`
+            Specifies which retry algorithm to use. Values are:
+
+            * `standard` - A standardized set of retry rules across the AWS SDKs.
```
jterapin: bit: Maybe reword as "Smithy-based" SDKs? 🤔

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
```diff
@@ -6,66 +6,31 @@ module Plugins
       # @api private
       class RetryErrors < Plugin
         option(
-          :retry_strategy,
+          :retry_mode,
           default: 'standard',
           doc_default: "'standard'",
-          doc_type: 'String, Class',
+          doc_type: String,
```
jterapin: Instead of string, what about Ruby symbols? `:standard`, `:adaptive` for examples.

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
jterapin, line 97: Not related to this PR since this was here before your time - I'm curious to know what is the motivation of passing `nil` here. I'm guessing it may have something to do with custom implementation but unsure.

gems/smithy-client/lib/smithy-client/plugins/stub_responses.rb
jterapin: I personally think keeping all the `stub_responses`-related logic contained here makes the overall sense to find all its configuration but I'm open to ways. At the end, We should probably lean one way however.

gems/smithy-client/lib/smithy-client/retry/exponential_backoff.rb
```diff
@@ -5,23 +5,33 @@ module Client
     module Retry
       # Default exponential backoff retry strategy for retrying requests.
       class ExponentialBackoff
```
jterapin, line 8: If there's no need to configure, does it make sense for us to even create instances? What state does this object carries that may be useful to keep?

gems/smithy-client/lib/smithy-client/retry/token.rb
```diff
@@ -8,6 +8,8 @@ class Token
         def initialize
           @retry_count = 0
           @retry_delay = 0
+          @capacity_amount = nil
```
jterapin, line 11: nil or 0?

gems/smithy-client/lib/smithy-client/retry/adaptive.rb
```diff
@@ -35,6 +31,12 @@ def initialize(options = {})
         # @return [Boolean]
         attr_reader :wait_to_fill
 
+        def request_bookkeeping(error_info)
```
jterapin: This method feels misleading since it is only handled at an error case. We can either rename this or consider handling the success case as well. I would look at this in the perspective if a custom implementor.

gems/smithy-client/lib/smithy-client/retry/exponential_backoff.rb
```diff
@@ -5,23 +5,33 @@ module Client
     module Retry
       # Default exponential backoff retry strategy for retrying requests.
```
jterapin: If we intend to use the wording, `retry_strategy` as a holder for the Retry instance, I think we should update this documentation. Reads as if it's an instance intended for `retry_strategy`

gems/smithy-client/lib/smithy-client/retry/exponential_backoff.rb
jterapin: I see that we do need to carry a state within AWS context. If we intend to allow for custom retries, should this be a component that customers can use? 🤔 Is there other components that customers may use outside of this?

jterapin (commented):
Nice, we probably should some have broader convos about the `retry_stategy` and custom impl.

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
sichanyoo: It seems the argument passed to it isn't even used, I think I can just remove it. Nice catch!

gems/smithy-client/lib/smithy-client/retry/token.rb
sichanyoo: Just nil, I don't think it really matters either way but nil is to signify it has not been assigned a quota capacity yet, whereas 0 means there was no capacity that could be assigned from the quota.

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
sichanyoo: Yeah I think taking custom retry instance via `:retry_strategy` and adding custom retry instance resolution to `after_initialize` makes sense. I can assert that the passed type has all the methods needed by retry flow before assigning it.

gems/smithy-client/lib/smithy-client/plugins/stub_responses.rb
sichanyoo: I agree with keeping all `stub_responses` related logic in a same place, I think what I'll do is create a AWS specific plugin and replace generic one with AWS specific one via default weld.

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
sichanyoo: Good catch, changed as suggested.

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
sichanyoo: That's cleaner for sure. Changed as suggested.

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
sichanyoo: Actually on second look, the token_scope comes from SRA for future extensibility. Maybe we can just keep it 🤷‍♂️

gems/smithy-client/lib/smithy-client/retry/exponential_backoff.rb
sichanyoo: Good catch! Removed the `strategy` from the doc string.

gems/smithy-client/lib/smithy-client/retry/adaptive.rb
sichanyoo: That's a good point. Refactored so that the function gets called on both success and error paths.

gems/smithy-client/lib/smithy-client/retry/exponential_backoff.rb
sichanyoo: Given `Quota` and `ClientRateLimiter` are both private, I think noting this type to be private also makes sense. I think for now just keeping customization point to be fully user-driven via `:retry_strategy` should suffice. Exposing reusable components can be done down the line as needed.

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
```diff
@@ -76,24 +40,54 @@ class RetryErrors < Plugin
...
           DOCS
 
+        option(
+          :retry_strategy,
+          doc_type: 'Object',
+          docstring: <<~DOCS)
+            The retry strategy used by the client. If not provided, a default strategy is built
+            based on `:retry_mode` — either `Standard` or `Adaptive`.
+
+            A custom strategy must respond to:
```
jterapin, line 50: Nice!

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
```diff
@@ -6,66 +6,30 @@ module Plugins
       # @api private
       class RetryErrors < Plugin
         option(
-          :retry_strategy,
-          default: 'standard',
-          doc_default: "'standard'",
-          doc_type: 'String, Class',
+          :retry_mode,
+          default: :standard,
+          doc_type: Symbol,
```
richardwang1124: What was the reasoning for turning this into a symbol? This config options for this when read from the environment variables or shared config will be strings and will need to be converted into symbols. Is this extra conversion necessary?

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
```diff
@@ -141,6 +151,11 @@ def reset_response(context, response)
             response.error = nil
           end
 
+          # TODO: Revisit after trait is finalized.
+          def long_polling_operation?(context)
+            context.operation.traits.key?('smithy.api#longPoll')
```
richardwang1124, line 156: Not sure how this is gonna look, but at least according to the SEP this trait is `aws.api#longPoll`.

richardwang1124 (approved):
Looks good to me overall. I'm okay with changing the option to a symbol, just wasn't sure if the extra conversion is necessary. Most options are also strings, I think only a few things like logging uses a symbol.

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
jterapin: This change was associated with my feedback. I personally prefer symbols when it comes to configurations options that are pre-determined and unique. Due to the following:
* Idiomatic Ruby, i would say by most ruby standards
* Symbols are immutable and cheaper to compare (identity vs character-by-character)
* Prevents subtle bugs like "standard " != "standard" (trailing whitespace in env vars)

I think conversation if needed, will happen once at client init. We can technically accept both gracefully which we have done this in V3 (for other configs? i think?)

But this is really a style choice so I don't have strong feeling on this so I'm good with whatever we decide on.

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
jterapin: Well, I should add a revisit task for this and ClockSkew on our board if there's isn't one already.

jterapin (commented):
LGTM!

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
sichanyoo: As discussed in person, I'll revert this option back to String to conform to preexisting pattern, but create a followup investigation task for looking into config rework which will include moving to using symbols instead of string literals where possible.

gems/smithy-client/lib/smithy-client/plugins/retry_errors.rb
sichanyoo: @richardwang1124 At least in Smithy website, it's under `smithy.api`. I guess we'll see where it gets namespaced to when it's finalized: https://smithy.io/2.0/trait-index.html.

PR #345, opened 2026-06-17
feat: Address pagination API gap with aligned approach

sichanyoo:

### Background & Decisions

Responses are Struct subclasses, and Ruby's Struct includes Enumerable by default. Without this fix, calling .each on a paginated response silently iterated struct field values instead of pages, a possible silent data-loss bug. This PR fixes the gap by adding .each to PageableResponse (delegating to each_page), undefining each on Smithy::Schema::Structure as a safety net, and introducing a PageEnumerator class that exposes only safe enumeration methods (map, select, flat_map, reduce,first, take, lazy) while blocking dangerous ones (count, sort, to_a, etc.) that would silently trigger full pagination with real API calls and real cost.

We chose a custom PageEnumerator over mixing Enumerable into the response because starting with a narrow surface and expanding later is non-breaking, whereas removing a dangerous method after customers depend on it is breaking. 

.each iterates pages rather than items because [items is optional in Smithy's @paginated trait](https://smithy.io/2.0/spec/behavior-traits.html#smithy-api-paginated-trait). About 21% of paginated operations at AWS have no items defined currently, and some operations return multiple result collections, making page iteration the only universally safe default.


gems/smithy-client/lib/smithy-client/page_enumerator.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
```
sichanyoo, line 1: This is the custom enumerator with limited API surface to safeguard against unsafe `Enumerable` methods

gems/smithy-schema/lib/smithy-schema/structure.rb
```diff
@@ -4,6 +4,10 @@ module Smithy
   module Schema
     # A module mixed into Structs that provides utility methods for Structure shapes.
     module Structure
```
sichanyoo, line 6: Add `each` undef to prevent silent data loss in the case `pageable_response` plugin doesn't run for whatever reason (e.g., customer customization on plugin list that leaves it out)

gems/smithy-client/lib/smithy-client/pageable_response.rb
```diff
@@ -36,6 +36,13 @@ def initialize(response)
     # This yields one response object per API call made. The SDK retrieves additional
     # pages of data to complete the request.
     #
```
sichanyoo, line 38: Updates the `PageableResponse` module that gets added to `Response` to return the custom enumerator when not give a block, and if given a block, enumerate using the custom enumerator and yield results. As it has been, `PageableResponse#each` will paginate over pages instead of items due to uniqueness of AWS background discussed offline.

sichanyoo (commented):
Comments for additional context

gems/smithy-client/lib/smithy-client/page_enumerator.rb
```diff
@@ -0,0 +1,40 @@
...
+      end
+
+      def take(val) = enum.take(val)
+      def lazy = enum.lazy
+
+      private
+
+      def enum
+        Enumerator.new(&@block)
+      end
```
jterapin, line 33: The current `PageEnumerator` creates a new Enumerator on every method call (map, select, first, etc.) via the private enum method. Is there a way to prevent this or was this intentional?

gems/smithy-client/lib/smithy-client/paginators/page_enumerator.rb
```diff
@@ -0,0 +1,40 @@
...
+      end
+
+      def map(&) = enum.map(&)
+      def select(&) = enum.select(&)
+      def filter(&) = enum.select(&)
+      def flat_map(&) = enum.flat_map(&)
+      def reduce(*, &) = enum.reduce(*, &)
+
+      def first(val = (no_arg = true
+                       nil))
```
jterapin: Fix alignment.

jterapin: Can we do: `def first(*args)` if we are just forwarding args?

gems/smithy-client/lib/smithy-client/pageable_response.rb
```diff
@@ -107,16 +126,29 @@ def each_page(&)
         end
```
jterapin, line 126: What about returning self here to enable chaining?

gems/smithy-schema/lib/smithy-schema/structure.rb
jterapin: Probably better if you turn the above into a code comment for future us.

gems/smithy-client/lib/smithy-client.rb
```diff
@@ -26,7 +26,8 @@
 require_relative 'smithy-client/log_param_formatter'
 require_relative 'smithy-client/managed_file'
 require_relative 'smithy-client/networking_error'
-require_relative 'smithy-client/pageable_response'
+require_relative 'smithy-client/paginators/page_enumerator'
+require_relative 'smithy-client/paginators/pageable_response'
```
jterapin: Ruby convention: a directory `paginators/` implies a module Paginators. But `PageEnumerator` and `PageableResponse` don't live under a `Smithy::Client::Paginators` module

Feels longwinded if we add a new module so maybe we should keep this flat.

gems/smithy-client/lib/smithy-client/pageable_response.rb
```diff
@@ -107,16 +126,29 @@ def each_page(&)
...
       # @yieldparam [Object] item
-      # @return [Enumerable, nil] Returns a new Enumerable if no block is given.
-      def each_item(&)
-        response = self
-        @paginator.items(response.data).each(&)
-        until response.last_page?
-          response = response.next_page
-          @paginator.items(response.data).each(&)
+      # @return [PageEnumerator, nil]
+      def each_item(&block)
```
jterapin, line 143: If the API does not define `items` and customer calls `.each_item`, they'll get a `NotImplementedError`. Should we document this?

jterapin (commented):
The overall approach looks good but curious what we can do about the enumerator being created at every method call. 🤔

Also - I was wondering if the PR description could contain overall info about the motivation of this approach and what decisions we make.

We should also have wangrch take a look when he comes back.

gems/smithy-client/lib/smithy-client/paginators/pageable_response.rb
```diff
@@ -107,16 +126,29 @@ def each_page(&)
...
+      # Yields the current and each following response to the given block.
+      # When called without a block, returns a {PageEnumerator}.
+      # This is an alias for {#each_page}.
+      # @yieldparam [Response] response
+      # @return [PageEnumerator, nil]
+      def each(&block)
+        return each_page unless block
+
+        each_page(&block)
+      end
```
richardwang1124: I think this could be one line? Maybe
```
def each(&block)
  each_page(&block)
end
```

richardwang1124 (commented):
Looks good overall, can take a second pass review once the comments are addressed.

gems/smithy-client/lib/smithy-client/page_enumerator.rb
sichanyoo: The allocation cost is negligible compared to the API calls each iteration triggers & returning brand new enumerator each time I think has a subtle benefit of thread safety with interleaved paging operations when it's used concurrently. But changing it to reuse enumerator is a one line change so if this is based on rubyist convention / you feel strongly for reusing, we can do that

gems/smithy-client/lib/smithy-client/paginators/page_enumerator.rb
sichanyoo: Good catch! Changed as suggested 👍

gems/smithy-client/lib/smithy-client/pageable_response.rb
sichanyoo: Q: What could be the usecase for returning self to allow chaining here? If user doesn't provide a block, they get enumerator they can chain to already, and if user provides a block, whatever the user needs to do would be specified within that block. Fwiw, V3 also returns nil instead of self for `each` when user provides a block: https://github.com/aws/aws-sdk-ruby/blob/313cb58ad76dac97c27ff6d63b99eaadd5ad4d00/gems/aws-sdk-core/lib/aws-sdk-core/pageable_response.rb#L188-L196

gems/smithy-client/lib/smithy-client/paginators/pageable_response.rb
sichanyoo: Good catch! Changed as suggested 👍

gems/smithy-client/lib/smithy-client/pageable_response.rb
sichanyoo: Good point, added a `@raise` doc comment.

gems/smithy-client/lib/smithy-client.rb
sichanyoo: I see! Updated to keep the paginator file structure flat.

gems/smithy-schema/lib/smithy-schema/structure.rb
sichanyoo: Added it as inline comment.

sichanyoo (commented):
Address initial review comments.

.github/workflows/ci.yml
```diff
@@ -29,6 +29,10 @@ jobs:
           version: '1.60.3'
       - run: smithy --version
 
+      - name: Pin rdoc for JRuby
+        if: startsWith(matrix.ruby, 'jruby')
+        run: echo 'gem "rdoc", "< 8.0"' >> Gemfile
```
jterapin, line 35: Related: https://github.com/ruby/rbs/issues/3018

Add a comment linking the ticket so that we can check on its status periodically

gems/smithy-client/lib/smithy-client/page_enumerator.rb
jterapin: I think this is fine for now, thanks for taking a look!

gems/smithy-client/lib/smithy-client/pageable_response.rb
jterapin: Makes sense; I forgotten that V3 returns `nil` instead of `self`. Let's stay consistent with V3. Thanks for the pointer!

gems/smithy-client/lib/smithy-client/pageable_response.rb
jterapin: One more question: if the items are modeled and there's no items in the response. Does that resolve to `nil` or some collection? We may want to handle it gracefully when `nil` is given instead of `NoMethodError` on `nil`.

jterapin (approved):
Added some minor comments/question. other than that, approved.

gems/smithy-client/lib/smithy-client/pageable_response.rb
sichanyoo: That's a good point; added safe navigation operator `&` to `each` call in both places so pages with zero items does not cause a failure.

.github/workflows/ci.yml
sichanyoo: Added comment link as suggested.

richardwang1124 (approved):

PR #347, opened 2026-07-07
feat: Restore instance-level plugins

jterapin:
## Context

Restores V3 behavior lost in the V4 rewrite:

```ruby
client = Service::Client.new(plugins: [MyPlugin])  # scoped to this client
```

* Single-pass only; plugins can't inject other plugins (deliberate vs V3).
* **String-name fix** — `require_plugin` resolves constants from the global namespace (`Kernel`) instead of `Smithy::Client::Plugin`, so a string-named plugin colliding with a constant under `Plugin` no longer mis-resolves.

## Other Notes
- De-dup matches V3 (by object identity): a class vs. its instance, and two instances of one class, are not deduped. Per-class uniqueness is a separate discussion, out of scope here. I made a TODO for a visit later with the team.

### DIFF
--- a/gems/smithy-client/lib/smithy-client/base.rb
+++ b/gems/smithy-client/lib/smithy-client/base.rb
@@ -47,6 +47,7 @@ module Smithy
       def build_config(plugins, options)
         config = Configuration.new
         config.add_option(:service)
+        config.add_option(:plugins)
         plugins.each do |plugin|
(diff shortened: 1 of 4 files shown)


sichanyoo (approved):
lgtm

richardwang1124 (approved):

PR #348, opened 2026-07-08
feat: Add generic Protocol plugin and protocol interface

jterapin:
## Context

Adds the generic `Protocol` plugin, a `Protocol` interface base class, and a `NoOpProtocol` default: the mechanism for per-client protocol selection built on instance-level plugins (#347).

```ruby
client = Service::Client.new(protocol: :rpc_v2_cbor)  # scoped to this client
```

The plugin resolves the `:protocol` option in `before_initialize` and installs build/parse handlers that delegate to the resolved protocol:

* **Default** (unset) — first registered protocol, or `NoOpProtocol` if none.
* **Named** (Symbol) — resolved from `client_class.protocols`.
* **Custom object** — used as-is (duck-typed; need not inherit `Protocol`).

Protocols are instances (not modules) so they can carry per-client state (e.g. `query_compatible`, codec instances).

## Other Notes

- **Dormant / define-only.** Not wired into any weld, template, or generated client — the live rpcv2Cbor path is unchanged. Hooking the mechanism up and migrating rpcv2Cbor onto this interface follows in a later PR.
- `Protocol` is a base class our own protocols inherit *and* a documented interface custom objects merely satisfy — there is deliberately no `is_a?(Protocol)` check.
- Named protocols use `Symbol` to match the rest of the config surface.
- Stacked on #347 — base that PR first; this PR's diff assumes it.
- The spec uses a test-only `protocols` registry (`define_singleton_method`) since the generated registry isn't restored until the mechanism is hooked up; a TODO marks its removal.


sichanyoo (dismissed):
lgtm

richardwang1124 (dismissed):

sichanyoo (approved):

PR #349, opened 2026-07-24
feat: Retrofit RPC v2 CBOR onto the generic Protocol plugin

jterapin:
## Issue

Follows #348 (generic Protocol plugin, define-only). That PR added the mechanism but nothing consumed it. CBOR was still served by a per-service plugin (`Plugins::RpcV2Cbor`) with its serialization split across two handler classes and a separate stubber, none of which used the new interface.

## Description

CBOR becomes the first real consumer of the generic Protocol plugin from #348, taking the mechanism live. The old per-service plugin (a plugin, two handlers, and a separate stubber) collapses into one `Client::RpcV2Cbor` object implementing the 5-method interface.

- **Interface:** drop the `Protocol` base class — it's duck-typed now, documented on `NoOpProtocol` (which also carries the stubbing methods).
- **Handlers:** `ParseHandler` deserializes success at `:build`; a separate `ErrorHandler` runs at `:sign`, inside the retry loop, so `RetryErrors` sees modeled retryable errors.
- **Default plugin:** `Plugins::Protocol` is protocol-agnostic (falls back to `NoOpProtocol`), so it's now a default. The weld only registers the protocol and its dependency.
- **Stubbing:** goes through `config.protocol` instead of a separate `:stubber`; the protocol object is its own stubber.
- **Registry:** an `add_protocols` weld hook populates `client.protocols` at codegen, replacing #348's test scaffold.
- **Docs:** the generated `:protocol` option now lists the service's supported protocols.

On errors, `parse_error` ignores anything outside `200..599` (so status-0 transport errors pass through), reads the envelope with a raw decode to get `__type`, then parses the modeled error through the codec.

## Testing

- smithy-client: 852 examples, 0 failures, 2 pending
- smithy unit: 505, 0 failures
- generated CBOR protocol tests: 113, 0 failures
- RuboCop clean

> **Note:** AWS protocols still use a String `:protocol` + a separate `:stubber` and are migrated in a follow-up PR.

---

_By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license._

_Written with AI assistance and reviewed by jterapin._


gems/smithy-client/lib/smithy-client/no_op_protocol.rb
```diff
@@ -1,19 +1,48 @@
...
+    # Also documents the protocol interface: a protocol serializes requests,
+    # deserializes responses, and builds stubbed responses for a specific
+    # wire format. A custom protocol passed via +Client.new(protocol:)+ is
+    # any object that responds to these methods.
+    #
+    # All methods are no-ops, so handlers and stubbing can delegate safely
+    # without a nil check.
     # @api private
-    class NoOpProtocol < Protocol
+    class NoOpProtocol
```
richardwang1124, line 14: nit: I'm not sure why, but this NoOpProtocol class name feels a little iffy to me. I understand we're moving away from base classes and inheritance towards duck-typing, but this class very much feels like a base class (since it also adds documentation and class usage) without compile time validation.

richardwang1124 (approved):
Looks good overall. Nice work!

gems/smithy-client/lib/smithy-client/no_op_protocol.rb
jterapin: I see your point. Let me think through this:
- Can a Smithy service client legitimately have no protocol trait? If that's not a valid state, we have a choice of: failing fast or build a client that falls back to a no-op. I could place TODO to some digging on how other smithy-based clients work. Maybe Chan will have some context on how swift client functions.
- I do agree that this is a wrong place for this documentation. I think once we have a Developer Guide for creating your own "SDK" - i think that's a better spot for this.

jterapin: Created a TODO and in-house task for this item to circle back on.

richardwang1124 (approved):

PR #351, opened 2026-08-14
refactor: drive JSON/CBOR serde off the data with a load-time wire-name map

jterapin:
## Context
This is the first PR in the current schema-serde stack.

V4 currently walks every declared structure member on each request and response and resolves wire names inline. This PR changes JSON and CBOR structure serde to work from the actual payload or input data instead, using a load-time member lookup to recover the same basic shape V3 had.

Follow-up PR #353 builds on this by replacing the temporary global lookup with protocol-aware schema extensions. I still want early feedback on the traversal change itself, which is why I am splitting the work this way.

## What changed
- `StructureShape` builds a load-time `@members_by_wire_name` index at `add_member`, keyed by each member's `location_name` and `jsonName` when present.
- JSON and CBOR structure build/parse now iterate the input values or payload keys and resolve members through that index instead of walking the full member list.
- XML is unchanged in this PR.

## Testing
- `bundle exec rake smithy-schema:spec`
- `bundle exec rake smithy-json:spec`
- `bundle exec rake smithy-cbor:spec`
- `bundle exec rake smithy:spec:unit`
- `bundle exec rake smithy:spec:protocols smithy:rbs:protocols`
- `bundle exec rake smithy-schema:rbs`
- `aws-sdk-ruby-staging` protocol validation green for `awsJson1_0`, `awsJson1_1`, `restJson1`, `awsQuery`, `ec2Query`, and `restXml`

## Notes
- The benchmark motivation for this pass is the wide-structure JSON/CBOR hot path, where resolving only the members that actually appear is materially better than walking the full modeled member list every time.
- PR #353 narrows the naming behavior so `jsonName` is only used when the active protocol actually requires it.

--
Written with AI assistance and reviewed by jterapin.

### DIFF
--- a/gems/smithy-schema/lib/smithy-schema/shapes.rb
+++ b/gems/smithy-schema/lib/smithy-schema/shapes.rb
@@ -235,7 +235,7 @@ module Smithy
         def initialize(options = {})
           super
           @members = {}
-          @members_by_wire_name = {}
+          @members_by_wire_name = {}  # Temporary wire-name map. Removed by schema extensions.
         end
 
         # @return [Hash<Symbol, MemberShape>]
@@ -247,6 +247,7 @@ module Smithy
         # @return [MemberShape]
         def add_member(name, member_shape)
           @members[name] = member_shape
+          # Temporary wire-name map. Removed by schema extensions.
           @members_by_wire_name[member_shape.location_name] = [name, member_shape]
           json_name = member_shape.traits['smithy.api#jsonName']
           if json_name && json_name != member_shape.location_name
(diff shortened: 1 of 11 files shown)


gems/smithy-schema/lib/smithy-schema/shapes.rb
```diff
@@ -246,6 +247,10 @@ def initialize(options = {})
         # @return [MemberShape]
         def add_member(name, member_shape)
           @members[name] = member_shape
+          register_wire_name(member_shape.location_name, name, member_shape)
+          json_name = member_shape.traits['smithy.api#jsonName']
+          register_wire_name(json_name, name, member_shape) if json_name && json_name != member_shape.location_name
+          member_shape
```
jterapin: Should handle both xml and jsonName somehow. We should mark this as temp-workaround until schema caching is introduced.

jterapin: naything that we may discard should have a TODO: to be discared once schema caching work is done as a heads up

richardwang1124 (approved):
Looks good to me! Do you have any benchmarks you can include to show the performance impact?

jterapin:
> Looks good to me! Do you have any benchmarks you can include to show the performance impact?

Good call. Let me try updating an overall performance improvements on the prs where I did the schema extensions improvements.

PR #357, opened 2026-08-24
feat: Pluggable transport interface + NetHTTP default transport.

sichanyoo:

> **_PR description below was written with help from AI._**

  # Pluggable transport interface + NetHTTP default transport
  
  Issue: #5196
  
  ## Summary
  
  Introduces a **pluggable HTTP transport** abstraction for `smithy-client`. The
  transport is now a swappable, duck-typed component behind a documented contract,
  replacing the previous single hard-wired `NetHTTP` handler. This is the
  groundwork that lets future transports (HTTP/2, async, CRT) drop in behind one
  stable interface, and it decouples "how bytes move" (transport) from "when to
  block and what to build on them" (handler stack).
  
  No behavioral change for existing operations — the default transport is still
  Net::HTTP over HTTP/1.1.
  
  ## What changed
  
  ### New transport/stream contracts (public swap point)
  - `Smithy::Client::Transport` — documents the transport contract. A transport is
    any object answering `REQUIRED_METHODS = [:transmit]`; `#transmit(request)`
    sends the request and returns a `Stream`.
  - `Smithy::Client::Stream` — documents the returned stream contract:
    `REQUIRED_METHODS = [:response_headers, :each_chunk, :write, :close_write,
    :abort]`. `#response_headers` returns `[status, headers]`; `#each_chunk` yields
    raw body chunks; `#write`/`#close_write` support post-transmit writes (raise
    `NotSupportedError` on HTTP/1.1); `#abort` cancels the exchange.
  - These are documented contracts (modules), not base classes — a custom
    transport only needs to answer the same methods. Shared compliance tests
    (`spec/support/stream_contract.rb`) exercise conformance.
  
  ### Default Net::HTTP transport
  - `Smithy::Client::NetHTTP::Transport` — the default HTTP/1.1 transport, built on
    Net::HTTP, owning connection management via the class-level, options-keyed
    `ConnectionPool`.
  - `Smithy::Client::NetHTTP::Stream` — the HTTP/1.1 stream. Body delivery is
    pull-based on the caller's thread via a `Fiber` that keeps Net::HTTP's
    `read_body` block alive across separate `#response_headers` and `#each_chunk`
    calls (no background reader thread on CRuby). Cross-thread `#abort` is
    mutex-guarded so cancellation cannot race normal completion; aborted/partial
    connections are discarded rather than returned to the pool.
  
  ### Handler stack
  - `Smithy::Client::SendHandler` — the `:send` step handler. Calls
    `transport.transmit`, decides *when* to block (immediately for plain
    request/response; skipped for duplex event streams), and bridges pulled bytes
    onto the push-based `Http::Response`. Guarantees the connection is released via
    `#abort` on error paths.
  - `Smithy::Client::Plugins::Transport` — the generic transport plugin. Registers
    the `:send` handler, resolves the `:transport` option (the documented swap
    point), and defines the **transport-agnostic** client options (`connect_timeout`,
    `read_timeout`, `ssl_verify_peer`, `ssl_ca_bundle`, `ssl_ca_directory`,
    `ssl_ca_store`, `http_proxy`, `http_wire_trace`, `logger`). Net::HTTP-specific
    knobs (keep-alive/continue timeouts, client certs, etc.) are configured by
    constructing `NetHTTP::Transport` directly and passing it via `:transport`.
  
  ### Errors & connection pool
  - `Smithy::Client::NotSupportedError` — raised by a stream when an operation is
    not supported by the transport (e.g. HTTP/1.1 post-transmit writes). A peer of
    `NetworkingError` at the `Smithy::Client` level.
  - `ConnectionPool` gains `#finish_session` (removes a session from the pool and
    finishes it under the pool mutex), used to fix an abort/session-ownership race:
    a cross-thread abort can no longer finish a session that has already been
    returned to the pool.
  
  ### Removed / replaced
  - Removed `Smithy::Client::NetHTTP::Handler` and `Smithy::Client::Plugins::NetHTTP`
    (and their specs).
  - The default-plugins weld now registers `Plugins::Transport` instead of
    `Plugins::NetHTTP`, so generated clients pick up the transport plugin.
  
  ## Behavioral notes / migration
  - **Client options change:** the per-client `http_*` Net::HTTP options
    (`http_open_timeout`, `http_read_timeout`, `http_verify_mode`, `http_ca_file`,
    …) are replaced by the transport-agnostic set above. Net::HTTP-specific tuning
    moves to a directly-constructed `NetHTTP::Transport` passed as `:transport`.
  - The timeout option docstrings note the effective Net::HTTP default (e.g.
    `connect_timeout`/`read_timeout` default to the transport's default, which is
    60s for Net::HTTP).
  
  ## Testing
  - New specs: transport contract, stream contract (+ shared compliance examples),
    NetHTTP transport, NetHTTP stream (headers/body/abort/truncation), send handler
    (incl. duplex path), connection-pool `#finish_session`.
  - Full `smithy-client` suite green; rubocop clean.


gems/smithy-client/lib/smithy-client/errors.rb
```diff
@@ -2,6 +2,13 @@
 
 module Smithy
   module Client
+    # Raised when an operation is not supported by a transport or its stream.
+    #
+    # For example, HTTP/1.1 streams cannot be written to after the request has
+    # been transmitted, so {NetHTTP::Stream#write} and
+    # {NetHTTP::Stream#close_write} raise this error.
+    class NotSupportedError < StandardError; end
```
jterapin: Is `NotSupportedError` intended to be part of the shared transport/stream contract, or is it really specific to the NetHTTP adapter? I also think this belongs in Errors module.

gems/smithy-client/lib/smithy-client/net_http/connection_pool.rb
```diff
@@ -238,16 +238,39 @@ def configure_http_connections(http)
...
+          end
 
-          configure_ssl_cert(http)
+          configure_ca_trust(http)
         end
 
-        def configure_ssl_cert(http)
+        # Server-certificate trust settings. Only meaningful when peer
+        # verification is enabled.
+        def configure_ca_trust(http)
```
jterapin: Much cleaner fit.

gems/smithy-client/lib/smithy-client/net_http/connection_pool.rb
```diff
@@ -238,16 +238,39 @@ def configure_http_connections(http)
         def configure_ssl(http)
           http.use_ssl = true
           http.ssl_timeout = http_ssl_timeout if http_ssl_timeout
-          return unless http_verify_mode == OpenSSL::SSL::VERIFY_PEER
+          # Net::HTTP defaults verify_mode to VERIFY_PEER; set it explicitly so
+          # a VERIFY_NONE request (ssl_verify_peer: false) is actually honored
+          # instead of silently falling back to verification.
+          http.verify_mode = http_verify_mode
```
jterapin, line 261: Nice fix!

gems/smithy-client/lib/smithy-client/net_http/connection_pool.rb
```diff
@@ -238,16 +238,39 @@ def configure_http_connections(http)
...
+
+          # Client certificate (mutual TLS) is independent of server-certificate
+          # verification, so it is always applied.
```
jterapin:
```suggestion
```

gems/smithy-client/lib/smithy-client/net_http/connection_pool.rb
```diff
@@ -238,16 +238,39 @@ def configure_http_connections(http)
...
+            return
+          end
 
-          configure_ssl_cert(http)
+          configure_ca_trust(http)
         end
 
-        def configure_ssl_cert(http)
+        # Server-certificate trust settings. Only meaningful when peer
+        # verification is enabled.
```
jterapin:
```suggestion
```

gems/smithy-client/lib/smithy-client/net_http/connection_pool.rb
```diff
@@ -238,16 +238,39 @@ def configure_http_connections(http)
...
+        def configure_ca_trust(http)
           http.ca_file = http_ca_file if http_ca_file
           http.ca_path = http_ca_path if http_ca_path
-          http.cert = http_cert if http_cert
           http.cert_store = http_cert_store if http_cert_store
+        end
+
+        # Client certificate and key for mutual TLS. Independent of server
+        # certificate verification.
```
jterapin:
```suggestion
```

gems/smithy-client/lib/smithy-client/net_http/stream.rb
```diff
@@ -0,0 +1,346 @@
...
+      # mutex so the abort is observed and cannot race the stream's normal
+      # completion.
+      #
+      # HTTP/1.1 cannot write to a request after it has been transmitted, so
+      # {#write} and {#close_write} raise {NotSupportedError}.
+      # @api private
+      class Stream
+        # Raised when the received response body is shorter than the advertised
+        # +Content-Length+. This is an HTTP/1.1 wire concern: HTTP/2 detects
+        # truncation via frame accounting / END_STREAM, not Content-Length.
```
jterapin:
```suggestion
        # Raised when the received response body is shorter than the advertised
        # +Content-Length+.
```

gems/smithy-client/lib/smithy-client/net_http/stream.rb
```diff
@@ -0,0 +1,346 @@
+# frozen_string_literal: true
+
+require 'net/http'
+
+module Smithy
+  module Client
+    module NetHTTP
+      # An HTTP/1.1 stream backed by +Net::HTTP+, implementing the stream
```
jterapin, line 8: I think some of the details here can be cut down to the most relevant parts.

gems/smithy-client/lib/smithy-client/net_http/stream.rb
```diff
@@ -0,0 +1,346 @@
...
+            super(msg)
+          end
+        end
+
+        # Internal signal raised inside the driving fiber when a cross-thread
+        # {#abort} raced ahead of session checkout. It is caught explicitly (not
+        # via the broad StandardError rescue) so it is never confused with a real
+        # networking failure. Not part of the stream contract.
+        # @api private
+        class Aborted < StandardError; end
```
jterapin: `Aborted` feels a little bare here and easy to confuse with a user-visible cancellation path. I’d consider a more specific internal name.

gems/smithy-client/lib/smithy-client/net_http/stream.rb
```diff
@@ -0,0 +1,346 @@
...
+            msg = "http response body truncated, expected #{bytes_expected} " \
+                  "bytes, received #{bytes_received} bytes"
+            super(msg)
+          end
+        end
+
+        # Internal signal raised inside the driving fiber when a cross-thread
+        # {#abort} raced ahead of session checkout. It is caught explicitly (not
+        # via the broad StandardError rescue) so it is never confused with a real
+        # networking failure. Not part of the stream contract.
```
jterapin:
```suggestion
        # Internal control-flow signal used when abort wins before session
        # checkout. Kept separate from networking errors and not part of the             
        # public stream contract.
```

gems/smithy-client/lib/smithy-client/net_http/stream.rb
```diff
@@ -0,0 +1,346 @@
...
+        #
+        # @raise [ArgumentError] If the request has an invalid HTTP method.
+        # @raise [NetworkingError] If a networking error occurs while sending
+        #   the request or reading the response headers.
+        # @return [self]
+        # @api private
+        def send_request
+          # Build (and validate) the Net::HTTP request before touching the
+          # network so an invalid verb raises ArgumentError without opening a
+          # connection or wrapping the error.
```
jterapin:
```suggestion
```

gems/smithy-client/lib/smithy-client/net_http/stream.rb
```diff
@@ -0,0 +1,346 @@
...
+          @aborted = false
+          # Guards the @done/@aborted/@error/@session transitions so a cross-thread
+          # {#abort} is observed by the driving thread and cannot race normal
+          # completion.
+          @mutex = Mutex.new
+        end
+
+        # Sends the request and reads the response status and headers, leaving
+        # the response body ready to be consumed via {#each_chunk}. Called by
+        # {Transport#transmit}; not part of the public +Stream+ contract.
```
jterapin:
```suggestion
        # Sends the request through response headers and leaves 
        # body consumption to {#each_chunk}. 
```

gems/smithy-client/lib/smithy-client/net_http/stream.rb
```diff
@@ -0,0 +1,346 @@
...
+        def send_request
+          # Build (and validate) the Net::HTTP request before touching the
+          # network so an invalid verb raises ArgumentError without opening a
+          # connection or wrapping the error.
+          net_request = build_net_request(@request)
+          @fiber = Fiber.new { run(net_request) }
+
+          # Net::HTTP applies its default Content-Type while sending the
+          # request, which happens during this first resume.
+          Thread.current[:net_http_skip_default_content_type] = true
```
jterapin: I believe newer Net::HTTP versions fixed this behavior, and we also carry patching around related cases on our side. I will suggest that you double-check whether this comment still reflects the current runtime behavior.

gems/smithy-client/lib/smithy-client/net_http/stream.rb
```diff
@@ -0,0 +1,346 @@
...
+          @fiber.resume
+          raise @error if @error
+
+          self
+        ensure
+          Thread.current[:net_http_skip_default_content_type] = nil
+        end
+
+        # @return [Array(Integer, Hash<String,String>)] The response status code
+        #   and headers. Available immediately after {#send_request} for HTTP/1.1.
```
jterapin:
```suggestion
        # Returns the response status and headers captured by {#send_request}.
```

gems/smithy-client/lib/smithy-client/net_http/stream.rb
```diff
@@ -0,0 +1,346 @@
...
+        end
+
+        # Yields raw response body chunks as they are read off the socket on the
+        # caller's thread. Blocks until the response body has been fully read.
+        # A body shorter than the advertised +Content-Length+ (HTTP/1.1
+        # truncation) is detected as the body is read and surfaced here.
+        # @yieldparam [String] chunk
+        # @raise [NetworkingError] If a networking error occurs while reading, or
+        #   if the body is shorter than the advertised +Content-Length+.
+        # @return [void]
```
jterapin:
```suggestion
        # Yields raw response body chunks on the caller's thread until EOF or
        # abort. If a fixed-length body ends before the advertised 
        # Content-Length, surfaces a NetworkingError after iteration.
```

“Blocks until the response body has been fully read” is not quite true. `each_chunk` can return early on cooperative abort (`return if aborted?`) so suggestion revision/tightening up

gems/smithy-client/lib/smithy-client/net_http/stream.rb
```diff
@@ -0,0 +1,346 @@
...
+
+          raise @error if @error
+
+          nil
+        end
+
+        # Part of the stream contract for transports that support writing to a
+        # request after transmit (e.g. HTTP/2 bidirectional event streams).
+        # HTTP/1.1 cannot write after transmit, so this always raises.
+        # @raise [NotSupportedError]
```
jterapin:
```suggestion
        # Net::HTTP HTTP/1.1 does not support post-transmit request writes.
```

Removing HTTP/2 context since that is not in scope with in the context. I would suggest you tightened up other spots as well.

gems/smithy-client/lib/smithy-client/net_http/stream.rb
```diff
@@ -0,0 +1,346 @@
...
+        def write(_bytes)
+          raise NotSupportedError, 'HTTP/1.1 does not support writing to a stream after transmit'
+        end
+
+        # Part of the stream contract for transports that support writing to a
+        # request after transmit (e.g. HTTP/2 bidirectional event streams).
+        # HTTP/1.1 cannot write after transmit, so this always raises.
+        # @raise [NotSupportedError]
```
jterapin:
```suggestion
        # Net::HTTP HTTP/1.1 does not support closing a write side 
        # after transmit.
```

gems/smithy-client/lib/smithy-client/net_http/stream.rb
```diff
@@ -0,0 +1,346 @@
...
+            session = @session
+          end
+          # Finish outside the lock to release the checked-out connection. The
+          # fiber is left suspended and will be collected; the session is
+          # intentionally not checked back into the pool. Any error from finishing
+          # is swallowed: abort runs on teardown/cleanup paths (e.g. #each_chunk's
+          # rescue, the SendHandler ensure), where it must never raise and mask
+          # the error that triggered the teardown.
+          begin
+            session&.finish
```
jterapin: I may be misreading this, but `Stream#abort` still appears to retain ownership of `@session` after `ConnectionPool#session_for` has re-added that session to the shared pool. Because `@done` is only set after `session_for` returns, it looks like there may be a gap where a cross-thread `abort` can still call `finish` on a session that has already been returned to the pool.

If that reading is right, I think we need a clearer ownership handoff here: either the stream should relinquish `@session` before pool check-in, or the pool needs an explicit path that prevents re-adding the session once abort wins.

gems/smithy-client/lib/smithy-client/net_http/stream.rb
```diff
@@ -0,0 +1,346 @@
...
+        # Aborts the stream, closing the underlying socket. The session is not
+        # returned to the connection pool (a partially-read or aborted
+        # connection is discarded). Safe to call from a thread other than the
+        # one driving {#each_chunk}: the state transition is guarded by a mutex
+        # so it cannot race normal completion, and closing the socket interrupts
+        # a blocked read on the driving thread (surfaced as a networking error),
+        # which is the desired cancellation behavior. Idempotent, and never
+        # raises (teardown errors are swallowed).
+        # @param [StandardError, nil] error
+        # @return [void]
```
jterapin:
```suggestion
        # Aborts the stream and discards the underlying session rather than
        # returning it to the pool. Safe to call from another thread, 
        # idempotent, and never raises.
```

I think the behavior here is important enough to document, but the current comment is trying to explain too much at once. I’d keep the key guarantees (cross-thread safe, idempotent, not returned to pool, does not raise) and trim the deeper control-flow narrative unless there is a specific non-obvious edge we need to preserve. 

If we could do the same for other blocks of comments, that would be great.

gems/smithy-client/lib/smithy-client/net_http/stream.rb
```diff
@@ -0,0 +1,346 @@
...
+            @aborted = true
+            @error ||= error
+            session = @session
+          end
+          # Finish outside the lock to release the checked-out connection. The
+          # fiber is left suspended and will be collected; the session is
+          # intentionally not checked back into the pool. Any error from finishing
+          # is swallowed: abort runs on teardown/cleanup paths (e.g. #each_chunk's
+          # rescue, the SendHandler ensure), where it must never raise and mask
+          # the error that triggered the teardown.
```
jterapin:
```suggestion
          # Finish outside the lock and swallow teardown errors so abort 
          # does not mask the original failure.
```

gems/smithy-client/lib/smithy-client/net_http/stream.rb
```diff
@@ -0,0 +1,346 @@
...
+        # each body chunk so the caller drives reads.
+        # @param [Net::HTTPRequest] net_request
+        def run(net_request)
+          @pool.session_for(@request.endpoint) do |session|
+            store_session(session)
+            # #abort may have raced ahead of checkout and captured a nil session
+            # (unable to finish it). Raise the Aborted sentinel so #session_for
+            # finishes this session and does not return it to the pool, preventing
+            # a leak of the checked-out connection. The abort is already recorded
+            # in @aborted.
```
jterapin:
```suggestion
            # Abort won before checkout; force session_for to 
            # discard this session.
```

gems/smithy-client/lib/smithy-client/net_http/stream.rb
```diff
@@ -0,0 +1,346 @@
...
+          end
+          nil
+        end
+
+        private
+
+        # Body of the driving fiber. Runs the full request within the pool's
+        # session block, suspending (via Fiber.yield) after headers and after
+        # each body chunk so the caller drives reads.
+        # @param [Net::HTTPRequest] net_request
```
jterapin:
```suggestion
        # Drives one request inside the pool session block, yielding 
        # after headers and each body chunk so the caller controls 
        # response consumption.
```

gems/smithy-client/lib/smithy-client/net_http/stream.rb
```diff
@@ -0,0 +1,346 @@
...
+            raise Aborted if aborted?
+
+            read_response(session, net_request)
+          end
+          mark_done
+          nil
+        rescue Aborted
+          # Intentional abort signal (raised above). #session_for has already
+          # finished the socket; nothing to surface - #each_chunk returns early
+          # when aborted.
```
jterapin:
```suggestion
          # Intentional abort path; session_for already discarded the session.
```

gems/smithy-client/lib/smithy-client/net_http/stream.rb
```diff
@@ -0,0 +1,346 @@
...
+          mark_done
+          nil
+        rescue StandardError => e
+          # A networking failure. The invalid-verb ArgumentError is validated in
+          # #send_request before the fiber exists, so any error here is a
+          # networking failure. It is recorded even if an abort is concurrently
+          # in progress; #each_chunk decides whether to surface it (it returns
+          # early when aborted, so an aborted stream stays quiet).
```
jterapin:
```suggestion
          # Record the failure; the consumer decides whether to surface it.
```

gems/smithy-client/lib/smithy-client/net_http/stream.rb
```diff
@@ -0,0 +1,346 @@
...
+          mark_error(NetworkingError.new(e))
+          nil
+        end
+
+        # Issues the request within the pool's session block, suspending after
+        # headers and after each body chunk so the caller drives reads.
+        # @param [Net::HTTPSession] session
+        # @param [Net::HTTPRequest] net_request
```
jterapin:
```suggestion
        # Runs the request inside the pool session block, yielding 
        # after headers and each body chunk so the caller drives 
        # response consumption.
```

gems/smithy-client/lib/smithy-client/net_http/stream.rb
```diff
@@ -0,0 +1,346 @@
...
+            @status = net_response.code.to_i
+            @headers = extract_headers(net_response)
+            Fiber.yield # headers are ready; hand control back to #send_request
+            net_response.read_body do |chunk|
+              @bytes_received += chunk.bytesize
+              Fiber.yield(chunk)
+            end
+            # Verify while still inside the pool session block: a truncated
+            # (peer-closed) body raises here, so #session_for finishes the
+            # socket instead of returning it to the pool.
```
jterapin:
```suggestion
            # Verify before the session can be returned to the pool so 
            # truncated bodies cause the connection to be discarded.
```

gems/smithy-client/lib/smithy-client/net_http/stream.rb
```diff
@@ -0,0 +1,346 @@
...
+            verify_content_length!
+          end
+        end
+
+        # Drives the fiber, yielding non-empty body chunks. Stops promptly if
+        # another thread requested cancellation between chunks (cooperative
+        # abort).
+        # @yieldparam [String] chunk
```
jterapin:
```suggestion
        # Resumes the fiber and yields non-empty body chunks until 
        # completion or abort.
```

gems/smithy-client/lib/smithy-client/net_http/stream.rb
```diff
@@ -0,0 +1,346 @@
...
+        def mark_error(error)
+          @mutex.synchronize do
+            @error ||= error
+            @done = true
+          end
+        end
+
+        # Constructs a +Net::HTTP+ request object from an {Http::Request}.
+        # @param [Http::Request] request
+        # @return [Net::HTTPRequest]
```
jterapin:
```suggestion
```

gems/smithy-client/lib/smithy-client/net_http/stream.rb
```diff
@@ -0,0 +1,346 @@
...
+        def build_net_request(request)
+          request_class = net_http_request_class(request)
+          req = request_class.new(request.endpoint.request_uri, net_headers_for(request))
+          # Net::HTTP adds a default Content-Type when a body is present.
```
jterapin: Update comment once we confirm the fix.

gems/smithy-client/lib/smithy-client/net_http/stream.rb
```diff
@@ -0,0 +1,346 @@
...
+          # Set the body stream when its size is unknown or greater than 0.
+          req.body_stream = request.body if !request.body.respond_to?(:size) || request.body.size.positive?
+          req
+        end
+
+        # @param [Http::Request] request
+        # @raise [ArgumentError] If the HTTP method is not a valid verb.
+        # @return [Class<Net::HTTPRequest>]
```
jterapin:
```suggestion
```

gems/smithy-client/lib/smithy-client/net_http/stream.rb
```diff
@@ -0,0 +1,346 @@
...
+        def net_http_request_class(request)
+          ::Net::HTTP.const_get(request.http_method.capitalize)
+        rescue NameError
+          raise ArgumentError, "`#{request.http_method}` is not a valid http verb"
+        end
+
+        # @param [Http::Request] request
+        # @return [Hash<String, String>]
```
jterapin:
```suggestion
```

gems/smithy-client/lib/smithy-client/net_http/stream.rb
```diff
@@ -0,0 +1,346 @@
...
+          # Setting 'identity' prevents automatic decompression.
+          headers = { 'accept-encoding' => 'identity' }
+          request.headers.each_pair do |key, value|
+            headers[key] = value
+          end
+          headers
+        end
+
+        # @param [Net::HTTPResponse] response
+        # @return [Hash<String, String>]
```
jterapin:
```suggestion
```

gems/smithy-client/lib/smithy-client/net_http/stream.rb
```diff
@@ -0,0 +1,346 @@
...
+          ::Net::HTTP.const_get(request.http_method.capitalize)
+        rescue NameError
+          raise ArgumentError, "`#{request.http_method}` is not a valid http verb"
+        end
+
+        # @param [Http::Request] request
+        # @return [Hash<String, String>]
+        def net_headers_for(request)
+          # Net::HTTP defaults decode_content=true and adds Accept-Encoding: gzip.
+          # Setting 'identity' prevents automatic decompression.
```
jterapin: Minor wording nit: `adds Accept-Encoding: gzip` is too specific here. Net::HTTP injects a default `Accept-Encoding` value and enables `decode_content` when the header is left unset; setting `Accept-Encoding` manually disables that path. I think the comment should describe that more generally rather than calling out `gzip` specifically.

gems/smithy-client/lib/smithy-client/net_http/stream.rb
```diff
@@ -0,0 +1,346 @@
...
+        end
+
+        # Verifies the number of bytes received against the advertised
+        # +Content-Length+ and raises {TruncatedBodyError} on a mismatch. Net::HTTP
+        # defaults +ignore_eof+ to +true+, so a short fixed-length body read
+        # returns normally rather than raising; this is the only place such a
+        # truncation is detected. Called from inside the pool session block so a
+        # truncated connection is finished rather than pooled.
+        # @raise [TruncatedBodyError]
+        # @return [void]
```
jterapin:
```suggestion
        # Detect short fixed-length bodies that Net::HTTP may otherwise 
        # tolerate when `ignore_eof` is true. Run this before leaving 
        # the pool session so truncated connections are discarded 
        # instead of returned to the pool.
```
The behavior here makes sense, but I think the comment is doing too much for a small private helper. The important part is that Net::HTTP may tolerate early EOF on fixed-length reads because `ignore_eof` defaults to true, so we verify `Content-Length` ourselves before the session is returned to the pool

gems/smithy-client/lib/smithy-client/net_http/transport.rb
```diff
@@ -0,0 +1,153 @@
...
+
+require 'openssl'
+
+require_relative 'connection_pool'
+require_relative 'stream'
+
+module Smithy
+  module Client
+    module NetHTTP
+      # The default HTTP/1.1 transport for {Smithy::Client}, built on Ruby's
```
jterapin, line 11: I think this doc block is trying to carry too much policy, and that makes it easy to drift or contradict itself.
The main thing that confused me is the config split:
  - it says transport-agnostic options come from client config and are forwarded here
  - it says Net::HTTP-specific options are only available by constructing the transport directly
  - but the example then passes `read_timeout` directly to `Transport.new`, which blurs that distinction immediately
 
I think we should either tighten this down to a shorter behavioral description and let the actual option declarations be the source of truth, or make the routing rules much more explicit and keep the examples aligned with them.

gems/smithy-client/lib/smithy-client/net_http/transport.rb
```diff
@@ -0,0 +1,153 @@
...
+      # returns a {Stream}. This is the default value
+      # of the +:transport+ option (see {Plugins::Transport}) and the swap point
+      # for supplying a custom transport.
+      #
+      # The transport owns its connection management through the class-level,
+      # options-keyed {ConnectionPool} (a global HTTP/1.1 pool shared across
+      # clients with identical configuration). Connections are opened on demand
+      # and returned to the pool after a complete read.
+      #
+      # ## Configuration and option mapping
```
jterapin: This section reads like it should belong in a Developer Guide/wiki when we get to it.

gems/smithy-client/lib/smithy-client/net_http/transport.rb
```diff
@@ -0,0 +1,153 @@
...
+      #
+      # The transport-agnostic options (connect/read timeouts, proxy, TLS
+      # verification and trust store, wire trace) are forwarded from client
+      # config by {Plugins::Transport}. The Net::HTTP-specific options are not
+      # exposed as client options; a customer who needs them constructs this
+      # transport directly and passes it via the +:transport+ option, for
+      # example:
```
jterapin:
```suggestion
      # The Net::HTTP-specific options are not exposed as client options. 
      # To use them, construct this transport directly and pass it via the
      # +:transport+ option, for example:
```
`customer` reads a little oddly in an API/runtime doc comment. I think `caller` or a more direct impersonal phrasing would read better here. My suggestion uses active voice instead.

gems/smithy-client/lib/smithy-client/net_http/transport.rb
```diff
@@ -0,0 +1,153 @@
...
+      #
+      #     Weather::Client.new(
+      #       transport: Smithy::Client::NetHTTP::Transport.new(
+      #         keep_alive_timeout: 30, read_timeout: 10
+      #       )
+      #     )
+      #
+      # ### Transport-agnostic options (client config, forwarded here)
```
jterapin:
```suggestion
      # ### Common client options forwarded to this transport
      
      # These client options are forwarded to the default Net::HTTP transport 
      # as shown below:
```
I think we could be clearer with intent here.  I would also suggest to update `Option` in the table below to `Client Option`

gems/smithy-client/lib/smithy-client/net_http/transport.rb
```diff
@@ -0,0 +1,153 @@
...
+      #     ssl_ca_directory   ca_path
+      #     ssl_ca_store       cert_store
+      #     http_proxy         proxy
+      #     http_wire_trace    set_debug_output
+      #     logger             (wire-trace target)
+      #
+      # ### Net::HTTP-specific options (NOT client config; set via this transport)
+      #
+      # These are configured only by constructing this transport and passing it
+      # as +:transport+; they are not client options.
```
jterapin:
```suggestion
      # ### Net::HTTP-only transport options
      #
      # To use these options, construct this transport directly and pass it via
      # +:transport+. They are not exposed as standard client config options.
```

In the below table:
I found rows like "continue_timeout -> continue_timeout" a little confusing because they are not really translations, just direct pass-through. 

That makes the table feel heavier than necessary. Maybe this section could simply say that these transport-only options map 1:1 onto the corresponding Net::HTTP settings? Then, list out the settings that we pass through?

gems/smithy-client/lib/smithy-client/net_http/transport.rb
```diff
@@ -0,0 +1,153 @@
...
+      #
+      #     Option              Net::HTTP setting
+      #     ------------------  ------------------
+      #     continue_timeout    continue_timeout
+      #     keep_alive_timeout  keep_alive_timeout
+      #     write_timeout       write_timeout
+      #     ssl_timeout         ssl_timeout
+      #     cert                cert
+      #     key                 key
```
jterapin: I found rows like `continue_timeout -> continue_timeout` a little confusing because they are not really translations, just direct pass-through. That makes the table feel heavier than necessary. Maybe this section could simply say that these transport-only options map 1:1 onto the corresponding Net::HTTP settings.

gems/smithy-client/lib/smithy-client/net_http/transport.rb
```diff
@@ -0,0 +1,153 @@
...
+      #
+      #     Note: the pool evicts idle connections based on keep_alive_timeout,
+      #     so that option covers both keep-alive and idle eviction.
+      #
+      # HTTP/2-specific options (for example max_concurrent_streams,
+      # read_chunk_size, enable_alpn) belong to an H2 transport, not this one.
```
jterapin:
```suggestion
```
Our scope is net-http so I would keep it to only HTTP/1.1 details.

Also this transport class is marked as api private.

gems/smithy-client/lib/smithy-client/net_http/transport.rb
```diff
@@ -0,0 +1,153 @@
...
+          stream
+        end
+
+        private
+
+        # Maps the transport options onto the Net::HTTP {ConnectionPool} options.
+        # Transport-agnostic options use transport-neutral names and are
+        # translated here; Net::HTTP-specific options are passed through.
+        # @return [Hash]
+        def pool_options
```
jterapin, line 112: We also maintain a list of pool options in the `ConnectionPool` - we should really streamline that. We could sanitize the options there or here. We should pick a one place.

gems/smithy-client/lib/smithy-client/plugins/transport.rb
```diff
@@ -0,0 +1,154 @@
...
+      #   proxy, TLS verification and trust store, wire trace) that every
+      #   built-in transport honors.
+      #
+      # These options use transport-neutral names and are forwarded to the
+      # default transport at construction. Transport-specific knobs (for example
+      # Net::HTTP's keep-alive/continue timeouts or client certificates) are not
+      # exposed as client options: a customer who needs them constructs a
+      # transport instance directly and passes it as +:transport+ (see
+      # {Smithy::Client::NetHTTP::Transport}). A customer-supplied transport is
+      # used as-is, so these options do not apply to it.
```
jterapin:
```suggestion
      # Generic transport plugin for Smithy clients. It:
      #
      # * registers the shared +:send+ handler ({SendHandler}),
      # * refines the common client transport options, and
      # * constructs the default transport when one is not supplied explicitly.
      #
      # These client options cover the shared transport settings exposed on
      # `Client.new(...)`. Transport-specific options remain adapter-specific and
      # must be configured on the transport instance itself. 
      # If a caller supplies a transport via +:transport+, that instance 
      # is used directly.
```
Tightened it up and using caller instead of customer.

gems/smithy-client/lib/smithy-client/plugins/transport.rb
```diff
@@ -0,0 +1,154 @@
...
+      # @api private
+      class Transport < Plugin
+        # The transport-agnostic client options forwarded to the default
+        # transport. Keeping them in one list keeps the option definitions and
+        # the forwarding in sync: add a name here and define the matching
+        # +option(...)+ below and it is forwarded automatically.
+        TRANSPORT_OPTIONS = %i[
```
jterapin, line 30: I think there’s a bit of a disconnect here now that the transport config surface and ConnectionPool::OPTIONS are maintained separately. 

Since Transport#pool_options is doing a handwritten translation, it feels easy for the two lists to drift. We can use this const as a reference point for default transport options?

gems/smithy-client/lib/smithy-client/plugins/transport.rb
```diff
@@ -0,0 +1,154 @@
...
+          connect_timeout read_timeout ssl_verify_peer ssl_ca_bundle
+          ssl_ca_directory ssl_ca_store http_proxy http_wire_trace logger
+        ].freeze
+
+        ## Connections
+
+        option(
+          :connect_timeout,
+          default: nil,
```
jterapin, line 39: Shouldn't we document net-http transport defaults here? How will the user find out the defaults if they are using default net-http. Same question for others listed here.

gems/smithy-client/lib/smithy-client/plugins/transport.rb
```diff
@@ -0,0 +1,154 @@
...
+        option(
+          :transport,
+          doc_type: 'Smithy::Client::NetHTTP::Transport',
+          docstring: <<~DOCS) do |config|
+            The transport used to send requests. Defaults to an HTTP/1.1 transport based on
+            Net::HTTP ({Smithy::Client::NetHTTP::Transport}), constructed with the resolved
+            transport-agnostic client options. Supply a custom object responding to
+            `#transmit(request)` (returning a stream) to swap the transport, or a
+            directly-constructed `NetHTTP::Transport` to set Net::HTTP-specific knobs. A
+            customer-supplied transport instance is used as-is.
```
jterapin: I would suggest that you update the wording, "customer-supplied" here

gems/smithy-client/lib/smithy-client/send_handler.rb
```diff
@@ -0,0 +1,73 @@
...
+    #
+    # This handler is transport-agnostic: it depends only on the stream contract
+    # (+#response_headers+, +#each_chunk+, +#abort+) and never on a concrete
+    # transport.
+    # Protocol-specific concerns - how blocking is implemented, connection
+    # pooling, HTTP/1.1 body-truncation detection - live in the transport and
+    # its stream. This handler only decides *when* to block (immediately, for
+    # plain request/response operations) and bridges the pulled bytes onto the
+    # push-based {Http::Response} the rest of the stack consumes.
+    # @api private
```
jterapin: I don’t think I’m asking for a redesign in this PR, but I do want to push on the `SendHandler` contract a bit more before approval.

My concern is that the current docs describe this boundary as more transport-agnostic than the code actually is. The handler is not tied to a concrete adapter like Net::HTTP, but it is still defining a fairly specific staged streaming model that transports must fit:
- `transmit` returns before body consumption is complete
- headers are available as a distinct phase
- body is consumed later via ordered chunk reads
- a live handle supports abort during that exchange

So I think the distinction here is less “agnostic” and more “adapter-independent but contract-shaping.” Alternate transports can plug in, but only if they adapt themselves to this stream model. 

Right now it feels like we are making every transport look like a staged pull stream, which means more push/event-style clients have to reshape themselves to fit us. That makes the transport boundary feel less swappable than the docs suggest, and it leaves implementers to infer what timing, ordering, and cancelation rules the SDK is actually expecting.

One idea I keep coming back to is whether more of this shaping should live inside the transport instead. In that model, transports could keep their own internal execution model, but turn it into the SDK’s expected response lifecycle at the boundary: headers first, then ordered body delivery, then one terminal done/error. The thing returned from `transmit` could then stay smaller and focus mostly on live control like `abort`, plus write-side methods if we need duplex. 

I’m not settled on this approach, but I wanted to float it as a possible direction.

jterapin (commented):
This is a meaningful step toward a cleaner transport boundary, and I appreciate how much groundwork this PR is laying. I still have a couple blocking concerns around abort/session ownership and how explicitly we define the `SendHandler` contract (we may need to regroup for team discussion), but I think the overall direction is promising.

It would also be helpful to include a PR description for review context. As a note, I’ll be reviewing the specs in a second pass.

gems/smithy-client/lib/smithy-client/send_handler.rb
```diff
@@ -0,0 +1,73 @@
...
+
+      # Resolves the response by reading headers and draining the body into the
+      # push-based {Http::Response}.
+      # @param [#response_headers, #each_chunk] stream
+      # @param [Http::Response] resp
+      # @return [void]
+      def resolve_response(stream, resp)
+        status, headers = stream.response_headers
+        resp.signal_headers(status, headers)
+        stream.each_chunk { |chunk| resp.signal_data(chunk) }
```
alextwoods, line 69: I forget - are we still relying injecting deserialization/unmarshalling logic into these event listeners?  Did we ever discuss removing the event based transport interactions and relying on our existing handler based stack for that instead?

gems/smithy-client/lib/smithy-client/plugins/transport.rb
```diff
@@ -0,0 +1,154 @@
...
+          doc_type: 'URI::HTTP, String',
+          docstring: <<~DOCS)
+            A proxy to send requests through. Formatted like 'http://proxy.com:123'.
+          DOCS
+
+        ## Transport
+
+        option(
+          :transport,
+          doc_type: 'Smithy::Client::NetHTTP::Transport',
```
alextwoods, line 119: I think this fits in with Juli's comment about the SendHandler contract - I think if we're trying to make this contract understandable we need to be explicit about it - I think for us that means having an abstract base/module that documents the required methods and their contractually required behaviors.  Going further, I think we would want a set of shared compliance tests that exercise that interface that any implementions should use to validate their conformance to that interface.

That goes I think both for `Transport` and `Stream` (I see right now both have a contract implementation in the net_http namespace), but I think both need tight contractual definition.

gems/smithy-client/lib/smithy-client/net_http/stream.rb
alextwoods: If we do still need this I think this is a fiber bug to set this here. Thread state is fiber local which means the fiber that needs it won't actually get this (I think, if I'm reading it correctly).  My guess is that we may no longer need this patch depending on what ruby versions we support, but should be checked.

gems/smithy-client/lib/smithy-client/net_http/stream.rb
```diff
@@ -0,0 +1,346 @@
...
+          # early when aborted, so an aborted stream stays quiet).
+          mark_error(NetworkingError.new(e))
+          nil
+        end
+
+        # Issues the request within the pool's session block, suspending after
+        # headers and after each body chunk so the caller drives reads.
+        # @param [Net::HTTPSession] session
+        # @param [Net::HTTPRequest] net_request
+        def read_response(session, net_request)
```
alextwoods: Was confused reading this for awhile - `read_response` to me indicates that it reads the response only, but this both initiates the request, sends it and then reads the response.

alextwoods (commented):
Overall I think this generally makes sense, but I have two big areas worth discussing: 
1. I think we need much more explicit interface/contract definition (see other comments). 
2. With the new transport model, I think there ends up being a mismatch between the way the SDK's handler stack interacts with responses and the way the transport works.  I'm thinking specifically of the response event hooks like `signal_headers` and `signal_data`.  The new transport contract has effectively already inverted these and its weird for Deserialization/unmarshallling/parsers (Whatever we're calling these) to register logic in event handlers rather than implementing the logic in the handler itself.

gems/smithy-client/lib/smithy-client/errors.rb
sichanyoo: It's intended as part of the shared stream contract, not NetHTTP-specific. Updated the doc string. On placement: I kept it at Smithy::Client
level to stay consistent with NetworkingError, which is the sibling
runtime transport/stream error and also lives at the top level (not in
Errors). The Errors module currently holds modeled/operation errors
(MissingContentLength). If the convention is that all error types live in
Errors, I'd suggest moving NetworkingError and NotSupportedError together
in a follow-up rather than splitting the two. Happy to go either way

gems/smithy-client/lib/smithy-client/net_http/stream.rb
sichanyoo: Trimmed the class doc to the essentials: pull-based/fiber model, the
same-thread resume + cross-thread abort rule, and the HTTP/1.1 no-write-
after-transmit note.

gems/smithy-client/lib/smithy-client/net_http/stream.rb
sichanyoo: Renamed to `InternalAbortSignal` & updated doc string to clarify it.

gems/smithy-client/lib/smithy-client/net_http/stream.rb
sichanyoo: @jterapin 
Good catch to double-check. Checked that net-http >= 0.7.0 removed the
behavior ([ruby/net-http#207](https://github.com/ruby/net-http/pull/207), [released Oct 2025](https://github.com/ruby/net-http/releases#release-v0.7.0)), but the workaround is
still live because our min supported Ruby (3.3, and 3.4) bundles an older
net-http. So I kept the code.

@alextwoods
Fixed bug by moving the flag setting inside
the fiber, immediately before session.request (and clear it once the response
block runs). Verified the in-fiber value is visible where the patch reads it.

gems/smithy-client/lib/smithy-client/net_http/stream.rb
sichanyoo: Changed to your
suggested wording (yields until EOF or abort; surfaces NetworkingError
after iteration on a short fixed-length body).

gems/smithy-client/lib/smithy-client/net_http/stream.rb
sichanyoo: Reduced to "Net::HTTP HTTP/1.1 does not support post-transmit
request writes." Also tightened the other over-scoped comments.

gems/smithy-client/lib/smithy-client/net_http/stream.rb
sichanyoo: Fixed with a pool-side ownership handoff:
- ConnectionPool#finish_session removes the session from the pool and
  finishes it under @pool_mutex, so abort-discard and check-in are
  serialized on the same lock.
- Stream#abort clears @session under @mutex and discards via the pool
  instead of calling finish directly.
- Stream#run now relinquishes ownership (release_session: @done=true,
  @session=nil under @mutex) inside the yielded block, before session_for
  re-pools — so a late abort no-ops.
Added regression specs (abort no-op after completion; discard via pool
mid-stream; #finish_session removes+finishes).

gems/smithy-client/lib/smithy-client/net_http/stream.rb
sichanyoo: Applied your condensed version (cross-thread safe, idempotent, not returned
to pool, does not raise) and dropped the control-flow narrative. Did the
same for the other heavy blocks (verify_content_length!, class doc).

gems/smithy-client/lib/smithy-client/net_http/stream.rb
sichanyoo: Applied: "Intentional abort path; session_for already discarded the session."

gems/smithy-client/lib/smithy-client/net_http/stream.rb
sichanyoo: Confirmed and updated: the default Content-Type applies on net-http < 0.7.0
only; Patches suppresses it during send. Comment now says that and cross-
references Patches.

gems/smithy-client/lib/smithy-client/net_http/stream.rb
sichanyoo: Agreed. Reworded generally: when Accept-Encoding is unset, Net::HTTP injects
a default and enables decode_content (transparent decompression); setting it
to 'identity' opts out. Dropped the gzip-specific mention.

gems/smithy-client/lib/smithy-client/net_http/stream.rb
sichanyoo: Applied your tighter version: detect short fixed-length bodies Net::HTTP
tolerates when ignore_eof is true; run inside the pool session block so
truncated connections are discarded rather than pooled.

gems/smithy-client/lib/smithy-client/net_http/transport.rb
sichanyoo: Agreed; the read_timeout example contradicted the split. Rewrote: the
example now only passes a Net::HTTP-only option (keep_alive_timeout), and
the two sections are consistent.

gems/smithy-client/lib/smithy-client/net_http/transport.rb
sichanyoo: Applied your suggestion; switched to "caller"/impersonal active voice.

gems/smithy-client/lib/smithy-client/net_http/transport.rb
sichanyoo: Applied: heading is now "Common client options forwarded to this transport,"
and the table column reads "Client Option."

gems/smithy-client/lib/smithy-client/net_http/transport.rb
sichanyoo: Applied your "Net::HTTP-only transport options" heading and the construct-
directly phrasing.

gems/smithy-client/lib/smithy-client/net_http/transport.rb
sichanyoo: Agreed; replaced the table with a one-liner: these map 1:1 onto the
corresponding Net::HTTP settings, followed by the list (continue_timeout,
keep_alive_timeout, write_timeout, ssl_timeout, cert, key).

gems/smithy-client/lib/smithy-client/net_http/transport.rb
sichanyoo: Removed the HTTP/2 paragraph & unneeded private tag.

gems/smithy-client/lib/smithy-client/net_http/transport.rb
sichanyoo: Agreed the two can drift. For this PR I documented the coupling and named
ConnectionPool::OPTIONS as the source of truth for pool settings. A full
consolidation (single sanitize point) is a slightly larger change; I'd like
to do it as a focused follow-up so it doesn't expand this PR. Thoughts?

gems/smithy-client/lib/smithy-client/plugins/transport.rb
sichanyoo: Applied your tightened version, using "caller" throughout.

gems/smithy-client/lib/smithy-client/plugins/transport.rb
sichanyoo: Agreed. Added a note making TRANSPORT_OPTIONS the reference for the
forwarded set, and calling out that NetHTTP::Transport translates onto
ConnectionPool::OPTIONS so the surfaces are coupled. Tied to the
consolidation follow-up in the pool_options thread.

gems/smithy-client/lib/smithy-client/plugins/transport.rb
sichanyoo: Kept the options as nil (so the transport stays the
source of truth and a custom transport can choose its own defaults), and
documented the effective Net::HTTP defaults at BOTH layers so they're
discoverable wherever a user is looking:

- generic client options (plugins/transport.rb): connect_timeout /
  read_timeout note "the default Net::HTTP transport uses 60 seconds."
- Net::HTTP-specific options (net_http/transport.rb @option docs):
  connect_timeout/read_timeout/write_timeout = 60, keep_alive_timeout = 2,
  continue_timeout/ssl_timeout = nil (no wait).

gems/smithy-client/lib/smithy-client/plugins/transport.rb
sichanyoo: Updated to "caller-supplied."

gems/smithy-client/lib/smithy-client/send_handler.rb
sichanyoo: Agreed with the characterization. Reworded the doc from "transport-agnostic"
to "adapter-independent but contract-shaping" and spelled out the staged
pull model the stream must satisfy: transmit returns before body consumption,
headers as a distinct phase, ordered each_chunk body pull, and live abort.

Went a step further to make the contract concrete (also per Alex's comment):
added abstract Transport and Stream contract classes documenting the required
methods/behaviors, plus shared compliance tests (spec/support/
transport_contract.rb, stream_contract.rb) wired into the NetHTTP specs. The
"push staging into the transport" redesign is a good follow-up for team
discussion; kept this PR to docs + contract definition.

gems/smithy-client/lib/smithy-client/send_handler.rb
sichanyoo: Deserialization logic already relies on handler based stack (Plugins::Protocol installs ParseHandler/ErrorHandler, which
call protocol.parse_data / parse_error and read context.http_response.body
.read synchronously. The protocols (RpcV2Cbor, NoOpProtocol) expose those as
plain methods). 

The callback here is only rly used for (unless I missed something) `response_target:` streaming. Whether we want to keep it for response_target or to fold it in to SendHandler I think can be followup discussion.

gems/smithy-client/lib/smithy-client/plugins/transport.rb
sichanyoo: Agreed, and done for this PR:
- lib/smithy-client/transport.rb: abstract Smithy::Client::Transport
  documenting the #transmit contract and required behaviors (send before
  return, headers available without body consumption, error semantics).
- lib/smithy-client/stream.rb: abstract Smithy::Client::Stream documenting
  the staged pull model (headers phase, then ordered body, then terminal),
  abort/write/close_write, and the concurrency rules.
- NetHTTP::Transport and NetHTTP::Stream now subclass these.
- Shared compliance tests: spec/support/transport_contract.rb ("a transport")
  and stream_contract.rb ("a stream"), included from the NetHTTP specs via
  it_behaves_like. Any future implementation runs the same suite to validate
  conformance.

gems/smithy-client/lib/smithy-client/net_http/stream.rb
sichanyoo: Agreed, the name was misleading. Renamed to #perform_exchange (sends the
request and reads the response) with an updated doc.

gems/smithy-client/lib/smithy-client/send_handler.rb
alextwoods: Ah right, wasn't sure how we did that.

The key tension and disconnect in the design still stands though - and I think this goes a bit to what Juli was describing on the SendHandler's contract vs behavior. The `each_chunk`  (pull) vs `response.signal_data` (push) feels like it creates two different response models.

sichanyoo: Yeah.. that's true. Instead of using the listener API on response, I think we can just fold it into SendHandler by calling status / header / body writes directly from SendHandler. As for specific logic that we DO handle via these handlers, we should be able to relocate them to appropriate places as needed.

Quick approach survey with AI:
```
Have SendHandler drive the stream directly and expose the lifecycle as plain method calls / return values instead of emitted
  events:
  
  def resolve_response(stream, resp)
    status, headers = stream.response_headers
    resp.status_code = status          # plain setters, no signal_headers/emit
    resp.headers = headers
    stream.each_chunk { |chunk| resp.body.write(chunk) }  # write directly, no signal_data
  end
  
  Then move the three ResponseTarget concerns to where the information exists:
  
  - Body-IO selection (currently on_headers): SendHandler (or a small collaborator) picks the body IO after reading status/headers
  but before draining, consulting context[:response_target]. This is the same decision, just inline instead of via a listener.
  - Success close / error cleanup (on_success/on_error): move into the existing ResponseTarget handler's post-@handler.call section —
  it already wraps the whole stack, so after the inner call returns it can inspect response.error/status and close or unlink. No
  events needed; it's ordinary handler control flow.
  - Response#on_error capture: Response can read context.http_response.error directly (the parse/error handlers already set it),
  instead of subscribing.
  - ParamConverter file close: do it in its own handler's ensure/post-call block instead of on_done.
  
  Tradeoff: SendHandler gains awareness of "there's a body IO to choose," which slightly couples it to the response-target feature.
  Cleaner: give SendHandler a tiny seam like resp.body_for(status, headers) that defaults to the existing StringIO and that
  ResponseTarget overrides — keeps SendHandler generic while removing the event bus.
```

gems/smithy-client/lib/smithy-client/net_http/transport.rb
jterapin: Per our conversation, yes we can do this in a follow-up pr.

gems/smithy-client/lib/smithy-client/plugins/transport.rb
jterapin: I think those are reasonable places to document this, but I wouldn’t characterize them as true replacement. 

The generated clients take these `option(...)` definitions, and their config docs inherit that information directly. Since users are already used to finding this information through that path, I’m not sure what the better alternative would be.

[Example of rendered configs](https://github.com/smithy-lang/smithy-ruby/blob/main/projections/weather/lib/weather/client.rb#L141-L143)

gems/smithy-client/lib/smithy-client/net_http/stream.rb
```diff
@@ -0,0 +1,342 @@
...
+        # drives reads. Runs inside the driving fiber.
+        # @param [Net::HTTPSession] session
+        # @param [Net::HTTPRequest] net_request
+        def perform_exchange(session, net_request)
+          # On net-http < 0.7.0, Net::HTTP applies a default Content-Type when a
+          # request has a body; {Patches} suppresses that via this flag. Set it
+          # here (inside the driving fiber) because Thread#[] is fiber-local and
+          # the request is sent from this fiber. No-op on net-http >= 0.7.0,
+          # which removed the behavior.
+          Thread.current[:net_http_skip_default_content_type] = true
```
jterapin, line 205: Make sure we have a TODO somewhere to drop after min Ruby version. I feel like this can be easily forgotten.

gems/smithy-client/spec/smithy-client/net_http/stream_spec.rb
```diff
@@ -0,0 +1,147 @@
...
+            expect(headers['x-foo']).to eq('bar')
+          end
+
+          it 'sends the request body' do
+            @body = StringIO.new('request-body')
+            stub = stub_request(:post, endpoint).with(body: 'request-body')
+            described_class.new(pool, Http::Request.new(
+                                        endpoint: endpoint, http_method: 'POST', body: @body
+                                      )).send_request
+            expect(stub).to have_been_requested
```
jterapin: Rough alignment here.

gems/smithy-client/lib/smithy-client/errors.rb
jterapin: I'm not entirely sure about the intention behind this module (whether it is reserved for model/operation). But if you want to keep this method under `Smithy::Client` namespace, you should have a dedicated file to follow the library convention.

I also feel like this can be used for other cases when we define abstract classes, maybe! 🤔

gems/smithy-client/lib/smithy-client/plugins/transport.rb
```diff
@@ -0,0 +1,154 @@
...
+        # @raise [ArgumentError] If a supplied transport does not respond to
+        #   +#transmit+.
+        def before_initialize(_client_class, options)
+          transport = options[:transport]
+          return if transport.nil? || transport.respond_to?(:transmit)
+
+          raise ArgumentError,
+                ':transport must respond to #transmit(request), got ' \
+                "#{transport.class}"
+        end
```
jterapin, line 153: I think right now... you can just pass an instance of this class and it will bypass the current check we have (since it responds to :transmit) OR you can update that check to do a .is_a? check. 

I think modules vs classes was an open choice here There's a lot of opinions on this but I do think inheritance is for shared behavior that subclasses inherit, and modules can carry shared behavior anyway. I'll leave that choice to you.

```ruby
  module Client
    # The transport contract: sends a request and returns a {Stream} the handler
    # stack consumes. This is the swap point exposed as the +:transport+ client
    # option (see {Plugins::Transport}); {NetHTTP::Transport} is the built-in
    # HTTP/1.1 implementation.
    #
    # ....
    module Transport
      # The messages every transport must answer. This is the compatibility
      # surface for third-party transports.
      REQUIRED_METHODS = %i[transmit].freeze
    end
  end
end
```

For transport's before init check, we can do something like:
```ruby

return if transport.nil? || 
   Client::Transport::REQUIRED_METHODS.all? { |m| transport.respond_to?(m) }

```
Maybe allowing duck-typing is the right call? so they won't have be locked into inheriting a class ? 🤔 defining method(s) satisfies the contract? I'm open to either ideas. I could see an argument for both.

gems/smithy-client/lib/smithy-client/stream.rb
```diff
@@ -0,0 +1,79 @@
...
+    # A live handle also supports {#abort} for cancellation, and
+    # {#write}/{#close_write} for transports that allow post-transmit writes
+    # (others raise {NotSupportedError}).
+    #
+    # ## Concurrency
+    #
+    # {#each_chunk} runs on the caller's thread. {#abort} MAY be called from
+    # another thread; implementations must make the abort/completion transition
+    # race-safe and must not deliver chunks after an abort.
+    # @api private
```
jterapin: we should remove the api private here and transport.

jterapin (commented):
We're almost there. Once you decide whether you'd rather switch to modules or update the `before_initialize` check to be subclass-friendly - I'll be comfortable approving. 

With the condition that we do the follow-up PRs we mentioned in this one.

gems/smithy-client/lib/smithy-client/net_http/stream.rb
sichanyoo: Added. Strengthened the pre-existing TODO in patches.rb. Also added a short additional TODO pointer on the stream.rb.

gems/smithy-client/spec/smithy-client/net_http/stream_spec.rb
sichanyoo: Fixed alignment; must've been edited by rubocop autofix.

gems/smithy-client/lib/smithy-client/errors.rb
sichanyoo: Moved NotSupportedError to its own file (lib/smithy-client/not_supported_error.rb), mirroring networking_error.rb, and kept it under Smithy::Client. Left the Errors module for modeled/operation errors (MissingContentLength). Agree it can serve other abstract-class cases; happy to reuse it as those land.

gems/smithy-client/lib/smithy-client/stream.rb
sichanyoo: Oops, yep, removed.

gems/smithy-client/lib/smithy-client/plugins/transport.rb
sichanyoo: You're right about the bypass. Went with your module + REQUIRED_METHODS suggestion and duck typing 👍

gems/smithy-client/lib/smithy-client/net_http/patches.rb
```diff
@@ -17,7 +17,9 @@ def self.apply!
         # is set. net-http 0.7.0+ removed this entirely, so the patch
         # is only applied when the method exists. Unable to remove this
         # completely due to bundled net-http versions in Ruby 3.2-3.3.
-        # TODO: re-evaluate when we determine min version for smithy-ruby/v4 GA.
+        # TODO: remove this patch, the Stream skip-flag that drives it, and its
+        # spec once the min supported Ruby ships net-http >= 0.7.0 (i.e. drops
+        # Ruby 3.3/3.4). Keyed on the min Ruby bump so it is not forgotten.
```
jterapin, line 22: Thanks! I suspect we might be extracting this out once we get close

jterapin (approved):
LGTM thank you!!

PR #367, opened 2026-09-14
perf: reuse cached helpers across serde codecs

jterapin:
## Summary

Applies behavior-preserving serde and logging optimizations on top of the client consumer migration.

- Reuses JSON, CBOR, and XML codec builder/parser helpers across calls.
- Uses typed union member and value accessors instead of repeated member-type lookup.
- Retains XML collection binding values during serialization and parsing.
- Migrates LogParamFilter to cached Schema metadata.
- Adds codec-reuse and flattened XML collection behavior coverage.

## Merge order

Merge before #369 and #366.

## Validation

- Focused CBOR builder and codec specs: 16 examples, 0 failures.
- Focused JSON builder and codec specs: 24 examples, 0 failures.
- Focused XML builder and codec specs: 23 examples, 0 failures.
- RuboCop and git diff --check passed.


gems/smithy-cbor/lib/smithy-cbor/codec.rb
```diff
@@ -6,22 +6,23 @@ module Cbor
...
+        @builder = Builder.new(options)
+        @parser = Parser.new(options)
       end
 
       # @param [Shape] shape
       # @param [Object] data
       # @return [String, nil]
       def build(shape, data)
-        Builder.new(@options).build(shape, data)
+        @builder.build(shape, data)
```
richardwang1124, line 17: Could using the same builder each time `build` is called cause multithreading issues? Previously each time `build` was called a new Builder was created.

jterapin: Good callout. I ended up making some changes so that we don't use any per-call state from now on. When we switch protocols, the codecs are re-init.

PR #369, opened 2026-09-14
fix: align runtime and test compatibility

jterapin:
## Summary

Applies small runtime, type-signature, and test-setup compatibility fixes on top of #367.

- Restores RFC 3986 query URI escaping through Smithy::Client::Util.
- Aligns nullable Schema shape RBS declarations with runtime initialization.
- Replaces deprecated SimpleCov filter calls in client and XML specs.

## Merge order

Merge after #367 and before #366.

## Validation

- Focused client utility spec: 2 examples, 0 failures.
- Focused XML builder spec: 21 examples, 0 failures.
- smithy-schema RBS task: 136 examples, 0 failures.
- RuboCop and git diff --check passed.


gems/smithy-xml/spec/spec_helper.rb
```diff
@@ -2,12 +2,12 @@
 
 require 'simplecov'
 SimpleCov.start do
-  add_filter '/spec/'
-  add_filter 'gems/smithy/'
-  add_filter 'gems/smithy-cbor/'
-  add_filter 'gems/smithy-client/'
-  add_filter 'gems/smithy-json/'
-  add_filter 'gems/smithy-schema/'
+  skip '/spec/'
```
richardwang1124, line 5: Also I just noticed these, but what does changing `add_filter` to `skip` do?

richardwang1124 (approved):

gems/smithy-xml/spec/spec_helper.rb
jterapin: There’s no behavior change. `skip` is the replacement for the deprecated `add_filter` API so we won't get annoying SimpleCov warning anymore! :D

